PR #31, opened 2017-05-12
HYD-7339 Verify targets mounted on expected hosts during create_filesystem_standard

tanabarr:
Make it standard practice to verify if filesystem has mounted targets on
expected hosts.

  - verify that the fs has targets mounted on the primary hosts that we
    set explicitly in set_volume_mounts during filesystem creation

### DIFF
--- a/chroma-manager/tests/integration/core/chroma_integration_testcase.py
+++ b/chroma-manager/tests/integration/core/chroma_integration_testcase.py
@@ -558,16 +558,6 @@ class ChromaIntegrationTestCase(ApiTestCaseWithTestReset):
         )
         self.assertTrue(response.successful, response.text)
 
-#    def verify_volume_mounts(self, volume, expected_primary_host_id, expected_secondary_host_id):
-#        """
-#        Verify that a given volume has the expected values for its primary and secondary hosts.
-#        """
-#        for node in volume['volume_nodes']:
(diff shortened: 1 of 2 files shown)


chroma-manager/tests/integration/core/chroma_integration_testcase.py
```diff
@@ -704,3 +705,48 @@ def targets_in_state(self, state):
         mounted_targets = [t for t in targets if t['state'] == state]
 
         return len(mounted_targets) == len(targets)
+
+    def check_targets_for_volumes_started_on_expected_hosts(self, filesystem_id, volumes_to_expected_hosts, assert_true):
+        """
+        Private function providing shared logic for public facing target active host checks.
```
johnsonw: This isn't private anymore is it?

tanabarr:
@brianjmurrell I agree with user not being concerned about what primary/secondary are, however this is about ensuring the assumptions that we make in the tests are accurate as soon we create a file system, not causing failures later on because primary/secondary are not what we think they are

chroma-manager/tests/integration/core/chroma_integration_testcase.py
tanabarr: no it's now being used outside the module

PR #39, opened 2017-05-18
Fix bugs in simulator corosync configuration due to race conditions

tanabarr:
Simulator tests failing due to https://jira.hpdd.intel.com/browse/HYD-7340

In order to remove the race condition

  - clear the simulator server data at the beginning of each test
  - don't return information from pacemaker/corosync when packages not installed

Change-Id: Iece96863aa1e2111cf355ac048c41a2255ef0529

### DIFF
--- a/cluster-sim/cluster_sim/fake_server.py
+++ b/cluster-sim/cluster_sim/fake_server.py
@@ -38,7 +38,7 @@ from cluster_sim import utils
 from chroma_agent.device_plugins.action_runner import CallbackAfterResponse
 from cluster_sim.fake_action_plugins import FakeActionPlugins
 from cluster_sim.fake_device_plugins import FakeDevicePlugins
-from chroma_agent.chroma_common.lib.agent_rpc import agent_result, agent_result_ok, agent_error
+from chroma_agent.chroma_common.lib.agent_rpc import agent_result, agent_result_ok
 from chroma_agent.chroma_common.lib.date_time import IMLDateTime
 
 # Simulated duration, in seconds, from the time a server shutdown is issued
@@ -91,6 +91,7 @@ class FakeServer(utils.Persisted):
         'lnet_up': False,
         'stats': None,
         'packages': {},
+        'client_mounts': {},
         'corosync': CorosyncState('stopped', 3000),
         'pacemaker': PacemakerState('stopped')
     }
--- a/cluster-sim/cluster_sim/fake_server.py
+++ b/cluster-sim/cluster_sim/fake_server.py
@@ -1,7 +1,7 @@
 #
 # INTEL CONFIDENTIAL
 #
-# Copyright 2013-2016 Intel Corporation All Rights Reserved.
+# Copyright 2013-2017 Intel Corporation All Rights Reserved.
 #
 # The source code contained or described herein and all documents related
 # to the source code ("Material") are owned by Intel Corporation or its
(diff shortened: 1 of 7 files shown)


cluster-sim/cluster_sim/fake_server.py
```diff
@@ -44,11 +44,11 @@
 # Simulated duration, in seconds, from the time a server shutdown is issued
 # until it's stopped. When simulating a shutdown, it will always take at
 # least this long.
-MIN_SHUTDOWN_DURATION = 10
+MIN_SHUTDOWN_DURATION = 5  # 10
```
chrisgearing: Why make this change?

cluster-sim/cluster_sim/fake_server.py
```diff
@@ -44,11 +44,11 @@
...
 # Simulated duration, in seconds, from the time a server is started until
 # it's running. When simulating a startup, it will always take at least
 # this long.
-MIN_STARTUP_DURATION = 20
+MIN_STARTUP_DURATION = 5  # 20
```
chrisgearing: And this change

cluster-sim/cluster_sim/fake_server.py
```diff
@@ -91,6 +91,7 @@ class FakeServer(utils.Persisted):
         'lnet_up': False,
         'stats': None,
         'packages': {},
+        'client_mounts': {},
```
chrisgearing: And this change

cluster-sim/cluster_sim/utils.py
```diff
@@ -88,26 +88,23 @@ class Persisted(object):
     filename = None
     default_state = {}
```
chrisgearing: Are any of the changes below this are required for this patch.

for server in self.servers.values():
   server.stop_corosync()
   server.stop_pacemaker()
   server.save()

Above is all that is required isnt it.

Adding in other changes just leads to unexpected changes in behaviour.

cluster-sim/cluster_sim/fake_server.py
tanabarr: shortens the duration of the sim tests with no adverse effects

cluster-sim/cluster_sim/fake_server.py
tanabarr: shortens the duration of the sim tests with no adverse effects

cluster-sim/cluster_sim/fake_server.py
tanabarr: this was required because when resetting state, at the beginning of one of the tests this key was expected

cluster-sim/cluster_sim/utils.py
tanabarr: unfortunately just resetting corosync and pacemaker state did not fix the issue. I've had to also ignore peers not in certain states during corosync configuration in order to work around this issue and to get the simulator tests passing reliably.

chrisgearing: If resetting corosync and pacemaker state didn't fix the issue then we did not understand the issue. Resetting more stuff feels like trying things in the hope something works.

The hypothesis is that the fake plugin returns corosync/pacemaker data when they are not virtually started. So stopping them (to simulate uninstall as clear_ha does) and making the code only return when they are started should fix it.

If that does not fix is the hypothesis is wrong and just doing more of the same seems strange to me.

cluster-sim/cluster_sim/fake_server.py
chrisgearing: But I thought this patch didn't have the rest code in it?

cluster-sim/cluster_sim/fake_server.py
chrisgearing: Then remove the # 20.

But tbh if you are trying to uncover a simulator issue, that is difficult to find, changing other timings are the same time seems to be a way to make the problem harder, not easier.

cluster-sim/cluster_sim/fake_server.py
chrisgearing: Then remove the # 20.

But tbh if you are trying to uncover a simulator issue, that is difficult to find, changing other timings are the same time seems to be a way to make the problem harder, not easier.

PR #52, opened 2017-05-24
Remove Downloading of npm during build.

jgrund:
We are currently using a bridged version of node
sans npm.

We were using this code to pull npm for builds, however
npm has been added to the builder, which obviates this.

Remove npm and the code that setup a new npm root.
Since we are staying current with node / npm there
is no need to jail a newer npm setup.


jgrund:
Landed in c92e59121243882b05d233bb8464528864ae8bc9

PR #53, opened 2017-05-24
HYD-7338 Fix exceptions due to missing config store directories

tanabarr:
During agent daemon start-up, the settings directory of the config
store sometimes cannot be found and results in an exception in the
chroma-agent damon log.

The explanation for this is that the chroma-agent service has been
started before agent registration (in agent_setup.py) has completed
(during which the config store directories are created).

This situation can occur due to a bug when checking if the service
is already running during chroma-agent registration. After the
fix is applied, if this situation occurs, a message should appear
in the agent daemon log indicating "service was running before
registration, stopping.".

  - stop agent service if it is already running at the beginning
    of the agent_setup.py 'register' function

### DIFF
--- a/chroma-agent/chroma_agent/action_plugins/agent_setup.py
+++ b/chroma-agent/chroma_agent/action_plugins/agent_setup.py
@@ -1,7 +1,7 @@
 #
 # INTEL CONFIDENTIAL
 #
-# Copyright 2013-2015 Intel Corporation All Rights Reserved.
+# Copyright 2013-2017 Intel Corporation All Rights Reserved.
 #
 # The source code contained or described herein and all documents related
 # to the source code ("Material") are owned by Intel Corporation or its
@@ -42,6 +42,7 @@ agent_service = ServiceControl.create('chroma-agent')
 
 
 def _service_is_running():
+    # returns True if running
     return agent_service.running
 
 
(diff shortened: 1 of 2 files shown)


chroma-agent/chroma_agent/action_plugins/agent_setup.py
```diff
@@ -1,7 +1,7 @@
 #
 # INTEL CONFIDENTIAL
 #
-# Copyright 2013-2015 Intel Corporation All Rights Reserved.
+# Copyright 2013-2017 Intel Corporation All Rights Reserved.
```
jgrund: Just an FYI that our copyright is going to change to:

// Copyright (c) 2017 Intel Corporation. All rights reserved.

So don't worry about updating the year for now.

chroma-agent/chroma_agent/action_plugins/agent_setup.py
```diff
@@ -42,6 +42,7 @@
 
 
 def _service_is_running():
+    # returns True if running
     return agent_service.running
```
jgrund: Is there an advantage to this being a method since it's just a property lookup?

jgrund:
Landed in: 49b39733ba67fc78aba137afdc03f5cf40202e16

PR #54, opened 2017-05-24
HYD-7337 Allow missing session objects in session table during terminate

tanabarr:
The chroma-agent.log displays the following trace back during SSI runs, they
result from a missing syslog key in the plug-in name to session object map when
running SessionTable.terminate('syslog') after the session/plug-in teardown
routine has been run.

The missing key should not cause an exception in the log because the session
will have already been successfully deactivated.

  - Except KeyError when deleting 'syslog' from _session map after session
    teardown


chroma-agent/chroma_agent/agent_client.py
```diff
@@ -296,7 +296,10 @@ def terminate(self, plugin_name):
         else:
             daemon_log.info("SessionTable.terminate %s/%s" % (plugin_name, session.id))
             session.teardown()
-            del self._sessions[plugin_name]
+            try:
+                del self._sessions[plugin_name]
+            except KeyError:
+                pass
```
jgrund: Do we want to log something here?

tanabarr: we could do, didn't feel it was necessary at the time. want me to add some logging?

jgrund:
Can we remove gerrit change id's?

chroma-agent/chroma_agent/agent_client.py
```diff
@@ -279,7 +279,10 @@ def terminate(self, plugin_name):
         else:
             daemon_log.info("SessionTable.terminate %s/%s" % (plugin_name, session.id))
             session.teardown()
-            del self._sessions[plugin_name]
+            try:
+                del self._sessions[plugin_name]
+            except KeyError:
+                daemon_log.warning("SessionTable.terminate session object already gone")
```
jgrund: It would be good to log the plugin_name here as well. I won't block for it though.

tanabarr: name is logged on line 280 which is not as good as doing it on the same line but should be able to work it out in the logs

tanabarr:
change-IDs added by pre-commit hook, I will investigate whether these are source versioned or just on my local system

chroma-agent/chroma_agent/agent_client.py
jgrund: Yeah, having all info in a single statement is best. + 280 only logs on info.

PR #76, opened 2017-06-05
Move tests for chroma_common code to chroma-common package

tanabarr:
Now chroma-common is a stand-alone package, move tests in agent and
manager test directories that exclusively test routines within
chroma-common to the chroma-common/tests directory. This step
requires that common test utilities become available in a stand-alone
package to be made accessible from other chroma top-level directories
(HYD-6745).

  - Move tests from chroma-agent/ and chroma-manager/ unit test dirs
  - Change relevant unit test classes to inherit from ImlUnitTestCase
  - Update relevant import paths
  - Update jenkins_steps/main for unit test job to run chroma-common
    unit tests


tanabarr:
this is a necessary step to move to being a stand-alone module, moving the unit tests into the package directory

jgrund:
Can we close this one?

PR #94, opened 2017-06-15
Add support for API key auth

jgrund:
As we move towards a distributed model, we need a way to authenticate against the api besides username / password combo.

 We currently proxy the session forward when needed, but this is sub-optimal.

This change is sub-optimal as well IMO (we should be using some sort of API gateway for auth and not concern ourselves with it in this service), but it enables further changes.

### DIFF
--- a/chroma-manager/chroma_core/lib/service_config.py
+++ b/chroma-manager/chroma_core/lib/service_config.py
@@ -29,6 +29,8 @@ from django.core.management import ManagementUtility
 from django.core.validators import validate_email
 from django.core.exceptions import ValidationError
 
+from tastypie.models import ApiKey
+
 from chroma_core.models.bundle import Bundle
 from chroma_core.services.http_agent.crypto import Crypto
 from chroma_core.models import ServerProfile, ServerProfilePackage, ServerProfileValidation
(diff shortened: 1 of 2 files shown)


chroma-manager/chroma_core/lib/service_config.py
```diff
@@ -571,6 +573,17 @@ def _setup_database(self, username, password, check_db_space):
...
+
+        try:
+            User.objects.get(username=API_USERNAME)
+            log.info("API user already created")
+        except User.DoesNotExist:
+            api_user = User.objects.create_superuser(API_USERNAME, "", User.objects.make_random_password())
+            api_user.groups.add(Group.objects.get(name='superusers'))
+            ApiKey.objects.get_or_create(user=api_user)
+            log.info("API user created")
+
```
petertix: Wondering if you want an "else" case for the try-except block to handle the case if the user create fails.
try:
    User.objects.get(username=API_USERNAME)
     log.info("API user already created")
except: User.DoesNotExist:
 +            api_user = User.objects.create_superuser(API_USERNAME, "", User.objects.make_random_password())
 +            api_user.groups.add(Group.objects.get(name='superusers'))
 +            ApiKey.objects.get_or_create(user=api_user)
 +            log.info("API user created")
   else:
       log.info('Could not create API user')

jgrund: I think we'd want the script to blow up and halt at that point. Having this user will become a hard requirement for other services.

PR #98, opened 2017-06-21
Create Stored Procedure for retrieving the API un/key combo

jgrund:
#94 added support for API key auth. However to retrieve the key and username, each service would need to construct a custom query.

Instead of that, add a stored procedure where anyone who has db credentials can issue a call.

There is still somewhat of a barrier in that a user needs db credentials, but this lowers the bar until we have a shared settings
store over a key value structure.


jgrund:
@tanabarr It has to be run. Once a dependency is added that uses this stored procedure, we can add a smoke test around that service + authing against the API.

PR #102, opened 2017-06-26
Bump tastypie to 0.9.16

jgrund:
0.9.11 only has options to do API key auth
with a querystring for GET.

0.9.12 has revamped auth code, and allows credentials to be sent via header as they should be.
However 0.9.12 has some other bugs, so we will bump to 0.9.16.

Even though this is listed as a patch change in semver, there are multiple api breaking changes.

They include:

  - `USE_TZ` is enforced at the API level, as a consequence tz's are dropped by default.
  - apply_limits is now a noop
  - `request` has been removed as a distinct param in many methods.
    We need to use `bundle.request` instead.
  - `override_urls` is deprecated, use prepend_urls instead.
  - Validation is moved later in the cycle. In addition, bundles
    are mutated at that point, meaning we need to restore
    earlier validation behavior.
  - PUTs now return a 200 instead of a 202 when returning data back.
  - Django auth limits seem to be verified now, where before they did not
    appear to be.
    An example is permissions missing for changing alerts, but users
    being allowed to dismiss anyway.

In addition to all of that, also removed some dead code that was trying to patch behave, but wouldn't apply.

### DIFF
--- a/chroma-manager/chroma_api/command.py
+++ b/chroma-manager/chroma_api/command.py
@@ -11,9 +11,9 @@ from tastypie.validation import Validation
 from tastypie.utils import trailing_slash
 from tastypie.api import url
 from tastypie import fields, http
+from tastypie.authorization import DjangoAuthorization
 
 from chroma_api.authentication import AnonymousAuthentication
-from chroma_api.authentication import PATCHSupportDjangoAuth
 from chroma_api.utils import custom_response
 from chroma_api.host import HostResource
 from chroma_api.chroma_model_resource import ChromaModelResource
@@ -22,6 +22,7 @@ from chroma_core.models import Command
 from chroma_core.models import SchedulingError
 from chroma_core.models import StepResult
 from long_polling_api import LongPollingAPI
+from chroma_api.validation_utils import validate
 
 
 class CommandValidation(Validation):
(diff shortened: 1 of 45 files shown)


chroma-manager/tests/framework/unit/jenkins_steps/main
```diff
@@ -27,48 +27,6 @@ cd $CHROMA_DIR/chroma-manager
 make requirements
 pip install -r requirements.txt
```
tanabarr: did you mention this removal of code in the issue comments? apologies if you did and I missed it

jgrund: I'll add it. This was patch could not be applied and was erroring out. We probably stopped needing it a while ago.

chroma-manager/chroma_api/utils.py
```diff
@@ -633,3 +652,14 @@ def _call_action(bulk_action_results, action, data, request, **kwargs):
                                   result)
 
     BulkActionResult = namedtuple('BulkActionResult', ['object', 'error', 'traceback'])
+
+class DateSerializer(Serializer):
+    """
+    Serializer to format datetimes in ISO 8601 but with timezone
+    offset.
+    """
+    def format_datetime(self, data):
```
johnsonw: So I guess this is the fix for the timezone issue we saw on Friday?

chroma-manager/chroma_api/command.py
```diff
@@ -109,15 +110,19 @@ def dismiss_all(self, request, **kwargs):
...
         for job in bundle.data['jobs']:
             # FIXME: HYD-1367: This is a hack to work around the inability of
             # the Job class to handle m2m references properly, serializing hosts
             # to a list of IDs understood by the HostListMixin class
             if 'hosts' in job['args']:
                 job_ids = []
                 for uri in job['args']['hosts']:
-                    job_ids.append(HostResource().get_via_uri(uri).id)
+                    job_ids.append(HostResource().get_via_uri(
+                        uri, bundle.request).id)
```
jgrund: I was using a python auto-formatting tool that did this, think it was autopep8.

Ended up reverting most of them, as it made the diff very noisy.

PR #113, opened 2017-07-02
Change permissions of all test_* files to non-executable so they are not missed by nose

tanabarr:
If a file is set as executable, it isn't automatically run by nose test runner.

@jgrund pointed out that we have a number of test_* tests that are not being run because they have one of the executable permission bits set. These should be set as non-executable (as should all python files in the project as we should always run with python explicitly, e.g. with an assigned interpreter, this would reduce potential bugs with different file permissions)


tanabarr:
TODO: fix permissions on all python project files

PR #177, opened 2017-07-26
print statement syntax fails intermittently in behave tests

tanabarr:
behave tests intermittently fail due to a strange Syntax error within error handling code

fix is to change the print statement to a print function


tanabarr:
the contents of this patch has been integrated into #180

jgrund:
Looks like: https://github.com/behave/behave/issues/287

PR #196, opened 2017-08-08
vagrant up fails on Windows as VBoxManage attachstorage --comment not implemented

tanabarr:
**DESCRIPTION**
The _vagrant up_ command fails on Windows because in Virtualbox (5.1.26), the _VBoxManage attachstorage_ command does not implement the _--comment_ option

The option does seem just to add description and can probably be removed (as per comment description):

--comment
Any description that you want to have stored with this medium (optional; for example, for an iSCSI target, "Big storage server downstairs"). This is purely descriptive and not needed for the medium to function correctly.

**SPECS**
Windows 8.1 Enterprise
VirtualBox 5.1.26
Vagrant 1.9.7
Cygwin 2.8.2(0.313/5/3) x86_64

**REFERENCE**
https://forums.virtualbox.org/viewtopic.php?f=7&t=34032
https://www.virtualbox.org/ticket/11846

This is a requirement of #201


tanabarr:
this patch is contained in and will be landed as part of #203

PR #199, opened 2017-08-09
Perform libvirt specific config only when it is the set vagrant provider 

tanabarr:
During create_cluster and destroy_cluster Makefile targets, a certain amount of configuration is libvirt specific, and doesn't apply if virtualbox is used as the vagrant provider.

Perform this specific configuration only if the guests setup by _vagrant up_ use libvirt as the provider.

This is a requirement of #201


tanabarr:
this patch is also contained as a committee included as part of #203

PR #207, opened 2017-08-11
Add fencing support for vagrant + vbox

### LINKED ISSUE #155: Add fencing support for vagrant + vbox
We currently have fencing support on some VM platforms using: https://github.com/intel-hpdd/intel-manager-for-lustre/blob/master/chroma-manager/scripts/fake_ipmi_kvm.py.

We need support for virtual box as well. We should be able to follow a similar approach, but using `fence_vbox` instead of `fence_virsh`

This work is listed as part of #201

jgrund:
Fixes #155.

We currently have fencing support on some VM platforms using: https://github.com/intel-hpdd/intel-manager-for-lustre/blob/master/chroma-manager/scripts/fake_ipmi_kvm.py.

We need support for virtual box as well. We should be able to follow a similar approach, but using fence_vbox instead of fence_virsh

This work is listed as part of #201 


chroma-agent/chroma_agent/lib/fence_agents.py
```diff
@@ -80,6 +80,33 @@ def delay_on():
...
+        else:
+            raise RuntimeError("Neither password nor identity_file were supplied")
+
+        super(fence_vbox, self).__init__(plug,
+                                          [agent, '-a', ipaddr, '-u', str(ipport), '-l', login, '-x'] + auth)
+
+    def on(self):
+        """Override super.on to wait 15 seconds then process as usual.
+
+        Real servers start slower then virtual ones do.  This simulates the production
```
tanabarr: sp: then->than

jgrund:
screen shot 2017-09-13 at 10 10 00 am

PR #251, opened 2017-08-30
Add device scanner to agent.

jgrund:
This patch adds the [device-scanner](https://github.com/intel-hpdd/device-scanner/) as an install time dependency to the agent.

In doing so, it replaces most of the imperative scanning that was being done for each device-scan.

In it's place, a persistent daemon is stood up and asynchronously handles uevents regarding relevant block devices.

Due to this strategy, we can realize some gains in the device scanning area.

Anecdotally, scan times went from ~33 seconds on a vagrant node to sub-second.

Additionally, SSI test times seem to have decreased by ~ 4 hours with this patch.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/251)

### DIFF
--- a/chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
+++ b/chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
@@ -70,12 +70,17 @@ def scanner_cmd(cmd):
     return json.loads(out)
 
 
+def getter(prop, default_value, x):
+    y = x.get(prop, default_value)
+    return y if y is not None else default_value
+
+
 def get_major_minor(x):
     return "%s:%s" % (x.get('MAJOR'), x.get('MINOR'))
 
 
 def as_device(x):
-    paths = sort_paths(x.get('PATHS', []))
+    paths = sort_paths(getter('PATHS', [], x))
     path = next(iter(paths), None)
 
     return {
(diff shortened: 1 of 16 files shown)


chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
```diff
@@ -4,98 +4,197 @@
...
-        retries = self.MAXRETRIES
-        while retries > 0:
-            try:
-                file_status = os.stat(path)
 
-                if path in self.non_existent_paths:
-                    self.non_existent_paths.discard(path)
-                    daemon_log.debug('New device started to respond %s' % path)
+def find_device_by_device_path(p, xs):
+    return next(iter([d for d in xs if d['device_path'] == p]), None)
```
tanabarr: this can be written: next((d for d in xs if d['device_path'] == p), None) to use generator

chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
```diff
@@ -4,98 +4,197 @@
...
+        # Never use 'w' in the built-in open() or it'll create
+        # a 0 length file where a
+        # device was removed!
+        fd = os.open("/dev/%s" % x['device_path'], os.O_WRONLY)
+    except OSError, e:
+        # EROFS: Device is read-only
+        # ENOENT: No such file or directory
+        # NO_MEDIA_ERRNO: No medium found
+        if e.errno in [errno.EROFS, errno.ENOENT, errno.NO_MEDIA_ERRNO]:
+            return False
```
tanabarr: what about when e.errno not in that list, don't you want to raise? do you want to use the device

jgrund: I think I can replace all of this with some udev rules.

PR #191, opened 2017-09-13
Maintain python2-iml-common minor versions as separate copr repos

tanabarr:
Unfortunately, COPR removes older versions of a package after 15 days, meaning we can only have one version of iml-common available. This is a problem because we will have different branches/patches using different versions in test and in production. 

It has been decided that we include the version within the name of the RPM deployed through the COPR repo for each new minor version of iml-common. Meaning different versions of iml-common will be listed in the COPR repo.

The resultant rpms would be of the form python2-iml-common1.1-1.1.x-1.el7.x86_64.rpm python2-iml-common1.2-1.2.x-1.el7.x86_64.rpm for examples. 

chroma-{agent,manager}.spec would then "Requires:" the API version of the package it wants e.g. Requires: python2-iml-common1.1

python requirements files will continue to pin on minor version

Pinning started at version 1.0

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/191)


brianjmurrell:
Reviewed 1 of 3 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/intel-hpdd/intel-manager-for-lustre/191)*

brianjmurrell:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/intel-hpdd/intel-manager-for-lustre/191#-:-KuOmOU0bzpcT7OR5Sm7:bnfp4nl)*

chroma-manager/requirements.dev
```diff
@@ -18,4 +18,4 @@ pyflakes
 pygraphviz
 sphinx==1.1.3
 werkzeug
-iml-common
+iml-common>=1.0,<1.1
```
jgrund: I think we should remove `requirements.dev` and only use a single set of rpm sourced deps.

tanabarr: I agree, given we go to the effort of creating a set of rpm deps, might as well use them for all testing

brianjmurrell: Ahhh.  Wait.  This dep is in the wrong place then!  It probably needs to go into https://github.com/intel-hpdd/intel-manager-for-lustre/blob/pin-iml-common/chroma-dependencies/Makefile.  The caveat being that that whole `chroma-dependencies/` subdir needs to go away entirely and the `requirements.txt` building that it assists in refactored to somewhere/thing else.

But that's where it is and how it is done currently, so that is where this needs to go.

brianjmurrell: > I think we should remove requirements.dev

So, these are (supposed to be) the deps that a "developer" would want (i.e. `pip`) installed in his "local" environment for doing development/testing and/or for "local source pool" type CI tests.

> and only use a single set of rpm sourced deps.

How does that work for developers on Windows and/or MacOS?

jgrund: They would use the VM setup for dev which many of us are already doing.

brianjmurrell:
:lgtm:

---

Reviewed 2 of 2 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/intel-hpdd/intel-manager-for-lustre/191#-:-KuZ2BLy3_E5yYFQrcHQ:bnfp4nl)*

PR #303, opened 2017-09-28
changes to address review comments from jgrund on #244

tanabarr:
separate PR in order to reduce delay in landing #244

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/303)

### DIFF
--- a/chroma-agent/chroma_agent/device_plugins/linux_components/zfs.py
+++ b/chroma-agent/chroma_agent/device_plugins/linux_components/zfs.py
@@ -27,13 +27,7 @@ strip_lines = partial(map, lambda x: x.strip())
 
 
 def write_to_store(key, value, filename=STORE_FILENAME):
-    data = {}
-
-    try:
-        with open(filename, 'r') as f:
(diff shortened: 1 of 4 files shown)


chroma-agent/chroma_agent/device_plugins/linux_components/zfs.py
```diff
@@ -27,13 +27,7 @@
 
 
 def write_to_store(key, value, filename=STORE_FILENAME):
-    data = {}
-
-    try:
-        with open(filename, 'r') as f:
-            data = json.loads(f.read())
```
jgrund: I'm wondering why we are duplicating the store location both within IML and iml-common. Ideally in modules you are encapsulating logic, if you don't you are actually making it harder to use a modular architecture than easier.

tanabarr: yes , this needs to be addressed, and will be in #290 + corresponding iml-common patch.

PR #306, opened 2017-09-29
Integrate device-scanner in IML agent

jgrund:
This patch integrates the device-scanner repo into the iml-agent. It is a replacement for the `BlockDevices` class and imperative lookup of devices.

Instead, a device map is held in a daemon, and updated whenever a udev event is received over a unix domain socket.

As a first pass, the BlockDevices class is updated to query state from the scanner so it is still polling (but quicker). The ultimate goal will be to move from a poll model to a push model which will necessitate further changes downstream.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/306)

### DIFF
--- a/chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
+++ b/chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
@@ -228,13 +228,8 @@ class BlockDevices(object):
         # matching MM to path back in a list.
 
         def build_paths(x):
-            out = []
-
-            for path in x['paths']:
-                if path.startswith(folder):
-                    out.append((x['major_minor'], path))
-
-            return out
+            return [(x['major_minor'], path) for path in x['paths']
+                    if path.startswith(folder)]
 
         return pipe(self.block_device_nodes.itervalues(),
                     cmapcat(build_paths), dict)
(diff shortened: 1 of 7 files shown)


chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
```diff
@@ -2,129 +2,219 @@
...
+        'parent': None
+    }
 
-    def __init__(self):
-        self.old_udev = None
-        self._major_minor_to_fstype = {}  # Build this map to retrieve fstype in _device_node
 
-        for blkid_dev in utils.BlkId().itervalues():
-            major_minor = self._dev_major_minor(blkid_dev['path'])
+def get_parent_path(p):
```
tanabarr: there is os.pardir for portability , but not much benefit in changing IMO

chroma-agent/chroma_agent/device_plugins/linux_components/block_devices.py
```diff
@@ -134,149 +224,25 @@ def composite_device_list(self, source_devices):
...
-        ndp.add_normalized_list(dev_nodes, by_path_nodes)
-        ndp.add_normalized_list(dev_nodes, by_id_nodes)
-        ndp.add_normalized_list(by_path_nodes, mapper_devs)
-        ndp.add_normalized_list(by_id_nodes, mapper_devs)
-
-        return block_device_nodes, node_block_devices
+        def build_paths(x):
+            out = []
 
+            for path in x['paths']:
```
tanabarr: return [(x['major_minor'], path) for path in x['paths'] if path.startswith(folder)]

jgrund: Done.

PR #309, opened 2017-10-02
Websockets blocked by invalid certs

jgrund:
Chrome has turned the security screws again, and we need to update our certs.

We are missing [SAN](https://geekflare.com/san-ssl-certificate/), and it appears even after adding it, we are still blocked by the self-signed cert.

As a first step, update our crypto generation to add a DNS name equivalent to the current CN.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/309)


johnsonw:
It turns out that the SAN is not causing the websocket issue here. Nevertheless, adding SAN to our certs is certainly a good addition. There is a bug in Chrome@61 that is causing the websockets to break within a web worker: https://bugs.chromium.org/p/chromium/issues/detail?id=760708. 

## Problem
> I think the specific problem is that WebSockets created in a Worker no longer have an  associated |frame_id|. As a result,  SSLManager::OnSSLCertificateSubresourceError() doesn't get passed a |frame_id| when it is called in WebSocketImpl::WebSocketEventHandler::OnSSLCertificateError(). As a result HandleSSLErrorOnUI() in //content/browser/ssl/ssl_manager.cc doesn't find a |web_contents| and  so cancels the request without ever looking up the cached user decision.

This issue exists in 62 and 63 as well but it sounds like a patch will be added to fix the issue:
> This is broken in latest Chrome dev(62.0.3202.9) and Canary(63.0.3208.0) as well, Any CL merged to M61 should be merged to M62 and M63.

PR #314, opened 2017-10-03
ZFS MMP support

jgrund:
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/314)


jgrund:
This is @tanabarr code that I pushed. 

As such I can't review it, but will leave an old-school 👍  here.

PR #317, opened 2017-10-04
Stonith operations unable to proceed due to missing -f flag on zpool imports

tanabarr:
When attempting to perform a stonith operation on a host, the target is unable to failover. 

This is because we've removed all `-f` flags during import. However, there seems to be a single case where we still need the `-f` and that's when a node has been fenced and the pool was not exported cleanly.

you can see the target unable to failover in the `corosync.log`:

```
Oct 04 09:51:05 [5766] oss2.lfs.local       lrmd:   notice: operation_finished:	fs-OST0000_c9e715_start_0:25186:stderr [ Error importing pool: 'Error (1) running 'zpool import oss1': '' 'cannot import 'oss1': pool was previously in use from another system. ]
Oct 04 09:51:05 [5766] oss2.lfs.local       lrmd:   notice: operation_finished:	fs-OST0000_c9e715_start_0:25186:stderr [ Last accessed by oss1.lfs.local (hostid=9ebf00b4) at Wed Oct  4 09:46:18 2017 ]
Oct 04 09:51:05 [5766] oss2.lfs.local       lrmd:   notice: operation_finished:	fs-OST0000_c9e715_start_0:25186:stderr [ The pool can be imported, use 'zpool import -f' to import the pool. ]
```

In this case, the output is indicating that a `zpool import -f` will import the pool, which it does.

We could possibly parse for this message during failover situations and run with `-f` in that specific case.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/317)


chroma-agent/chroma_agent/device_plugins/linux_components/zfs.py
```diff
@@ -30,7 +30,9 @@ def write_to_store(key, value, filename=ZFS_OBJECT_STORE_PATH):
     try:
         with open(filename, 'r') as f:
             data = json.loads(f.read())
-    except IOError:
+    except:
+        # not readable carry on
```
jgrund: Why is this not readable though?

chroma-agent/chroma_agent/device_plugins/linux_components/zfs.py
```diff
@@ -25,19 +25,26 @@
...
+                x = y
     except IOError:
+        # not readable carry on
+        daemon_log.info('write_to_store(): failed reading zfs data from %s.' % filename)
         pass
 
-    daemon_log.info('write_to_store(): writing zfs data to %s. key: %s' % (filename, key))
+    dataDict = json.loads(x)
+
+    daemon_log.info('write_to_store(): writing zfs dataDict to %s. key: %s' % (filename, key))
```
brianjmurrell: Is this really `info` or `debug`?

chroma-agent/chroma_agent/action_plugins/manage_targets.py
```diff
@@ -428,7 +428,17 @@ def mount_target(uuid, pacemaker_ha_operation):
     # This is called by the Target RA from corosync
     info = _get_target_config(uuid)
 
-    if agent_result_is_error(import_target(info['device_type'], info['bdev'])):
+    import_retries = 100
+    errored = True
+
+    for i in xrange(import_retries):
+        errored = agent_result_is_error(import_target(info['device_type'], info['bdev'], pacemaker_ha_operation))
```
brianjmurrell: Why does this need to be repeated (without any sleep between attempts even) up to 100(!) times to succeed?

PR #335, opened 2017-10-11
refactor zfs device scan to avoid race conditions

tanabarr:
Re-factor zfs full_scan so that potential for race conditions is minimised. After first discovering pools, don't rely on initial stale state and simply read from store info for those pools not available.

Cover the case where initially a pool is seen as imported but when it comes to processing that pool it is no longer imported, in this case the id is no longer available and the pool is retrieved by name from the store.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/335)


johnsonw:
I tested this patch in ZFS managed mode and was able to verify STONITH as well as failover / failback.

jgrund:
Tested with ZFSMO as well without issue.

PR #336, opened 2017-10-11
cleanup files after uninstall

tanabarr:
clean /var/lib/iml* on uninstall as store not getting truncated up now when agent stops/starts

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/336)


tanabarr:
this will probably be unnecessary as the store probably going to go away due to the work in https://github.com/intel-hpdd/intel-manager-for-lustre/issues/440

PR #342, opened 2017-10-12
Remove all forcing in device / pool setup / teardown

jgrund:
We currently use `-f` extensively for clearing devices. We should be able to run tests without having to force clear devices. In other words, we should either be in a state between where cleanup is possible without `-f` or we should know why we need `-f` and fix those issues.

This patch removes all `-f` from zpool creation / destruction / labelclear.

It also removes simultaneous cleanup of shared storage by multiple hosts; the first host will clear the shared storage and then the other hosts, will run a `partprobe` followed by a `udevadm settle`.

It also replaces `dd`ing the start of disks with `wipefs` to clear all labels.
Finally, it adds a `zpool labelclear` clearing devices.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/342)

### DIFF
--- a/chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
+++ b/chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
@@ -1042,9 +1042,13 @@ class ApiTestCaseWithTestReset(UtilityTestCase):
             if lustre_device['backend_filesystem'] == 'ldiskfs':
                 linux_device = TestBlockDevice('linux', first_test_server['device_paths'][lustre_device['path_index']])
 
-                self.execute_simultaneous_commands(linux_device.destroy_commands,
+                self.execute_commands(linux_device.destroy_commands,
+                                        first_test_server['fqdn'],
+                                        'clear block device %s' % linux_device)
+
+                self.execute_simultaneous_commands(['partprobe', 'udevadm settle'],
                                                    [server['fqdn'] for server in test_servers],
-                                                   'clear block device %s' % linux_device)
+                                                   'sync partitions')
 
     @property
     def quick_setup(self):
--- a/chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
+++ b/chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
@@ -157,10 +157,7 @@ class ApiTestCaseWithTestReset(UtilityTestCase):
                 self.cleanup_linux_devices(self.TEST_SERVERS)
 
                 # cleanup zfs pools
-                self.cleanup_zfs_pools(self.config_servers,
-                                       self.CZP_EXPORTPOOLS | (self.CZP_RECREATEZPOOLS if config.get('new_zpools_each_test', False) else self.CZP_REMOVEDATASETS),
-                                       None,
-                                       True)
+                self.cleanup_zpools()
 
             # Enable agent debugging
             self.remote_operations.enable_agent_debug(self.TEST_SERVERS)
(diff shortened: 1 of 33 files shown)


chroma-manager/tests/integration/utils/test_blockdevices/test_blockdevice_lvm.py
```diff
@@ -39,7 +39,7 @@ def device_path(self):
 
     @classmethod
     def clear_device_commands(cls, device_paths):
-        return ["if vgdisplay %s; then vgremove -f %s; else exit 0; fi" % (TestBlockDeviceLvm('lvm', device_path).vg_name,
+        return ["if vgdisplay %s; then vgremove %s; else exit 0; fi" % (TestBlockDeviceLvm('lvm', device_path).vg_name,
```
brianjmurrell: I think removing the `-f` from `vgremove` will make it want to prompt for confirmation.

jgrund: It did, but it appeared to mention there were still some lv's.

I added a lvremove before that, so that should remove the confirmation.

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
```diff
@@ -1025,26 +1025,30 @@ def dataset_match(datasets, device_path):
...
+            return 'wipefs -a {}'.format(x)
+
+        device_paths = [
+            get_device_path(x['path_index']) for x in config['lustre_devices']
+            if x['backend_filesystem'] == 'ldiskfs'
+        ]
+
+        desc = 'clear block devices {}'.format(', '.join(device_paths))
+        wipe_commands = map(cleanup_str, device_paths)
+        self.execute_commands(wipe_commands, first_test_server['fqdn'], desc)
```
brianjmurrell: What if the disks aren't fully meshed such as they could/would be if they were only dual-ported?

chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
```diff
@@ -152,8 +153,7 @@ def create_lustre_filesystem_for_test(self):
                                      'mgt',
                                      self.fsname,
                                      None,
-                                     ['--reformat',
-                                      '--mdt' if self.combined_mgt_mdt else '',
+                                     ['--mdt' if self.combined_mgt_mdt else '',
```
brianjmurrell: Why are we not `--reformat`ting the MGT?

chroma-manager/tests/integration/utils/test_blockdevices/test_blockdevice_lvm.py
```diff
@@ -20,27 +20,40 @@ def preferred_fstype(self):
...
 
     @property
     def device_path(self):
         return "/dev/%s/%s" % (self.vg_name, self.lv_name)
 
     @classmethod
     def clear_device_commands(cls, device_paths):
-        return ["if vgdisplay %s; then vgremove -f %s; else exit 0; fi" % (TestBlockDeviceLvm('lvm', device_path).vg_name,
-                                                                           TestBlockDeviceLvm('lvm', device_path).vg_name) for device_path in device_paths]
+        lv_format = [
```
brianjmurrell: These aren't really _format_ commands.  Can we change the variable name to be more accurate about what they really are, which is _destroy_, or _unconfigure_, or some such.

chroma-manager/tests/integration/utils/test_blockdevices/test_blockdevice_lvm.py
```diff
@@ -20,27 +20,40 @@ def preferred_fstype(self):
...
+            "if lvdisplay /dev/{0}/{1}; then lvchange -an /dev/{0}/{1} && lvremove /dev/{0}/{1}; else exit 0; fi".format(
+                TestBlockDeviceLvm('lvm', device_path).vg_name,
+                TestBlockDeviceLvm('lvm', device_path).lv_name)
+            for device_path in device_paths
+        ]
+        vg_format = [
```
brianjmurrell: Ditto.

chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
```diff
@@ -107,8 +107,8 @@ def _clear_current_target_devices(self):
 
         for server in config['lustre_servers']:
             self.remote_command(server['address'],
-                                'partprobe; udevadm settle',
-                                expected_return_code=None)
+                                'reboot',
```
brianjmurrell: It seems drastic to need a reboot just to remedy the MGT not mounting.  What kind of failure was happening on mount?

chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
jgrund: Yes.

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
jgrund: Will have to refactor this further when that becomes the case. Right now these disks are shared across all hosts.

chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
```diff
@@ -203,73 +208,60 @@ def get_unused_device(self, server_name):
     def umount_devices(self, server_name):
         lustre_server = self.get_lustre_server_by_name(server_name)
         for device in lustre_server['device_paths']:
-            self.remote_command(
-                server_name,
-                "if mount | grep %s; then umount %s; fi;" % (device, device))
+            self.remote_command(server_name,
+                                "if mount | grep %s; then umount %s; fi;" %
```
brianjmurrell: This is a (common mistake and) dangerous pattern match.  When looking for a string in a file, it is usually best to delimit the string if possible to avoid substring matches.  In the case of matching against `mount` (or `/proc/mounts`), a common false positive is to be `grep`ping for `/dev/sda` and to match on `/dev/sda1`.

In this case, you probably want to add a space delimiter before and after the `%s`, since you are changing this line of code anyway.

chroma-manager/tests/integration/utils/test_blockdevices/test_blockdevice_lvm.py
jgrund: Sure.

chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
jgrund: This isn't new code, I just reformatted it.

chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
```diff
@@ -203,73 +208,60 @@ def get_unused_device(self, server_name):
...
+                                (device, device))
 
-        self.remote_command(
-            server_name,
-            "sed -i '/lustre/d' /etc/fstab")
+        self.remote_command(server_name, "sed -i '/lustre/d' /etc/fstab")
```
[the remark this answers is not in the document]
jgrund: This isn't new code, I just reformatted it.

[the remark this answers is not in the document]
brianjmurrell: I realise that, but if you are going to touch the lines of code, you may as well fix the bugs in them in an effort to make the code more solid.

chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
brianjmurrell: Right.  As before, if the code deserves reformatting, it at least deserves having obvious bugs fixed also.

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
brianjmurrell: The `config` currently tells you which disks are available on which nodes.  Each `lustre_servers` has a list of disks it's connected to.

jgrund: Indeed, but the test config is currently wired to list devices by backing fs type + idx anyway, so there is a baked in presumption that it's shared between all hosts:

```
[
  {
    "backend_filesystem": "zfs",
    "path_index": 0
  },
  {
    "backend_filesystem": "ldiskfs",
    "path_index": 1
  },
  {
    "backend_filesystem": "zfs",
    "path_index": 2
  },
  {
    "backend_filesystem": "ldiskfs",
    "path_index": 3
  },
  {
    "backend_filesystem": "zfs",
    "path_index": 4
  }
]
```

chroma-manager/tests/integration/shared_storage_configuration/test_power_control.py
```diff
@@ -9,10 +9,16 @@ class ChromaPowerControlTestCase(ChromaIntegrationTestCase):
 
     def setUp(self):
         super(ChromaPowerControlTestCase, self).setUp()
+        
+        # Even though the tests only need 1 server, we need to add a server
```
brianjmurrell: Isn't `self.TEST_SERVERS` already (i.e at lesat all four storage server in) `config['lustre_servers']` as defined `class ApiTestCaseWithTestReset`?

chroma-manager/tests/integration/shared_storage_configuration/test_power_control.py
```diff
@@ -9,10 +9,16 @@ class ChromaPowerControlTestCase(ChromaIntegrationTestCase):
...
+        # and its HA peer in order to ensure that the peer doesn't send
+        # outdated CIB data over. There is an assumption here that the
+        # servers listed in the configuration are ordered by HA peer groups.
+        servers = config['lustre_servers'][0:2]
```
brianjmurrell: Why the switch form `self.TEST_SERVERS` to using `config['lustre_servers']` directly rather than redefining `self.TEST_SERVERS`?

I have a feeling that inconsistent use (or lack thereof) of `self.TEST_SERVERS` is part of our inconsistent environment problems.

jgrund: > I have a feeling that inconsistent use (or lack thereof) of self.TEST_SERVERS is part of our inconsistent environment problems.

Redefining: `self.TEST_SERVERS` is what causes inconsistent setup and teardown. By not redefining it explicitly, we ensure we do the full setup / teardown.

chroma-manager/tests/integration/shared_storage_configuration/test_power_control.py
jgrund: Yes. But I'd rather not rely on superclass properties.

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
```diff
@@ -887,7 +887,7 @@ def zfs_devices_exist(cls):
         return any(lustre_device['backend_filesystem'] == 'zfs' for lustre_device in config['lustre_devices'])
 
     def create_zpools(self):
-        xs = config['lustre_servers']
+        xs = config['lustre_servers'][:4]
```
brianjmurrell: This is starting to get very specific to our particular test clusters.

While I can understand the desire to get tests running reliably on those particular test clusters, I think we want to avoid creating a nightmare of a debugging problem for some future test cluster (ours or somebody else's) that does not conform to our current specific configuration.

So rather than confining to a list of servers can we make the list all excluding the server (type?)s that you don't want to include?

chroma-manager/tests/integration/shared_storage_configuration/test_power_control.py
brianjmurrell: > Redefining: self.TEST_SERVERS is what causes inconsistent setup and teardown. By not redefining it explicitly, we ensure we do the full setup / teardown.

Well, only because we defer the actions of tearing down until the next test's setUp() by which time TEST_SERVERS has been redefined to be for that next test.  Maybe we just need to bite off on cleaning up after tests actually after the test runs rather than waiting until the next test starts to do the clean up.

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
jgrund: Unfortunately, we are already in nightmare debugging territory with these tests. 

To be clear, my goal here is to get these tests running with some semblance of reliability; it is not to carry this work forward any further than it needs to. There are a number of abstractions we should be making in totality and they go far beyond the scope of this work.

For instance, the config should not operated on directly in the general case. Instead there should be functions that interface with it in a *read only* fashion (we currently mutate the config during our first test).

chroma-manager/tests/integration/shared_storage_configuration/test_power_control.py
jgrund: > Maybe we just need to bite off on cleaning up after tests actually after the test runs rather than waiting until the next test starts to do the clean up.

We should do teardown operations in teardown, yes. But again, the goal here is to get these tests operating reliably, it's not to refactor more of the test framework than I need to.

The work to refactor more of the test framework should happen in: https://github.com/intel-hpdd/intel-manager-for-lustre/issues/327

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
```diff
@@ -891,160 +886,115 @@ def linux_devices_exist(cls):
...
+            self.remote_operations.reset_server(
+                x) for x in fqdns
+        ]
+
+        [
+            self.remote_operations.await_server_boot(
+                x) for x in fqdns
+        ]
+
+    def cleanup_linux_devices(self, test_servers):
```
tanabarr: at some point it would be nice if these were shared between efs & ssi so we use consistent cleanup methods

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
```diff
@@ -891,160 +886,115 @@ def linux_devices_exist(cls):
...
 
-        for server in test_servers:
-            for lustre_device in config['lustre_devices']:
-                if lustre_device['backend_filesystem'] == 'zfs':
-                    zfs_device = TestBlockDevice('zfs', first_test_server['zpool_device_paths'][lustre_device['path_index']])
+        [
+            self.execute_simultaneous_commands(
+                x.clear_device_commands([x.device_path]),
+                fqdns,
+                'destroy zpool %s' % x,
```
tanabarr: will this recursively destroy datasets?

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
jgrund: Agreed.

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
jgrund: Yes

chroma-manager/tests/integration/utils/test_blockdevices/test_blockdevice_lvm.py
jgrund: Fixed.

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
brianjmurrell: Can we open a ticket for this assumption of fully-meshed disk connectivity? At least to point the first person who comes to us with a configuration that isn't fully-meshed at.

chroma-manager/tests/integration/shared_storage_configuration/test_filesystem_same_name.py
```diff
@@ -73,18 +72,47 @@ def create_for_mgs(name, hosts_to_use, reformat=False):
...
+
+        for pool in pools:
+            self.execute_simultaneous_commands(
+                ['zpool export %s' % pool],
+                fqdns,
+                'export pool %s' % pool,
+                expected_return_code=None)
+
+        self.remote_operations.start_agents(s['address']
+                                            for s in self.TEST_SERVERS[:4])
```
brianjmurrell: A ticket for this assumption that there are only 4 lustre servers would be good too i think.  Just to keep track of the technical debt.

jgrund: Will do.

jgrund: Opened: https://github.com/intel-hpdd/intel-manager-for-lustre/issues/352 as a general issue on SSI / EFS cleanup (feel free to add to it).

chroma-manager/tests/integration/core/api_testcase_with_test_reset.py
jgrund: Opened: https://github.com/intel-hpdd/intel-manager-for-lustre/issues/352 as a general issue on SSI / EFS cleanup (feel free to add to it).

PR #347, opened 2017-10-16
Recreate failing log-sort test case

petertix:
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/347)


jgrund:
Dup of #275

PR #354, opened 2017-10-20
PoC i18n for cluster-sim (passed build & simplest test)

jn0:
This patch makes possible to use translations for any user visible messages
in cluster-sim python code.

Common python locale/gettext facilities were involved.

Unicode is a must here (cyrillics has about a dosen of possible encodings,
which you don't want to be aware of).

One may check deployed setup by running
    LANG=ru_RU.UTF-8 python cluster-sim/cluster_sim/i18n.py
and look for translated output.

Added:
- cluster-sim/cluster_sim/i18n.py -- 1st (successful) attempt to implement
- cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po -- translations
(partial, about 1/3)

Patched to provide build for i18n/l10n (use target "translations" and update
LOCALE_LANGS var):
- cluster-sim/Makefile

Patched to include i18n/l10n into the build:
- cluster-sim/MANIFEST.in
- cluster-sim/chroma-cluster-sim.spec

Patched to employ the _() for messages:
- cluster-sim/cluster_sim/benchmark.py
- cluster-sim/cluster_sim/cli.py
- cluster-sim/cluster_sim/fake_action_plugins.py
- cluster-sim/cluster_sim/fake_client.py
- cluster-sim/cluster_sim/fake_cluster.py
- cluster-sim/cluster_sim/fake_controller.py
- cluster-sim/cluster_sim/fake_device_plugins.py
- cluster-sim/cluster_sim/fake_devices.py
- cluster-sim/cluster_sim/fake_hsm_coordinator.py
- cluster-sim/cluster_sim/fake_hsm_copytool.py
- cluster-sim/cluster_sim/fake_power_control.py
- cluster-sim/cluster_sim/fake_server.py
- cluster-sim/cluster_sim/queue_monitor.py
- cluster-sim/cluster_sim/simulator.py
- cluster-sim/cluster_sim/utils.py

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/354)

### DIFF
--- a/cluster-sim/cluster_sim/fake_power_control.py
+++ b/cluster-sim/cluster_sim/fake_power_control.py
@@ -2,6 +2,7 @@
 # Use of this source code is governed by a MIT-style
 # license that can be found in the LICENSE file.
 
+from cluster_sim.i18n import _
 
 import glob
 import json
@@ -36,7 +37,7 @@ OUTLET_CYCLE_TIME = 3
 
 class PDUSimulatorTcpHandler(SocketServer.BaseRequestHandler):
     def handle(self):
-        log.debug("Handling PDU request from %s:%s" % self.client_address)
+        log.debug(_("Handling PDU request from %s:%s") % self.client_address)
         self.server.pdu_simulator.handle_client(self.request,
                                                 self.client_address)
 
@@ -53,7 +54,7 @@ class PDUSimulatorTcpServer(SocketServer.TCPServer):
 
 class PDUSimulatorServer(ExceptionCatchingThread):
     def __init__(self, pdu):
-        log.info("Creating PDU server for %s on %s:%s" %
+        log.info(_("Creating PDU server for %s on %s:%s") %
                  (pdu.name, pdu.address, pdu.port))
 
         self.server = PDUSimulatorTcpServer((pdu.address, pdu.port),
@@ -98,7 +99,7 @@ class PDUSimulator(Persisted):
             self.save()
 
     def handle_client(self, sock, address):
-        log.info("%s: received connection from %s:%s" % (self.__class__.__name__,
+        log.info(_("%s: received connection from %s:%s") % (self.__class__.__name__,
                                                          address[0], address[1]))
         fd = sock.makefile()
         fd.write("NOT IMPLEMENTED\r\n")
@@ -130,7 +131,7 @@ class PDUSimulator(Persisted):
     def toggle_outlet(self, outlet, state):
         with self._lock:
             self.state['outlets'][outlet] = state
-            log.info("POWER: Toggled %s:%s to %s" % (self.name, outlet,
+            log.info(_("POWER: Toggled %s:%s to %s") % (self.name, outlet,
                                                      self.outlet_state(outlet)))
             self.save()
 
@@ -154,7 +155,7 @@ class PDUSimulator(Persisted):
 
 class APC79xxSimulator(PDUSimulator):
     def handle_client(self, sock, address):
-        log.info("%s: received connection from %s:%s" % (self.__class__.__name__,
+        log.info(_("%s: received connection from %s:%s") % (self.__class__.__name__,
                                                          address[0], address[1]))
         self.client_address = address
         self.socket = sock
@@ -163,7 +164,7 @@ class APC79xxSimulator(PDUSimulator):
         try:
             self.login()
         except socket.error:
-            log.info("%s: client %s:%s disconnected" % (self.__class__.__name__,
+            log.info(_("%s: client %s:%s disconnected") % (self.__class__.__name__,
                                                         address[0], address[1]))
 
     def navigation_prompt(self, prompt, forward_choices, forward, backward, banner=None):
@@ -215,7 +216,7 @@ class APC79xxSimulator(PDUSimulator):
                     caller = inspect.stack()[1][3]
 
                     decoded_response = [ord(c) for c in response]
-                    log.error("Unhandled response in %s: '%s' (%s)" %
+                    log.error(_("Unhandled response in %s: '%s' (%s)") %
                               (caller, response, decoded_response))
 
                     self.fd.write("\r\n%s" % prompt)
@@ -242,16 +243,16 @@ class APC79xxSimulator(PDUSimulator):
         # telnet control characters for logging.
         username = self._strip_controls(self.fd.readline().rstrip())
         self.fd.readline()
-        log.debug("Received username: %s" % username)
+        log.debug(_("Received username: %s") % username)
         self.fd.write("\rPassword : ")
         self.fd.flush()
         password = self.fd.readline().rstrip()
         self.fd.readline()
-        log.debug("Received password: %s" % password)
+        log.debug(_("Received password: %s") % password)
         self.control_console()
 
     def control_console(self, *args):
-        banner = """\n\n
+        banner = _("""\n\n
 American Power Conversion               Network Management Card AOS      v3.7.3
 (c) Copyright 2009 All Rights Reserved  Rack PDU APP                     v3.7.3
 -------------------------------------------------------------------------------
@@ -261,8 +262,8 @@ Location  : Unknown                                   User : Administrator
 Up Time   : 7 Days 15 Hours 31 Minutes                Stat : P+ N+ A+
 
 Switched Rack PDU: Communication Established
-"""
-        prompt = """
+""")
+        prompt = _("""
 ------- Control Console -------------------------------------------------------
 
      1- Device Manager
@@ -271,12 +272,12 @@ Switched Rack PDU: Communication Established
      4- Logout
 
      <ESC>- Main Menu, <ENTER>- Refresh, <CTRL-L>- Event Log
-> """
+> """)
         self.navigation_prompt(prompt, "1", self.device_manager,
                                self.control_console, banner)
 
     def device_manager(self, *args):
-        prompt = """
+        prompt = _("""
 ------- Device Manager --------------------------------------------------------
 
      1- Phase Management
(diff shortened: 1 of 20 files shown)


jn0:
*[cluster-sim/Makefile, line 110 at r1](https://reviewable.io:443/reviews/intel-hpdd/intel-manager-for-lustre/354#-KwtQsJX0qJVeFxmtZ3o:-KwtQsJX0qJVeFxmtZ3p:b-eqp2j6) ([raw file](https://github.com/intel-hpdd/intel-manager-for-lustre/blob/e638e553391680009f5716762a4d00dea5be640c/cluster-sim/Makefile#L110)):*
> ```Makefile
> 
> $(LOCALE_POT): $(PYGETTEXTR_PY) $(LOCALE_DIR)
> 	$(PYGETTEXTR_PY) -v -d $(LOCALE_DOM) -o $@ cluster_sim/*.py
> ```

I think, I have to add
```
    --no-location
        Do not write filename/lineno location comments.
```
option to not confuse the `diff` on minor changes...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/intel-hpdd/intel-manager-for-lustre/354)*

cluster-sim/cluster_sim/fake_power_control.py
```diff
@@ -242,16 +243,16 @@ def login(self):
...
         self.fd.flush()
         password = self.fd.readline().rstrip()
         self.fd.readline()
-        log.debug("Received password: %s" % password)
+        log.debug(_("Received password: %s") % password)
         self.control_console()
 
     def control_console(self, *args):
-        banner = """\n\n
+        banner = _("""\n\n
```
brianjmurrell: This the UI output of a simulated APC PDU.  Do you know that this PDU supports multiple languages and if so, that the translation is accurate?

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
```diff
@@ -0,0 +1,966 @@
...
+"American Power Conversion               Network Management Card AOS      v3.7.3\n"
+"(c) Copyright 2009 All Rights Reserved  Rack PDU APP                     v3.7.3\n"
+"-------------------------------------------------------------------------------\n"
+"Name      : RackPDU                                   Date : 02/13/2013\n"
+"Contact   : Unknown                                   Time : 10:49:37\n"
+"Location  : Unknown                                   User : Administrator\n"
+"Up Time   : 7 Days 15 Hours 31 Minutes                Stat : P+ N+ A+\n"
+"\n"
+"Switched Rack PDU: Communication Established\n"
+msgstr ""
```
brianjmurrell: Why is the `msgstr` empty here?  Will this result in an empty string being assigned to `banner` at [fake_power_control.py:255](https://github.com/intel-hpdd/intel-manager-for-lustre/pull/354/files#diff-d1460e059877e95da2e0c3bee77682caR255)?

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
```diff
@@ -0,0 +1,966 @@
...
+"        The key line expected by fence\_apc is the next one:\n"
+"\n"
+"        Enter 'YES' to continue or <ENTER> to cancel : "
+msgstr ""
+"\n"
+"        -----------------------------------------------------------------------\n"
+"        Я слишком ленив, чтобы продираться через всевозможные перестановки.\n"
+"        Нужная строчка, которую ждёт fence\_apc, вот эта:\n"
+"\n"
+"        Введите 'YES', чтобы продолжить или <ENTER>, чтобы отменить: "
```
brianjmurrell: How do we know that all of the above translations are accurate to the translations on the actual PDU being simulated.

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
```diff
@@ -0,0 +1,966 @@
...
+msgid "stopping %s in poweroff\_hook"
+msgstr "останавливаю %s в poweroff\_hook"
+
+#: fake\_power\_control.py:563
+msgid "starting %s in poweron\_hook"
+msgstr "запускаю %s в poweron\_hook"
+
+#: fake\_power\_control.py:567
+msgid "starting server for %s"
+msgstr ""
```
brianjmurrell: Does the empty `msgstr` here (and below) result in an empty string being logged?  Or even an exception due to the arguments not having `%s`(s) to substitute into?

cluster-sim/cluster_sim/fake_power_control.py
jn0: I've used some PDU/UPS from APC but never seen anything other than en_US there :)
Is there any reference to their localized manuals?

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
jn0: Perhaps, it's a typo or just the current end-of-translation (I've translated about 1/3 so far).

cluster-sim/cluster_sim/fake_power_control.py
brianjmurrell: So, until we can verify that APC PDUs are localized we should probably not try to translate the simulated UI, yes?

Additionally, we probably don't want to be testing non-localized IML code against even a simulated localized PDU i think.

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
jn0: The only way to ensure it is to find the relevant APC manual for respective market and grab the text from there... Would it be a copyright violation, BTW?

Or it may be left intact (no translation).
Or provide it "as is".
Or provide a mean to select "enforced translation" at command line or config...

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
jn0: It should be the current "end of translation".
As I noted in the commit msg, I've translated about 1/3 of the file.
I consider this commit as a PoC to make sure I do it in an acceptable way.

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
jn0: BTW, I ain't sure, if it's a real device message... It sounds way too jokingly to be hardwired into equipment...

cluster-sim/cluster_sim/fake_power_control.py
jn0: Heh... There *is* a manpage in russian...
http://www.apc.com/salestools/PMAR-8P9G2C/PMAR-8P9G2C_R2_RU.pdf

I'll check it out.

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
brianjmurrell: Or we just don't translate the simulatd PDU interface and require that (until such questions can be answered, and translations updated accurately) IML users set their PDU interface to be default English locale.

The simulated PDU interface shouldn't be translated until the IML code that interfaces with it has translations is localized in any case.

cluster-sim/cluster_sim/fake_power_control.py
jn0: No way.
It's untranslated.
What we have in russian manual:
image
versus the original one:
image

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
brianjmurrell: I'm not a localization expert by any stretch of the imagination so I probably don't personally have much of an opinion on how it's done.  Other than the notes I have made here already, I don't have a problem with the implementation.

One thing that might be worth considering though is if each package should have it's own set of `.po` files or there should be a single (optional even -- if possible) _localization_ package for all of IML.  That could be a bridge to cross in the future though.

Ultimately, I don't think you finishing the rest of the translations in the .po files is going to be wasted work even if we do decide to restructure the packaging somewhat.

One thing I would say though is that the cluster-simulator is probably the item least in need of localization.  I would think there is more value in localizing IML-proper.

cluster-sim/cluster_sim/fake_power_control.py
brianjmurrell: So, we should probably remove the localization from the PDU interface simulation as well as avoiding localizing any IML code that interacts with the APC PDUs.

jn0: Yes, I're right.
And there should be comments for translators to leave things as they are where appropriate.

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
jn0: I've started off with the most simple-to-localize (from my PoV) chunk of code :)
It's pure python without django (which has its own set of i18n/l10n rules), where my knowledge is quite limited.

If you have no major contras on *how* it was done, I'll, probably, finish it up and then move onto the other parts.

As of separation... It's quite common to have separate language packs.
But for the routine, one will have to update automatically generated `.pot` file *within* the package anyway, while hand written `.po` files can live separately for sure.
As soon, as an IML langpack repo become available, they can be moved there.

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
jn0: ok

cluster-sim/cluster_sim/locale/ru/LC_MESSAGE/cluster_sim.po
brianjmurrell: I have no major issues at this point.  I think it's worthwhile continuing.  I have no issues with the way the translations are being imported in python.  Even if we do some reorganisation in the future, the value of the actual translations and how the python code utilises them files remains.

cluster-sim/cluster_sim/fake_power_control.py
brianjmurrell: :+1:

brianjmurrell:
@jn0 Let me know when you think this work is complete and I will push it through testing.

jgrund:
@jn0 We are removing the cluster-sim in IML 4.1. I am going to close this, but we would appreciate any i18n efforts on the manager / agent repos. A really good place to start is https://github.com/intel-hpdd/help/blob/master/help.json where we have extracted a number of strings we use throughout the product. Once they have been converted into .po files we can consume them easily where needed.

PR #358, opened 2017-10-23
zpool parsing bug when scrub in status

tanabarr:
It appears that there is a bug in `zpool import / status` parsing when a scrub has been performed (or any value matching `(\w+):[^/]`).

a completed scrub will contain a timestamp which will match the regex:

```
re.split('(\w+):[^/]', 'scrub repaired 0B in 0h0m with 0 errors on Sun Oct 22 15:25:35 2017')
Out[2]:
['scrub repaired 0B in 0h0m with 0 errors on Sun Oct 22 ',
 '15',
 '',
 '5',
 '5 2017']
```

Ideally we could use a parser-combinator for this output (which would be sensitive to point of match ), but for now we should avoid splitting on values with ':'

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/358)

### DIFF
--- a/chroma-agent/chroma_agent/device_plugins/linux_components/zfs.py
+++ b/chroma-agent/chroma_agent/device_plugins/linux_components/zfs.py
@@ -150,10 +150,10 @@ def get_zpools(active=True):
         return []
 
     transform_pools = compose(_parse_line,
-                              lambda x: clean_list(re.split('(\w+):[^/]', x)),
-                              lambda x: 'pool: ' + x)
+                              lambda x: clean_list(re.split(r'\n\s*(\w+):\s', x)),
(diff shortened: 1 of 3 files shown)


johnsonw:
I just tried this patch and added 4 monitored only servers. I seem to get an error when adding the servers:
image

Also, I cannot modprobe lnet:
```
root@mds1 ~]# systemctl stop firewalld; systemctl disable firewalld
[root@mds1 ~]#     systemctl start ntpd
[root@mds1 ~]#     echo 'options lnet networks=tcp0(enp0s9)' > /etc/modprobe.d/lustre.conf
[root@mds1 ~]#     modprobe lnet
```

PR #131, opened 2017-11-09
Consume device-scanner ZED changes in agent

tanabarr:
Now that `device-scanner` contains ZED updates, we can consume those changes in the agent.

This ticket regards processing the output of the device scanner (which contains a map of inactive and active zpools) from within the zfs component of the Linux agent device plug-in. 

The end result should be that the agent no longer imports and exports zpools during device scans and instead uses information from the device scanner and through commands issued via libzfs bindings.

Currently the following zfs details are not provided through the device scanner but are needed during agent device scans:

- [ ] zvols have yet to be added to device-scanner and therefore agent output https://github.com/intel-hpdd/intel-manager-for-lustre/issues/400
- [ ] size of zfs objects - as there is no block device returned from the scanner for the top level zpool it means that probably need to issue command through libzfs to retrieve the size of zpools and datasets
- [ ] Initial state of active and inactive zpools and constituent zvols and datasets, including properties should be discovered using calls to zfs library bindings.
- [ ] Drive major minor numbers of drives that have been added to a zpool should be discovered using calls to zfs library bindings.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/131)


tanabarr:
looks like the current method of retrieving zpool drive mms doesn't take into account all partitions (part9 specifically)

PR #403, opened 2017-12-02
Map hosts that can access shared storage to ZFS objects discovered on other hosts 

tanabarr:
*SITUATION*

Using _device-scanner_ output to detect ZFS objects in the agent _linux_ device plug-in, _zpools_ imported on other hosts but on shared storage will not be currently reported.

*MEDIATION*

Create `VolumeNode` and related resource entries for zfs objects on a given host when the relevant agent is reporting an underlying block device that has been previously identified as containing _zpools/datasets/zvols_ on another agent/host. This logic will be implemented either in the manager component of the device plug-in or in the resource manager.

*RATIONALE*

Given ZFS objects exist atop standard block devices, it is a valid assumption that any host that can access the given underlying block device should also be able to access extant hosted ZfsPools.

*IMPLEMENTATION*

Use device serial numbers, which are set on shared devices, to identify the underlying _ScsiDevice_ object.

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/403)


tanabarr:
replaced by this approach: https://github.com/intel-hpdd/intel-manager-for-lustre/pull/458

PR #419, opened 2017-12-11
Add migration to add primary parameter to update managed targets moun…

### LINKED ISSUE #418: Upgrade issue from 3.1 to 4.0
In the patch that added MMP support to ZFS (4.0), we added a new field to the kwargs that are passed in with `UpdateManagedTargetMount`. 

The class (and kwargs) are ultimately pickled and saved in the db as a text field under `chroma_core_step_result.args`.

This causes an issue after upgrading to 4.0 as when the (3.1) field is hydrated back to a dict it's missing the `primary` key and throws.

We need to figure out a way of either working around that field, or deriving it's value from 3.1 data.

#419  fixes this

tanabarr:
…t to fix update tests

fixes #418 

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/419)

### DIFF
--- a/chroma-manager/chroma_core/migrations/0038_auto__add_primary_umtm_steps.py
+++ b/chroma-manager/chroma_core/migrations/0038_auto__add_primary_umtm_steps.py
@@ -13,7 +13,9 @@ class Migration(DataMigration):
         pipe(orm['chroma_core.stepresult'].objects.all(),
              cfilter(lambda x: 'UpdateManagedTargetMount' in str(x.step_class)),
              cfilter(lambda x: 'primary' not in x.args),
-             cmap(lambda x: x.args.update({'primary': True})))
+             cmap(lambda x: x.args.update({'primary': True}))
+             cmap(lambda x: x.save()))
+
 
     def backwards(self, orm):
         """backwards implementation"""
--- a/chroma-manager/chroma_core/migrations/0038_auto__add_primary_umtm_steps.py
+++ b/chroma-manager/chroma_core/migrations/0038_auto__add_primary_umtm_steps.py
@@ -18,7 +18,7 @@ class Migration(DataMigration):
 
 
     def backwards(self, orm):
-        """backwards implementation"""
+        pass
 
     models = {
         'auth.group': {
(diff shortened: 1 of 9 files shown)


chroma-manager/chroma_core/migrations/0038_auto__add_primary_umtm_steps.py
```diff
@@ -0,0 +1,1503 @@
...
+from toolz.curried import filter as cfilter
+from toolz.curried import map as cmap
+
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        # get stepresults for UpdateManagedTargetMount class and update stepresults to include 'primary' arg
+        pipe(orm['chroma_core.stepresult'].objects.all(),
+             cfilter(lambda x: 'UpdateManagedTargetMount' in str(x.step_class)),
```
johnsonw: Nice

chroma-manager/chroma_core/migrations/0038_auto__add_primary_umtm_steps.py
```diff
@@ -0,0 +1,1503 @@
+# -*- coding: utf-8 -*-
+import cPickle as pickle
```
jgrund: Is this import needed?

tanabarr: it's a better performing implementation and I have been in the habit of using it in preference to the original implementation

jgrund: Sure, I just meant I don't see `pickle` being used directly anywhere.

chroma-manager/tests/integration/installation_and_upgrade/test_data_consistancy.py
```diff
@@ -10,6 +11,7 @@ def test_access_all_data_all_endpoints(self):
                                 '/api/test_host/']
 
         end_points = self.get_json_by_uri("/api/", args={'limit': 0})
+        time.sleep(7200)
```
jgrund: Do we need this sleep here?

chroma-manager/chroma_core/migrations/0039_add_primary_umtm_stepresults.py
```diff
@@ -0,0 +1,1505 @@
+# -*- coding: utf-8 -*-
+import cPickle as pickle
```
jgrund: Sorry to keep harping on this, but is this import being used anywhere?

tanabarr: it is a quicker implementation than the standard pickle module, I'm in the habit of using it

chroma-manager/chroma_core/migrations/0039_add_primary_umtm_stepresults.py
```diff
@@ -0,0 +1,1499 @@
+# -*- coding: utf-8 -*-
+from south.v2 import DataMigration
+from toolz import pipe
+from toolz.curried import filter as cfilter
```
jgrund: We don't need toolz here anymore.

PR #464, opened 2018-02-02
Update manager and agent dependencies for reactive architecture

tanabarr:
Update manager to require `iml-device-aggregator` and agent to require `iml-scanner-proxy` (which itself requires `iml-device-scanner2`).

requirements include
* availability of relevant packages via relevant repositories (currently temporarily sourced from bintray)
* landing of https://github.com/intel-hpdd/scanner-proxy/pull/7

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/464)

### DIFF
--- a/chroma-agent/chroma-agent.spec
+++ b/chroma-agent/chroma-agent.spec
@@ -32,7 +32,7 @@ Requires: python2-iml-common1.3
 Requires: systemd-python
 Requires: python-tzlocal
 Requires: python2-toolz
-Requires: iml-device-scanner
+Requires: iml-scanner-proxy
 %if 0%{?rhel} > 5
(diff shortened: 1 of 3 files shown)


jgrund:
> If iml-scanner-proxy and iml-device-aggregator are ready to be published, you could just "subrepo" the module-tools repo into the their repos and push a tag (assuming tests pass) and they will be published to Copr.

It's more complicated than this, since the building will happen in copr instead of travis.

With the current approach, we build the node tarfile locally then use that to run mock. With copr we will need to first publish to npm (so copr can fetch the source tarball), then publish to copr.

Still doable I think, just needs to be done in phases.

Issue #477, opened 2018-02-23
Make sure to disable chroma-agent before rebooting


jgrund:
In:

https://github.com/intel-hpdd/intel-manager-for-lustre/blob/db03404348da2712ac4b075375b574d83ff56326/chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py#L115-L118

we restart the agents *after* we stop the agent daemons.

This ends up starting the agents again, which is going to cause us issues.

use ```systemctl disable chroma-agent``` on each node before rebooting. We'll want to make sure to enable them as well when we start the agents back up again.

tanabarr:
probably due to device scanning importing (sometimes with the force flag) and exporting pools, while this shouldn't be able to cause corruption due to double imports with MMP enabled there is a fair chance there are situations during the restart process where a disk might be wiped when imported on another note during agent device scanning

jgrund:
I think we can keep everything running in 4.1. Closing.

PR #478, opened 2018-02-23
Prevent chroma-agent.service from starting after restarting agents.

johnsonw:
Fixes GH-477.

- Prevents chroma-agent.service from automatically starting when
  restarting the agent nodes in create_lustre_filesystem integration
  test.
- enable chroma-agent.service and start the service after the filesystem
  is created.

Signed-off-by: Will Johnson <william.c.johnson@intel.com>

---
This change is [Reviewable](https://reviewable.io/reviews/intel-hpdd/intel-manager-for-lustre/478)

### DIFF
--- a/chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
+++ b/chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
@@ -112,9 +112,12 @@ class CreateLustreFilesystem(UtilityTestCase):
 
         next((self.clear_devices(x['nodename']) for x in config['lustre_servers']), None)
 
+        # There are some fs commands that need to be run when the chroma-agent daemon is not running; otherwise,
+        # it has the potential to cause disk I/O errors. Prevent the chroma-agent service from running after 
+        # rebooting the node.
         for server in config['lustre_servers']:
             self.remote_command(server['address'],
-                                'reboot',
+                                'systemctl disable chroma-agent.service && reboot',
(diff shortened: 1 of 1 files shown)


chroma-manager/tests/integration/existing_filesystem_configuration/utils/create_lustre_filesystem.py
```diff
@@ -191,6 +194,12 @@ def create_lustre_filesystem_for_test(self):
 
         self._save_modified_config()
 
+        for server in config['lustre_servers']:
```
jgrund: If we don't do this, are we not starting the agents?

jgrund: If we don't do this, are we not starting the agents?

johnsonw: It seemed that we would want to do it after setting up the server since it gets disabled before the restart. Would there be another place that does this?

jgrund: I think I may have thought this was an oversight, but if the agent isn't explicitly started already, it may have been by design.

johnsonw: Ok I will remove this line then :+1

johnsonw:
Will revisit once reactive architecture lands.

jgrund:
I don't think this is needed anymore as we won't import / export pools in 4.1. 

Because of that, we should be able to always keep the agents running.

PR #529, opened 2018-04-03
update device format for new device-scanner

jgrund:


jgrund:
Subsumed into reactive stack.

PR #554, opened 2018-04-16
Ensure stats get stored correctly in monitor mode failover

tanabarr:
If someone installs IML in monitor mode, but uses HA outside IML, they will not see stats upon failover.

https://github.com/intel-hpdd/intel-manager-for-lustre/blob/18f001a08e0b1e3027e38d3bd77d3bce777d9762/chroma-manager/chroma_core/services/lustre_audit/update_scan.py#L247-L262

We should still support stats display upon failover of a non-managed target.

This should be doable by checking if we are in monitor mode for the target and taking a different resolution path to determining if the target is runnable on that node.


tanabarr:
verified by hijacking an EFS run with ext4 backed targets and manually failing over (unmounting on one host and mounting on another with access to the same shared storage). statistics of activity before and after failover were reported, although as expected, Target was still listed as off-line (which I think would not be the case if we used the IML RA?)

PR #588, opened 2018-05-10
Settle before reading from device-scanner

jgrund:
Because the agent pulls data from device-scanner,
it is very important that we wait for udev to settle
before pulling a new state so we don't pull
an intermediate state.


chroma-agent/chroma_agent/device_plugins/block_devices.py
```diff
@@ -26,6 +27,11 @@
 
 
 def scanner_cmd(cmd):
+    # Because we are pulling from device-scanner,
+    # It is very important that we wait for
+    # the udev queue to settle before requesting new data
+    AgentShell.run(["udevadm", "settle"])
```
johnsonw: good call.

PR #593, opened 2018-05-10
Remove caching of localMounts

jgrund:
localMounts are currently being cached at module load time, and are
never evaluated afterwards.

This is causing us a bunch of issues where mount state is not as
expected.

Remove the caching, so we evaulate local mounts each time we call the
function.


jgrund:
Oops, was a little overzealous removing that part.

PR #605, opened 2018-05-14
Restart all services when restarting the agent.

jgrund:
Ensure we restart everything when the agent is scheduled for a restart.


jgrund:
Moved into https://github.com/intel-hpdd/iml-agent/pull/12

PR #606, opened 2018-05-14
Disable and stop all iml-storage-server.target services on remove

jgrund:
We only disable / stop the agent on host removal.
Instead, we should disable / stop all relevant
storage server units.


jgrund:
Moved to: https://github.com/intel-hpdd/iml-agent/pull/12

PR #612, opened 2018-05-16
Fixup tests to accommodate changes in iml-common#41

tanabarr:
Relates to changes made in https://github.com/intel-hpdd/iml-common/pull/41


tanabarr:
moving of imports is needed so that relevant mocking in base classes is effective, importing of test targets at module level stops mocking in base class initialisation methods not work

PR #660, opened 2018-06-22
Fix Makefile from subrepo

utopiabound:
This fixes general build targets in root: all rpms docs download subdirs substs clean_substs

Since the merged include/Makefile didn't support them, but was included in SUBDIRS.


jgrund:
Closing this, work will be subsumed in a future commit.

PR #669, opened 2018-07-02
Enable building in Jenkins

jgrund:
Cleanup some more cruft, and move the jenkins_build
back to the top-level of scripts so it will be picked up
by the jenkins job.


scripts/jenkins_build
```diff
@@ -56,18 +53,22 @@ pre_build()
...
+    # this precommit hook. They need to be
+    # fixed at some point.
+    # run_pre_commit_hook
 }
 
 build()
 {
     export CACHE=/mnt/builder_cache/
-    make
+    make rpms
```
johnsonw: Is this to be more explicit or do we have a different default task in the make file?

jgrund: It's a different task now.

PR #691, opened 2018-07-22
Move installation steps to service_config

### LINKED ISSUE #690: Move installation steps to service_config
Now that IML is available in docker, we need to move the config (much of it systemd specific) into `chroma-config setup`.

The reason this is needed is that when the RPM is installed in the docker build step, systemd is not available, but is available once someone starts the container.

Once this is done, starting the container and running `chroma-config setup` will yield a fully functional IML manager.

jgrund:
Fixes #690

Now that IML is available in docker, we need to move the config (much of it systemd specific) into chroma-config setup.

The reason this is needed is that when the RPM is installed in the docker build step, systemd is not available, but is available once someone starts the container.

Once this is done, starting the container and running chroma-config setup will yield a fully functional IML manager.


chroma_core/lib/service_config.py
```diff
@@ -193,10 +194,12 @@ def _setup_ntp(self, server):
                                                                                    error))
             raise RuntimeError("Failure when writing ntp config: %s" % error)
 
-        error = firewall_control.add_rule("123", "udp", "ntpd")
-        if error:
-            log.error("firewall command failed:\n%s" % error)
-            raise RuntimeError("Failure when opening port in firewall for ntpd: %s" % error)
+        if ServiceControl.create('firewalld').running:
```
johnsonw: Did we decide to take out firewalld?

chroma_core/lib/service_config.py
```diff
@@ -518,14 +518,78 @@ def _populate_database (self, username, password):
...
+
+        with open(conf_template, "r") as f:
+            config = f.read()
+            for setting in nginx_settings:
+                config = config.replace("{{%s}}" % setting, str(getattr(settings, setting)))
+
+            with open('/etc/nginx/conf.d/chroma-manager.conf', 'w') as f2:
+                f2.write(config)
+
+    def _create_fake_bundle(self):
```
johnsonw: Why do we need a fake bundle?

chroma_core/lib/util.py
```diff
@@ -163,22 +163,24 @@ def __str__(self):
 
 
 class CommandLine(object):
-    def try_shell(self, cmdline, mystdout = subprocess.PIPE,
-                  mystderr = subprocess.PIPE, stdin_text = None):
-        rc, out, err = self.shell(cmdline, mystdout, mystderr, stdin_text)
+    def try_shell(self, cmdline, mystdout=subprocess.PIPE,
+                  mystderr=subprocess.PIPE, stdin_text=None, shell=False):
+        rc, out, err = self.shell(
+            cmdline, mystdout, mystderr, stdin_text, shell=shell)
```
johnsonw: Where is "shell" coming from here?

jgrund: https://docs.python.org/2/library/subprocess.html#popen-constructor

chroma_core/lib/service_config.py
jgrund: We decided to not install the dep for docker, as it's not needed there. We still need to support it for non-docker installs.

chroma_core/lib/service_config.py
jgrund: Because the agent is deployed in a separate repo now, but we still need to simulate it as a bundle for the moment.

PR #692, opened 2018-07-23
Add flag if running in docker

jgrund:
Add a flag `service_config.py` that checks if
we are running in docker. We can use this flag
to skip setup of services that have been extracted into their
own container.

In addition, changes the requires to a wants for `iml-settings-populator`
so we don't die if postgres isn't present.

### DIFF
--- a/chroma_core/lib/service_config.py
+++ b/chroma_core/lib/service_config.py
@@ -52,6 +52,7 @@ firewall_control = FirewallControl.create()
 
 IS_DOCKER = os.path.exists('/.dockerenv')
 
+
 class ServiceConfig(CommandLine):
     REQUIRED_DB_SPACE_GB = 100
     bytes_in_gigabytes = 1073741824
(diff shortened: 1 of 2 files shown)


chroma_core/lib/service_config.py
```diff
@@ -305,16 +306,16 @@ def _stop_services(self):
...
-        if rc != 0:
-            if 'is not empty' not in out:
-                log.error("Failed to initialize postgresql service")
-                log.error("stdout:\n%s" % out)
-                log.error("stderr:\n%s" % err)
-                raise CommandError("service postgresql initdb", rc, out, err)
-            return
-        # Only mess with auth if we've freshly initialized the db
-        self._config_pgsql_auth(database)
+            rc, out, err = self.shell(["service", "postgresql", "initdb"])
```
johnsonw: Did these just get spaced out 4 more spaces?

jgrund: Not sure how that happened. Just auto-formatted the file instead.

PR #709, opened 2018-08-15
Update config move.

jgrund:
Update the triggerun so it fires correctly for the python2-iml-manager
subpackage.


jgrund:
> Should the echo be removed? 

I can take it out.

PR #711, opened 2018-08-16
Add back update_package in lustre_audit

utopiabound:
Allows upgrading from pre-4.1


jgrund:
add to whitelist

PR #713, opened 2018-08-21
Accept linux plugin device data

utopiabound:
Always try to use device-aggregator data first, but allow IML 4.0
to pass device data via linux plugin (/agent/message).

### DIFF
--- a/chroma_core/plugins/linux.py
+++ b/chroma_core/plugins/linux.py
@@ -200,17 +200,20 @@ class Linux(Plugin):
     def teardown(self):
         log.debug("Linux.teardown")
 
-    def agent_session_continue(self, host_id, _):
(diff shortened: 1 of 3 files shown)


chroma_core/plugins/linux.py
```diff
@@ -200,18 +200,21 @@ def __init__(self, resource_manager, scannable_id=None):
...
-        # devices = data
+    def agent_session_start(self, host_id, data, initial_scan=True):
         initiate_device_poll = False
         reported_device_node_paths = []
 
         fqdn = ManagedHost.objects.get(id=host_id).fqdn
         devices = get_devices(fqdn)
 
+        # use info from IML 4.0
+        if not devices and data:
```
jgrund: What get's populated for `data` after an upgrade?

PR #722, opened 2018-08-31
IML container tuning

jgrund:
Upon trying IML in docker on a larger system, I noticed a few areas of
improvement.

   1. We should scale down our gunicorn workers a bit (max out at 8).
This improves memory usage considerably while still maintaining the same
level of responsiveness.
  2. Add a retry to intermittent queue connection issues. While
debugging https://success.docker.com/article/ipvs-connection-timeout-issue I
noticed that kombu (rabbitMQ client lib) has a retry option. I've added
this to at least notify that we have an issue and might retry.
  3. The RabbitMQ container can be ephemeral; there is nothing durable
that needs to be stored. In addition we can use the management version
which will allow us to view connections / queues / exchanges etc... over
port 15672.

### DIFF
--- a/chroma_core/services/rpc.py
+++ b/chroma_core/services/rpc.py
@@ -30,6 +30,7 @@ import kombu.pools
 from kombu.common import maybe_declare
 from kombu.mixins import ConsumerMixin
 from kombu.messaging import Queue, Producer
+from kombu.entity import TRANSIENT_DELIVERY_MODE
 
 from chroma_core.services.log import log_register
 from chroma_core.services import _amqp_connection, _amqp_exchange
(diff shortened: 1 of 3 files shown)


chroma_core/services/rpc.py
```diff
@@ -160,9 +161,21 @@ def translate_error(err):
...
+
+            connection.ensure_connection(**retry_policy) 
+
             with Producer(connection) as producer:
-                maybe_declare(_amqp_exchange(), producer.channel)
-                producer.publish(result, serializer="json", routing_key=self.body['response_routing_key'], delivery_mode = 1, immedate = True, mandatory = True)
+
+                maybe_declare(_amqp_exchange(), producer.channel, True, **retry_policy)
+                producer.publish(result, serializer="json", routing_key=self.body['response_routing_key'],
+                                 delivery_mode=TRANSIENT_DELIVERY_MODE, retry=True, retry_policy=retry_policy, immedate=True, mandatory=True)
```
johnsonw: The formatting looks off here, or is github just rendering it weird?

chroma_core/services/rpc.py
```diff
@@ -349,9 +362,19 @@ def _send(self, connection, request):
         """
         log.debug("send %s" % request['request_id'])
         request['response_routing_key'] = self._response_routing_key
+
+        def errback(exc, _):
```
johnsonw: Both errback and retry_policy appear to be deplicated. Is it possible to define them once?

chroma_core/services/rpc.py
jgrund: maybe, I used the autoformatter in vscode to reformat it.

chroma_core/services/rpc.py
jgrund: It's possible, but don't think it buys much

PR #726, opened 2018-09-05
Restart nginx after upgrade

utopiabound:
When processes are asked to start, either them or if they are running, reload their configs.
    


chroma_core/lib/service_config.py
```diff
@@ -308,7 +308,13 @@ def _start_services(self):
         for service in self.CONTROLLED_SERVICES:
             controller = ServiceControl.create(service)
 
-            error = controller.start()
+            if controller.running:
+                if service.endswith('.target'):
```
jgrund: This is ok for now, but I wonder if longer term we want to target nginx explicitly rather than filtering out the target unit?

Issue #731, opened 2018-09-16
bundled Robinhood not working with Lustre 2.10.5


jgrund:
When Attempting to install a Robinhood server I am seeing:

```
Error: nothing provides liblustreapi.so.1()(64bit) needed by robinhood-lhsm-2.5.5-2.lustre.el7.centos.x86_64
```

We should upgrade to a newer robinhood that supports Lustre 2.10.x

utopiabound:
This should be fixed with upgrade to Lustre 2.12.1 (#902)

utopiabound:
We no longer get an error about `liblustreapi.so.1` not being provided, but we do get the following error:
```
--> Processing Dependency: httpd for package: php-5.4.16-46.el7.x86_64
Package httpd is obsoleted by python2-iml-manager, but obsoleting package does not provide for requirements
```
Because `python2-iml-manager` is in the same repository as iml-agent, it's obsoletes are used in package resolution and because it obsoletes httpd, it's placed as a replacement in the dependency solving tree.

*WORKAROUND*
Add `--exclude python2-iml-manager` to yum installation for robinhood rpms.

jgrund:
We should just remove the obsoletes for `httpd`.

PR #739, opened 2018-10-29
Only selected unused volume for failover test

utopiabound:
Only unused volumes can have primary/secondary set.

Fixes Issue #731


tests/integration/core/remote_operations.py
```diff
@@ -859,7 +861,7 @@ def get_pacemaker_targets(self, server, running=False):
         crm_resources = result.stdout.split('\n')
         targets = []
         for r in crm_resources:
-            if not re.search('chroma:Target', r):
+            if not re.search('lustre:Lustre', r):
```
jgrund: Does this need to cover ZFS as well?

utopiabound: No, we only care about the main lustre resource.

PR #753, opened 2018-11-19
Fix target mount for mpath

jgrund:


chroma_core/models/target.py
```diff
@@ -858,7 +858,7 @@ def run(self, kwargs):
             # in the case that the volume node is missing, attempt to import target volume
             self.invoke_agent_expect_result(kwargs['host'],
                                             'import_target',
-                                            {'device_type': ('linux' if device_type == 'ext4' else device_type),
+                                            {'device_type': ('linux' if device_type in ['ext4', 'mpath_member', ''] else device_type),
```
johnsonw: The device type can be an empty string?

johnsonw: I guess that would just mean that linux is the default right?

jgrund: the device_type is the filesystem_type, which can be null.

Ex from older device-aggregator:

```json
      "253:0": {
        "major_minor": "253:0",
        "path": "/dev/mapper/mpatha",
        "paths": [
          "/dev/mapper/mpatha",
          "/dev/disk/by-id/dm-name-mpatha",
          "/dev/disk/by-id/dm-uuid-mpath-3600140546b2d5d3365a44829376a534b",
          "/dev/mgt",
          "/dev/dm-0"
        ],
        "serial_80": "SLIO-ORG mgt1            46b2d5d3-365a-4482-9376-a534b46981b0",
        "serial_83": "3600140546b2d5d3365a44829376a534b",
        "size": "536870912",
        "filesystem_type": null,
        "filesystem_usage": null,
        "device_type": "disk",
        "device_path": "/devices/virtual/block/dm-0",
        "partition_number": null,
        "is_ro": false,
        "is_zfs_reserved": false,
        "parent": null,
        "dm_multipath": null,
        "dm_lv": null,
        "lv_uuid": null,
        "dm_vg": null,
        "vg_uuid": null,
        "dm_slave_mms": [
          "8:16",
          "8:48"
        ],
        "dm_vg_size": null,
        "md_uuid": null,
        "md_device_paths": []
      },
```

jgrund: The purpose of this is to map from a storage type to a blockdevice subclass in iml-common.

There are currently three subclasses, as seen here: https://github.com/whamcloud/iml-common/tree/master/iml_common/blockdevices.

If the fs type doesn't match one of those subclasses, it's an error.

This is an over-abstraction imo, but it's something we can tackle later.

PR #768, opened 2018-12-19
Bump Django to 1.6

jgrund:
Update Django from 1.4.5 to 1.6.11.

This is a very large patch out of necessity due to numerous breaking
changes post 1.4.

The biggest change is transaction management has been overhauled.
This is actually a "good thing"; autocommit per statement is the default
and transactions are explicit, not open for the duration of a http
request.

Atomicity is also strongly enforced within transactions.

Read more about the change [here](https://realpython.com/transaction-management-with-django-1-6/).

Because of this change, I've dropped all manual commits throughout the
manager. autocommit should enforce statement level commits, and if
autocommit is off, we are in an atomic transaction that can't be
committed until the transaction closes.

DSE has been replaced with massiviu, which I had to custom build and
host in Copr to work around a packaging bug in their code. The bug has
been fixed via a PR I sent, but a proper version still needs to be released.

I also needed to update the gevent config to account for threading
issues with db connections.

In addition, I added Warning output to the unit tests and fixed all the
warnings (there were lots) in the IML code. Some warnings still remain
in 3rd party code.

There were also many imports and function / method calls that slightly
changed and needed updating.

Files that were parsable by Python3 were also formatted by black.

Needless to say, since this patch touches so much of the core, it should
also get a thorough manual review.

9acd317


chroma_api/storage_resource.py
```diff
@@ -10,6 +10,7 @@
 from chroma_core.lib.storage_plugin.base_resource import BaseStorageResource
 
 from chroma_core.models import StorageResourceRecord, StorageResourceStatistic
+from chroma_core.services import dbutils
```
johnsonw:
```suggestion
```

chroma_core/lib/storage_plugin/query.py
```diff
@@ -47,6 +47,8 @@ def resource_get_alerts(self, resource):
 
         resource_alerts = StorageResourceAlert.filter_by_item_id(StorageResourceRecord, resource._handle)
 
+        resource_alerts = StorageResourceAlert.filter_by_item_id(StorageResourceRecord, resource._handle)
```
johnsonw:
```suggestion
```

chroma_core/models/server_profile.py
```diff
@@ -19,14 +19,16 @@ class ServerProfile(models.Model):
     ui_name = models.CharField(max_length=50, help_text="String, human readable name")
     ui_description = models.TextField(help_text="Description of the server profile")
     managed = models.BooleanField(help_text="Boolean, True if the host will be managed")
-    worker = models.BooleanField(help_text="Boolean, True if the host is available to be used as a Lustre worker node")
+    worker = models.BooleanField(
+        help_text="Boolean, True if the host is available to be used as a Lustre worker node", default=False
```
johnsonw: For consistency, should the "default=False" be the first entry in the tuple?

chroma_core/models/server_profile.py
```diff
@@ -19,14 +19,16 @@ class ServerProfile(models.Model):
     ui_name = models.CharField(max_length=50, help_text="String, human readable name")
     ui_description = models.TextField(help_text="Description of the server profile")
     managed = models.BooleanField(help_text="Boolean, True if the host will be managed")
```
johnsonw: Shouldn't this field also have a default value?

tests/unit/chroma_api/test_user.py
```diff
@@ -24,6 +24,7 @@ def test_HYD995_superuser_muggle(self):
 
         muggle = User.objects.create_user("muggle", "", "muggle123")
         muggle.groups.add(Group.objects.get(name="filesystem_users"))
+
```
johnsonw:
```suggestion
```

chroma_core/models/server_profile.py
```diff
@@ -14,50 +14,46 @@ class ServerProfile(models.Model):
...
-        help_text = "Boolean, True if the host will manage corosync"
+        help_text="Boolean, True if the host will manage corosync"
     )
     corosync2 = models.BooleanField(
-        help_text = "Boolean, True if the host will manage corosync2"
+        help_text="Boolean, True if the host will manage corosync2"
     )
     pacemaker = models.BooleanField(
-        help_text = "Boolean, True if the host will manage pacemaker"
+        help_text="Boolean, True if the host will manage pacemaker", default=False
```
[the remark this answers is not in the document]
jgrund: The rest that needed it should have it now.

chroma_core/models/server_profile.py
jgrund: Sure.

chroma_core/models/server_profile.py
jgrund: It doesn't seem to need one.

PR #773, opened 2019-01-08
Add support for Lustre 2.12.1

### LINKED ISSUE #743: Unable to update after update-check returns True
The iml-update-handler will raise an alert whenever it gets a message that updates are available.

However, IML currently tracks this via a `needs_update`  property on a `ManagedHost`. This property is distinct from the update alerts, which means it is a separate piece of state. 

Since it is distinct, we currently cannot update servers from the IML GUI without toggling this flag.

A few options to fix:

1. have iml-update-handler also do a `PUT` to /host to toggle the `needs_update` flag as needed.
2. Ignore the needs_update flag in the GUI, instead check if there are any `UpdatesAvailableAlert`s currently in progress. This approach will need vetting against an upgrade scenario before landing.

jgrund:
This patch adds support for Lustre 2.12.0

Fixes #743 

### DIFF
--- a/chroma_api/alert.py
+++ b/chroma_api/alert.py
@@ -305,6 +305,10 @@ class UpdatesAvailableAlertResource(Resource):
                 )
             )
 
+        if mh.needs_update is not available:
+            mh.needs_update = available
+            mh.save(update_fields=['needs_update'])
+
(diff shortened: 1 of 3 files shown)


utopiabound:
This is superseded by #825 
Pieces that was unrelated was split out to #829

PR #780, opened 2019-01-11
Create iml-warp-drive-service

jgrund:
Currently for each stateful object requested via the API,
we tack on extra information regarding it's available jobs +
transitions, and locks.

This is done in:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/5f8dcc3615388527c833df348617119352f51d07/chroma_api/utils.py#L198-L280

This is primarily done for convenience, but adds a bunch of overhead
both in RPC calls and database access.

Instead of fetching locks on demand, we can move to a push model.

Add a `locks` queue to rabbit that will push `LockChange` events from
the job scheduler.

Then create a new service `iml-warp-drive` that will consume this queue
and push the current lock state using Server-Sent Events (SSE) to any
interested clients.

Wire this service into Nginx so it can be accessed over the same URL +
port as everything else.

This service is both written in Rust and colocated with other services
in the IML manager.


iml-warp-drive/rust-iml-warp-drive.spec
```diff
@@ -0,0 +1,60 @@
...
+%global crate iml-warp-drive
+
+Name:           rust-%{crate}
+Version:        0.1.0
+Release:        1%{?dist}
+Summary:        streaming IML messages with Server-Sent Events
+
+License:        MIT
+
+URL:            https://crates.io/crates/warp-drive
```
johnsonw:
```suggestion
URL:            https://crates.io/crates/warp-drive
```

iml-warp-drive/src/main.rs
```diff
@@ -0,0 +1,198 @@
...
+  let state2 = state.clone();
+
+  // Save the sender in our list of connected users.
+  state2.lock().unwrap().insert(my_id, tx);
+
+  // Create channel to track disconnecting the receiver side of events.
+  // This is little bit tricky.
+  let (mut dtx, mut drx) = oneshot::channel::<()>();
+
+  // When `drx` will dropped then `dtx` will be canceled.
```
johnsonw:
```suggestion
  // When `drx` is dropped then `dtx` will be canceled.
```

iml-warp-drive/src/main.rs
```diff
@@ -0,0 +1,198 @@
...
+    unreachable!("unbounded rx never errors");
+  })
+}
+
+fn send_message(msg: String, state: &SharedUsers) {
+  log::debug!("Sending message {:?} to users {:?}", msg, state);
+
+  for (_, tx) in state.lock().unwrap().iter() {
+    match tx.unbounded_send(Message::Data(msg.clone())) {
+      Ok(()) => (),
```
johnsonw: What does this do?

iml-warp-drive/src/main.rs
jgrund: Done.

iml-warp-drive/src/main.rs
jgrund: Basically consumes the result and ignores it.

iml-warp-drive/src/main.rs
```diff
@@ -0,0 +1,137 @@
...
+    tokio::run(lazy(move || {
+        warp::spawn(
+            connect(
+                &env::get_addr(),
+                ConnectionOptions {
+                    username: env::get_user(),
+                    password: env::get_password(),
+                    vhost: env::get_vhost(),
+                    heartbeat: 0, // Turn off heartbeat due to https://github.com/sozu-proxy/lapin/issues/152
+                    frame_max: 65535,
```
johnsonw: Would it not make sense to make this a constant? I'm assuming this value is in documentation somewhere. Do you think it would be helpful to put a link to that documentation in here?

johnsonw:
Couple of questions but nothing blocking

iml-warp-drive/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "iml-warp-drive"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
```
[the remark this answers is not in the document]
jgrund: This is referring to the 2018 edition of rust, not what year this package is from.

docker/wait-for-dependencies.sh
```diff
@@ -1,11 +1,38 @@
...
+TMP=$PROXY_HOST
+TMP2=$AMQP_BROKER_HOST
+TMP3=$DB_HOST
+
+set -a
+source /var/lib/chroma/iml-settings.conf
+set +a
+
+
+if [[ ! -z "$TMP" ]]; then
```
johnsonw: So it's possible these variables didn't get set when it loaded the settings file?

jgrund: This is used by a bunch of upstream dockerfiles. Some of them will have the vars set, some will not.

johnsonw: ok that makes sense.

jgrund: This whole bit is messy really, I need a way to extract all env vars nicely into docker-compose, but haven't found a DRY way to do that with both compose and swarm.

chroma_core/services/job_scheduler/lock_cache.py
```diff
@@ -123,9 +123,14 @@ def _decorator(func):
 
 
 def to_lock_json(lock, add_remove=LockCache.LOCK_ADD):
+    if getattr(lock.locked_item, "downcast", None) and callable(lock.locked_item.downcast):
```
johnsonw: What was happening if it wasn't being downcasted? Was it throwing an error?

jgrund: It would return a different content_type_id iirc. You were seeing it on the MGT page during testing.

PR #784, opened 2019-01-17
Enforce no transactions before sending to notify queue

jgrund:

### DIFF
--- a/chroma_core/services/corosync/__init__.py
+++ b/chroma_core/services/corosync/__init__.py
@@ -45,9 +45,6 @@ class Service(ChromaService):
 
         self._queue = AgentRxQueue(Service.PLUGIN_NAME)
 
-    # Using transaction decorator to ensure that subsequent calls
-    # see fresh data when polling the ManagedHost model.
-    @transaction.atomic
     def on_data(self, fqdn, body):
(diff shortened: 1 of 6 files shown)


chroma_core/services/corosync/__init__.py
```diff
@@ -156,7 +152,7 @@ def is_new(peer_node_identifier):
                     #  Raise an Alert - system suppresses duplicates
                     log.debug("Alert notify on %s: active=%s" % (host, not host_reported_online))
                     HostOfflineAlert.notify(host, not host_reported_online)
-                    if host_reported_online == False:
+                    if host_reported_online is False:
```
utopiabound: Isn't this change worse? (c.f. [PEP8](https://www.python.org/dev/peps/pep-0008/#programming-recommendations))
```suggestion
                    if not host_reported_online:
```

jgrund: Flake8 seems to be indifferent:

screen shot 2019-01-18 at 9 39 12 am

I'm also indifferent, so I'll update.

jgrund: Though codefactor follows pep 8 and didn't like `is False`

jgrund:
Not worth pursuing further.

PR #785, opened 2019-01-17
Enforce default-options are non-null

jgrund:
The `PowerControlType` model specifies that default_options can be null.
If this happens, a template can end up sending power control commands to
the agent with `None` for options.

Update the model and add a SchemaMigration / DataMigration to convert
all Null options to "".


jgrund:
> Looks good to me. One question I had is whether or not the two migrations could be combined into one?

They are different types. One is a `DataMigration`, one is a `SchemaMigration`.

PR #787, opened 2019-01-25
Create /action endpoint

jgrund:
Currently for each stateful object requested via the API,
we tack on extra information regarding it's available jobs +
transitions, and locks.

This is done in:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/5f8dcc3615388527c833df348617119352f51d07/chroma_api/utils.py#L198-L280

This is primarily done for convenience, but adds a bunch of overhead
both in RPC calls and database access.

Add a /action endpoint that can be called on demand, and remove the actions being appended to utils.

This will make any API polling calls for stateful objects cheaper, as additional queue / db traffic is mitigated.

To call the endpoint, do something like the following:

```
https://localhost:8443/api/action/?composite_ids=62:1&composite_ids=62:2
```

Where 62 is the content_type_id, and 1,2 are the resource ids.

The endpoint can be called to resolve 1-N records at once.

### DIFF
--- a/chroma_api/action.py
+++ b/chroma_api/action.py
@@ -1,0 +1,123 @@
+# Copyright (c) 2018 DDN. All rights reserved.
+# Use of this source code is governed by a MIT-style
+# license that can be found in the LICENSE file.
+
+
+from django.contrib.contenttypes.models import ContentType
+from tastypie.authorization import DjangoAuthorization
+from tastypie.validation import Validation
+from tastypie.resources import Resource
+from tastypie import fields
+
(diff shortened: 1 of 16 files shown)


chroma_api/action.py
```diff
@@ -0,0 +1,123 @@
+# Copyright (c) 2018 DDN. All rights reserved.
```
johnsonw: 2019?

jgrund:
Landed in: https://github.com/whamcloud/integrated-manager-for-lustre/commit/82316d5c4634bdf18baf4ce0915c09692e2da14f

PR #790, opened 2019-01-30
Add wasm-components

jgrund:
This patch adds a new workspace: `wasm-components` to IML.

`wasm-components` are components compiled to WebAssembly and delivered
via the `/wasm_components` endpoint in Nginx.

Downstream consumers can then load the wasm components into their pages
by loading the wasm binary from the endpoint and invoking it in their
code.

Invocation can be done something like:

```
        <script src='/wasm-components/package.js'></script>

        <script>
            // the `wasm_bindgen` global is set to the exports of the Rust module
            const { render } = wasm_bindgen;
            // we'll defer our execution until the wasm is ready to go
            function run() {
                render();
            }
            // here we tell bindgen the path to the wasm file so it can run
            // initialization and return to us a promise when it's done
            wasm_bindgen('/wasm-components/package_bg.wasm')
                .then(run)
                .catch(console.error);
        </script>
```


iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+            button![
+                class!["btn", "btn-primary", "dropdown-toggle"],
+                mouse_ev("click", move |ev| {
+                    ev.stop_propagation();
+                    ev.prevent_default();
+                    next_open.clone()
+                }),
+                "Actions"
+            ],
+            div![
```
johnsonw: Is this where we will put the record elements?

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    let records:Vec<Record> = x.into_serde().unwrap();
+
+    let model = Model {
+        records,
+        ..Default::default()
+    };
+
+    let app = seed::App::build(model, update, view)
+        .window_events(window_events)
+        .mount("iml-add")
```
johnsonw: I was just reading up on `mount` and it looks like it will render the app to the element holding an id of `iml-add`. By definition we can only have one element with that id. It may not affect our stuff but it's probably something to pay attention to. Do you think it's something to be concerned with?

johnsonw: From https://seed-rs.org/guide/2:

> This will render your app to the element holding the id you passed; in the case of this example, “main”. The only part of the web page Seed will interact with is that element, so you can use other HTML not part of Seed, or other JS code/frameworks in the same document.

iml-wasm-components/iml-action-dropdown/src/lib.rs
jgrund: This is where the actions will go.

iml-wasm-components/iml-action-dropdown/src/lib.rs
jgrund: This won't look like this in final code.

We will need to generate unique id's for each call and add them as the id.

johnsonw: ah so it will be dynamic. That makes sense.

jgrund:
## What's Left

- [ ] Generate a uuid to pass *into* action-dropdown. Create node with that id outside the dropdown.
- [ ] Add mouseover listener on action button in wasm to start fetch
- [ ] Add fetch once per 10s to /action endpoint to get current actions.
- [ ] Stop fetch once component is destroyed.
- [ ] Emit event out of the component when the user clicks an action. Should have relevant action data within the event.
- [ ] propagate the event upwards to the store where we can dispatch the command modal.
- [ ] Reasonable testing.

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -9,6 +9,12 @@ use serde;
 
 type Record = (i64, i64);
 
+#[derive(serde::Deserialize, serde::Serialize, Debug, Eq, PartialEq, Hash, Clone)]
```
jgrund: Does this need to derive `Hash`?

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -131,16 +137,22 @@ fn window_events(model: Model) -> Vec<seed::Listener<Msg>> {
...
+
     let app = seed::App::build(model, update, view)
         .window_events(window_events)
-        .mount("iml-add")
+        .mount(staticUuid)
         .finish()
         .run();
 }
+
+fn string_to_static_str(s: String) -> &'static str {
```
jgrund: Put fixme here linking to https://github.com/David-OConnor/seed/issues/40

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -62,6 +62,14 @@ fn show_class(model: &Model) -> &str {
     }
 }
 
+fn composite_ids_to_query_string(compositeIds: Vec<Record>) -> String {
+    compositeIds
+        .iter()
+        .map(|(c, i)| format!("{}:{}", c, i))
```
jgrund: Since you've already created separate modules besides lib, you should organize new code into modules.

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -62,6 +62,14 @@ fn show_class(model: &Model) -> &str {
     }
 }
 
+fn composite_ids_to_query_string(compositeIds: Vec<Record>) -> String {
```
jgrund: Since you've already created separate modules besides lib, you should organize new code into modules.

johnsonw: Sure, for this I could put it in `api-transforms.rs`?

iml-wasm-components/iml-action-dropdown/src/lib.rs
johnsonw: agreed, where would be a good place for this to live? `composite-transforms.rs`?

iml-wasm-components/iml-action-dropdown/src/lib.rs
jgrund: maybe just in locks.rs for now? It's related to locks.

iml-wasm-components/iml-action-dropdown/src/lib.rs
jgrund: Also in locks.rs

iml-wasm-components/iml-action-dropdown/src/fetch.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use futures::Future;
```
jgrund: Take a look at this, it seems like a simpler way to do fetch: https://seed-rs.org/guide/5

johnsonw: Ah nice, it looks like they've already implemented something similar. I'll use this instead.

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -73,7 +113,8 @@ fn view(state: seed::App<Msg, Model>, model: &Model) -> El<Msg> {
     };
 
     let next_open = Msg::Open(!model.open);
-    let lock_count = locks::lock_list(&model.locks, &model.records).len();
+    let records = model.records.clone();
```
jgrund: Why clone here?

johnsonw: My first attempt was to do the following:
```
did_mount(move |_| spawn_local(get_actions(state.clone(), &model.records.clone())))
```
However, I get the following error message:
```
Explicit lifetime required in the type of 'model'
```
I did play around with an explicit lifetime but I wasn't having much luck.

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    Locks(Locks),
+    AvailableActions(AvailableActionsApiData),
+}
+
+/// The sole source of updating the model; returns a fresh one.
+fn update(msg: Msg, model: Model) -> Update<Model> {
+    match msg {
+        Msg::Open(open) => Render(Model { open, ..model }),
+        Msg::Locks(locks) => Render(Model { locks, ..model }),
+        Msg::AvailableActions(AvailableActionsApiData {
```
jgrund: You don't need to define fields you don't use:

```rust
Msg::AvailableActions(AvailableActionsApiData {
            objects: available_actions,
           ..
})
```

johnsonw: So interestingly, this was my first approach, but I put the `..` on the top, like this:
```
..,
objects: available_actions
```
and it gave me a syntax error. Putting the `..` on the bottom appears to work though. Thanks.

iml-wasm-components/iml-action-dropdown/src/api_transforms.rs
```diff
@@ -42,10 +43,34 @@ pub fn group_actions_by_label(objects: Vec<AvailableAction>, records: RecordMap)
         })
 }
 
+/// Sort items by display_group, then by display_order. Mark the last item in each group
+pub fn sort_actions(mut actions: Vec<AvailableAction>) -> Vec<AvailableAction> {
```
jgrund: Why do we need to know if something is last?

johnsonw: It's used to create a divider in the dropdown.

jgrund: Example? This seems unnecessary. I.E. We should be able to partition actions into resource groups, and we already do with ActionMap.

johnsonw: Ok, i'm building out the elements now so I'll do it without that property and remove it.

jgrund: Is this being called over keys on the ActionMap? If so, you should be able to iterate the HashMap which will tell you where each divider goes.

johnsonw: Right, yeah that makes sense so i'll remove the `last` property.

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -199,6 +214,11 @@ fn view(
         fetched_data,
     }: &Model,
 ) -> El<Msg> {
+    if records.len() == 0 {
+        log!("Returning empty");
```
jgrund: Do we need this log?

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -265,8 +297,11 @@ fn view(
             class!["btn-group dropdown", &open_class],
             btn,
             ul![class!["dropdown-menu", &open_class], record_els],
-            will_unmount(move |_| state.update(Msg::StopFetch)),
-        ]
+        ],
+        will_unmount(move |_| {
+            log!("Unmounting!");
```
jgrund: extra log

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -188,6 +194,15 @@ fn open_class(open: &bool) -> &str {
     }
 }
 
+/// Sends the custom event up to the window, carrying with it the data. Should we move this into its own module?
+fn dispatch_custom_event(_type: &str, data: &JsValue) {
```
jgrund: Instead of taking a `JsValue` and forcing the user to serialize it should take anything that implements `Serialize` and it unsized.

iml-wasm-components/iml-action-dropdown/src/lib.rs
johnsonw: removed

iml-wasm-components/iml-action-dropdown/src/lib.rs
johnsonw: removed

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+                    let x2 = x.clone();
+                    li![
+                        class!["tooltip-container", "tooltip-hover"],
+                        a![x.verb],
+                        mouse_ev(Ev::Click, move |ev| {
+                            ev.stop_propagation();
+                            ev.prevent_default();
+                            dispatch_custom_event(
+                                "action_selected",
+                                &JsValue::from_serde(&x2).unwrap(),
```
jgrund: This is unnecessary now.

iml-wasm-components/iml-action-dropdown/src/lib.rs
```diff
@@ -0,0 +1,347 @@
...
+        "open"
+    } else {
+        ""
+    }
+}
+
+/// Sends the custom event up to the window, carrying with it the data. Should we move this into its own module?
+
+fn dispatch_custom_event<T: serde::Serialize>(_type: &str, data: &T) {
+    let jsValue = JsValue::from_serde(data).unwrap();
```
jgrund: Careful with unwrap here, at least do an expect.

iml-wasm-components/iml-action-dropdown/src/lib.rs
johnsonw: I should have caught that

johnsonw:
Here's what the button looks like when it is in the waiting state. Note that the spinner is spinning:
image

johnsonw:
Deferred action dropdown:
1. Waits for mouseover before fetching the multistream
2. Once multistream is received, it renders the action dropdown button and passes the records to the component
3. Selected actions are propagated up through event bubbling just as it does on the server page.


[deferred-action-dropdown.mp4.zip](https://github.com/whamcloud/integrated-manager-for-lustre/files/2891704/deferred-action-dropdown.mp4.zip)

johnsonw:
Tooltips displaying in the position and size that is specified when instantiating the component:
1. Default position = 'left'
2. Default size = 'large'

image

.travis.yml
```diff
@@ -61,6 +61,14 @@ jobs:
         - pip install black
       script:
         - black --check ./
+    - stage: test
```
jgrund: Can this be built in azure instead?

chroma-manager.conf.template
```diff
@@ -133,6 +134,17 @@ server {
...
 
+    location /wasm-components {
+        alias /usr/share/iml-manager/iml-wasm-components;
+
+        etag on;
+        expires 1y;
+        add_header Cache-Control "public";
+
+        gzip on;
+        gzip_types text/css application/x-javascript application/javascript text/javascript application/json application/wasm;
```
jgrund: Do we need the css type here?

.travis.yml
johnsonw: Probably should. I'll update.

jgrund:
Landed as: https://github.com/whamcloud/integrated-manager-for-lustre/commit/c667aed5bd59c0907138d5a0216e2fd8cfcb61ed

PR #808, opened 2019-02-19
Jobstats: Do not return empty list when only one point fall in range

AlexTalker:
Hello again

The problem in this case that if only one value is returned from the model,
and `rate` operator specified(which is default for `fetch_jobs`), then related "Jobstats" page
says that no data is available.

However, that's not true.

In any way, you seem to have strange representation of timestamps since on the server side they tend to change up to 10 seconds and highly depend on client-side time, which is strange since you do display charts correctly BUT the user may not find associated `Jobstats` by clicking on the heat-map chart, since the values falls out of transformed `begin` and `end` range.
Also, hell knows why you do not show target names on the chart but the `PK` instead.

### DIFF
--- a/chroma_core/lib/metrics.py
+++ b/chroma_core/lib/metrics.py
@@ -72,7 +72,8 @@ class MetricStore(object):
         "Return datetimes with dicts of field names and values."
         result = collections.defaultdict(dict)
         types = set()
-        end = Stats[0].floor(end)  # exclude points from a partial sample
+        begin = Stats[0].round(begin)  # exclude points from a partial sample
+        end = Stats[0].round(end)  # exclude points from a partial sample
         for series in Series.filter(self.measured_object, name__in=fetch_metrics):
             types.add(series.type)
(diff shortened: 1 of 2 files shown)


AlexTalker:
Shrinking the time window breaks correlation between the data and timestamps shown in the `GUI`.
You probably might want to round both arguments but I changed only existing approach.

AlexTalker:
@jgrund @utopiabound Would you kindly review this one?

AlexTalker:
Actually, skipping subtraction was a bad idea for `/ui/jobstats` route, since it were a bit confused by incoming data and showed invalid values.
So I reverted this approach and just keep the time window intact instead.

jgrund:
test this please

jgrund:
test this please

jgrund:
test this please

AlexTalker:
@jgrund Could you do something about the tests?
Just for the note, I go to the test page, the only indicator for me is status of this PR.

PR #826, opened 2019-03-19
Remove another instance of crossing boundary while in transaction.

jgrund:


chroma_core/services/lustre_audit/update_scan.py
```diff
@@ -218,17 +218,19 @@ def update_target_mounts(self):
                         ["mounted", "unmounted"],
                     )
 
-            if target_mount.target.active_mount == None:
-                TargetRecoveryInfo.update(target_mount.target, {})
-                TargetRecoveryAlert.notify(target_mount.target, False)
+            if target_mount.target.active_mount is None:
```
utopiabound: Should this check be inside the atomic transaction?

jgrund: Like this?

```python
with transaction.atomic():
    if target_mount.target.active_mount is None:
        TargetRecoveryInfo.update(target_mount.target, {})
        TargetRecoveryAlert.notify(target_mount.target, False)
    elif mounted_locally:
        recovering = TargetRecoveryInfo.update(target_mount.target, recovery_status)
        TargetRecoveryAlert.notify(target_mount.target, recovering)
```

utopiabound: I think so.

jgrund: Sure, I'll update.

PR #831, opened 2019-03-23
Use a single "mode" for rpcs

jgrund:
IML currently has two modes for performing rpc calls.

- A lightweight mode, where no additional threads are spawned, but
separate AMQP connections are used for each RPC, and short lived queues
are created to handle RPC responses.

- A threaded mode, where a reponse handler thread takes responses. This
mode is meant to reduce the number of connections and queues.

Lightweight mode is actually very expensive in terms of resource from
connections and queues, so most code tries to use threaded mode.

Threaded mode also tries to limit active connections.

The lightweight mode is currently connection oriented, which seems strange
as best practice in rabbitmq is to use many channels over a single
connection.

As I dug into this, I realized the reason why is gevent will not
share open sockets between greenlets, which will naturally happen with
channels.

Overall, the lightweight mode is closer to how I think we should be
interfacing with RabbitMQ but has two problems.

1. Connections are spawned for concurrent requests.
1. Queues are created for each pending RPC response.

We can solve both of these and unify on a single "mode" by doing the
following:

1. Use a single connection per `ServiceRpcInterface` (One per process).
1. Create a channel for each call over `ServiceRpcInterface`.
1. Use a gevent [Semaphore](http://www.gevent.org/api/gevent.lock.html#gevent.lock.Semaphore) to enable connection / channel sharing between greenlets.
1. Use [Direct reply-to](https://www.rabbitmq.com/direct-reply-to.html) to avoid creating any new queues for RPC reponses.

This patch does all these things, unifying on a single mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,5 @@ localenv
 storage_server.repo
 base.repo
 copr-local.mk
+target/
+.cargo/
--- a/chroma_core/management/commands/chroma_service.py
+++ b/chroma_core/management/commands/chroma_service.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2018 DDN. All rights reserved.
+# Copyright (c) 2019 DDN. All rights reserved.
 # Use of this source code is governed by a MIT-style
 # license that can be found in the LICENSE file.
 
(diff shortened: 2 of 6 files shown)


jgrund:
This won't work as designed. It appears that gunicorn wants shorter-lived connections.

PR #832, opened 2019-03-25
Stratagem runner

utopiabound:
Initial stratagem runner work


stratagem-runner/.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
```
jgrund: This file isn't needed, these files are already ignored in the dir one level up.

stratagem-runner/build.rs
```diff
@@ -0,0 +1,68 @@
...
+// license that can be found in the LICENSE file.
+
+extern crate bindgen;
+
+use std::env;
+
+fn main() {
+    let out_file = env::current_dir().unwrap().join("src").join("bindings.rs");
+
+    // Tell cargo to tell rustc to link the system bzip2
```
jgrund:
```suggestion
    // Tell cargo to tell rustc to link the system liblustreapi
```

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+pub mod raw {
+    include!("bindings.rs");
+}
+use std::convert::From;
+use std::fmt;
+use std::ffi::{CStr, CString};
+use std::io;
+static PATH_BYTES: usize = 4096;
+
+// pub struct Statfs {
```
jgrund: Dead code

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    pub seq: u64,
+    pub oid: u32,
+    pub ver: u32,
+}
+impl fmt::Display for Fid {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "[0x{:x}:0x{:x}:0x{:x}]", self.seq, self.oid, self.ver)
+    }
+}
+// impl From<String> for Fid {
```
jgrund: Dead code.

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+impl From<raw::lu_fid> for Fid {
+    fn from(fid: raw::lu_fid) -> Self {
+        Fid { seq: fid.f_seq,
+              oid: fid.f_oid,
+              ver: fid.f_ver,
+        }
+    }
+}
+
+pub fn fid2path(device: &String, fidstr: &String) -> Option<String> {
```
jgrund:
```suggestion
    pub fn fid2path(device: &str, fidstr: &str) -> Option<String> {
```

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+    if !fidstr.starts_with("[") {
```
jgrund:
```suggestion
    if !fidstr.starts_with('[') {
```

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+            devptr,
+            fidptr,
+            ptr,
+            buf.capacity() as i32,
+            &mut recno as *mut std::os::raw::c_longlong,
+            &mut linkno as *mut std::os::raw::c_int,
+        )
+    };
+    if rc != 0 {
+        return None;
```
jgrund: What if the rc is non-zero? We should probably be returning a `Result` from this fn.

You could instead return something like:

```
Err(::std::io::Error::from_raw_os_error(r))?
```

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    unsafe {
+        let s = CStr::from_ptr(ptr);
+        let len = s.to_bytes().len();
+        buf.set_len(len);
+    };
+
+    let cstr = CString::new(buf).expect("Found invalid UTF-8");
+
+    Some(cstr.into_string().expect("Invalid UTF-8"))
```
jgrund: Instea of panicing here and on line 81, these should return the Error from the results.

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+//pub fn obd_statfs(filepath: &String) -> Option<raw::obd_statfs> {
+
+    //Some(raw::obd_statfs { })
```
jgrund: Dead code

stratagem-runner/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+//}
+
+pub fn rmfid(device: &String, fidlist: &Vec<String>) -> Result<(), io::Error> {
```
jgrund:
```suggestion
    pub fn rmfid(device: &str, fidlist: &[String]) -> Result<(), io::Error> {
```

stratagem-runner/build.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    // The bindgen::Builder is the main entry point
+    // to bindgen, and lets you build up options for
+    // the resulting bindings.
+    let bindings = bindgen::Builder::default()
+        .header("wrapper.h")
+        .constified_enum_module("boolean")
+        // // Logging
+        // .constified_enum_module("llapi_message_level")
```
johnsonw: Are all of these comments needed?

stratagem-runner/src/lib.rs
johnsonw: Shouldn't the formatter catch this?

jgrund: @johnsonw rust fmt doesn't but clippy does.

stratagem-runner/.gitignore
```diff
@@ -0,0 +1,2 @@
+**/*.rs.bk
```
jgrund: This file isn't needed.

stratagem-runner/Cargo.lock
```diff
@@ -0,0 +1,397 @@
+[[package]]
```
jgrund: This file isn't needed.

stratagem-runner/liblustreapi-sys/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "liblustreapi-sys"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+build = "build.rs"
+
+[dependencies]
+
+[build-dependencies]
+bindgen = "0.43.1"
```
jgrund: newer bindgen is available.

stratagem-runner/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "stratagem_warning"
+path = "src/stratagem_warning.rs"
+
+[[bin]]
+name = "stratagem_purge"
+path = "src/stratagem_purge.rs"
+
+
+[workspace]
+members = ['liblustreapi', 'liblustreapi-sys']
```
jgrund: These can't be nested, move these members up to the workspace root at: integrated-manager-for-lustre

stratagem-runner/liblustreapi-sys/build.rs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+extern crate bindgen;
+
+use std::env;
```
jgrund: This is not necessary

```suggestion
```

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+#![crate_name = "liblustreapi"]
+
+use std::convert::From;
```
jgrund: This shouldn't be needed.

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+use std::ffi::{CStr, CString};
+use std::fmt;
+use std::io;
+extern crate liblustreapi_sys as sys;
+static PATH_BYTES: usize = 4096;
+
+#[derive(Copy, Clone)]
+pub struct Fid {
+    pub seq: u64,
```
jgrund: Do we need to derive Clone, if this is Copy?

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+            seq: fid.f_seq,
+            oid: fid.f_oid,
+            ver: fid.f_ver,
+        }
+    }
+}
+
+#[derive(Copy, Clone)]
+pub struct StatfsState {
+    degraded: bool,
```
jgrund: Does this need to be Clone?

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    readonly: bool,
+    noprecreate: bool,
+    enospc: bool,
+    enoino: bool,
+}
+impl fmt::Display for StatfsState {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut list: Vec<String> = Vec::new();
+        if self.degraded {
+            list.push("degraded".to_string());
```
jgrund:
```suggestion
let mut list: Vec<String> = vec![];
```

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+}
+
+//pub fn obd_statfs(filepath: &String) -> Option<sys::obd_statfs> {
+
+//Some(sys::obd_statfs { })
+//}
+
+pub fn rmfid(device: &str, fidlist: &[String]) -> Result<(), io::Error> {
+    use std::fs; // replace with sys::llapi_rmfid
+    for fidstr in fidlist.iter() {
```
jgrund:
```suggestion
    for fidstr in fidlist {
```

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+        if let Ok(path) = fid2path(device, fidstr) {
```
jgrund: This will leave the `Error` variant unhandled.

Instead, do this:

```rust
let path = fid2path(device, &fidstr)?;

fs::remove_file(path)?;
```

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+        Err(x) => Err(io::Error::new(io::ErrorKind::InvalidData, x)),
+        Ok(x) => x
+            .into_string()
+            .map_err(|x| io::Error::new(io::ErrorKind::InvalidData, x)),
+    }
+}
+
+//pub fn obd_statfs(filepath: &String) -> Option<sys::obd_statfs> {
+
+//Some(sys::obd_statfs { })
```
jgrund: Dead code.

stratagem-runner/src/bindings.rs
```diff
@@ -0,0 +1,106 @@
+/* automatically generated by rust-bindgen */
```
jgrund: This file looks like a dup.

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+//}
+
```
jgrund: Overall:

```rust
pub fn rmfid(device: &str, fidlist: impl IntoIterator<Item = String>) -> Result<(), io::Error> {
    use std::fs; // replace with sys::llapi_rmfid

    for fidstr in fidlist {
        let path = fid2path(device, &fidstr)?;

        fs::remove_file(path)?;
    }

    Ok(())
}
```

stratagem-runner/src/stratagem_purge.rs
```diff
@@ -0,0 +1,20 @@
...
+use liblustreapi;
+use std::env;
+use std::process;
+
+fn main() {
+    let mut args = env::args();
+    let device = args.nth(1).expect("No device specified");
+
+    let fidlist: Vec<_> = args.map(|arg| arg).collect();
+    if let Err(e) = liblustreapi::rmfid(&device, &fidlist[..]) {
```
jgrund: With the change to rmfid, this can just be:

```suggestion
    if let Err(e) = liblustreapi::rmfid(&device, args) {
```

stratagem-runner/src/stratagem_purge.rs
```diff
@@ -0,0 +1,20 @@
...
+        println!("Failed to remove all fids {}", e);
```
jgrund: prefer using the log crate + env_logger crate over `println!`

stratagem-runner/liblustreapi/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "liblustreapi"
+version = "0.1.0"
+authors = ["Nathaniel Clark <nclark@whamcloud.com>"]
```
jgrund: Up to you, but I've been putting IML Team in this field.

stratagem-runner/src/stratagem_warning.rs
```diff
@@ -0,0 +1,19 @@
...
+use liblustreapi;
+use std::env;
+
+fn main() {
+    let mut args = env::args();
+    let device = args.nth(1).expect("No device specified");
+
+    for fid in args {
+        if let Ok(path) = liblustreapi::fid2path(&device, &fid) {
+            println!("{}, {}", path, fid);
```
jgrund: prefer using the log crate + env_logger crate over println!

stratagem-runner/liblustreapi-sys/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#![allow(non_upper_case_globals)]
```
jgrund: Missing copyright.

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+#![crate_name = "liblustreapi"]
```
jgrund: This shouldn't be needed.

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+
+use std::convert::From;
+use std::ffi::{CStr, CString};
+use std::fmt;
+use std::io;
+extern crate liblustreapi_sys as sys;
+static PATH_BYTES: usize = 4096;
+
+#[derive(Copy, Clone)]
```
jgrund: Do we need to derive Clone, if this is Copy?

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    fn from(fid: sys::lu_fid) -> Self {
+        Fid {
+            seq: fid.f_seq,
+            oid: fid.f_oid,
+            ver: fid.f_ver,
+        }
+    }
+}
+
+#[derive(Copy, Clone)]
```
jgrund: Does this need to be Clone?

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+pub struct StatfsState {
+    degraded: bool,
+    readonly: bool,
+    noprecreate: bool,
+    enospc: bool,
+    enoino: bool,
+}
+impl fmt::Display for StatfsState {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut list: Vec<String> = Vec::new();
```
jgrund:
```suggestion
        let mut list: Vec<String> = vec![];
```

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+            .map_err(|x| io::Error::new(io::ErrorKind::InvalidData, x)),
+    }
+}
+
+//pub fn obd_statfs(filepath: &String) -> Option<sys::obd_statfs> {
+
+//Some(sys::obd_statfs { })
+//}
+
+pub fn rmfid(device: &str, fidlist: &[String]) -> Result<(), io::Error> {
```
jgrund: Overall:

```rust
pub fn rmfid(device: &str, fidlist: impl IntoIterator<Item = String>) -> Result<(), io::Error> {
    use std::fs; // replace with sys::llapi_rmfid

    for fidstr in fidlist {
        let path = fid2path(device, &fidstr)?;

        fs::remove_file(path)?;
    }

    Ok(())
}
```

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+            degraded: state & sys::obd_statfs_state_OS_STATE_DEGRADED != 0,
+            readonly: state & sys::obd_statfs_state_OS_STATE_READONLY != 0,
+            noprecreate: state & sys::obd_statfs_state_OS_STATE_NOPRECREATE != 0,
+            enospc: state & sys::obd_statfs_state_OS_STATE_ENOSPC != 0,
+            enoino: state & sys::obd_statfs_state_OS_STATE_ENOINO != 0,
+        }
+    }
+}
+
+#[derive(Copy, Clone)]
```
jgrund: This could just be `Copy`

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub seq: u64,
+    pub oid: u32,
+    pub ver: u32,
+}
+impl fmt::Display for Fid {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "[0x{:x}:0x{:x}:0x{:x}]", self.seq, self.oid, self.ver)
+    }
+}
+fn num2u32(num: &str) -> u32 {
```
jgrund: Instead of unwrapping, these should return errors.

stratagem-runner/liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    if num.starts_with("0x") {
+        u32::from_str_radix(num.trim_start_matches("0x"), 16).unwrap()
+    } else if num.starts_with("0X") {
+        u32::from_str_radix(num.trim_start_matches("0X"), 16).unwrap()
+    } else {
+        num.parse().unwrap()
+    }
+}
+fn num2u64(num: &str) -> u64 {
```
jgrund: Instead of unwrapping, these should return errors.

stratagem-runner/liblustreapi/src/lib.rs
utopiabound: I can derive only Clone, but not only Copy
```
error[E0277]: the trait bound `Fid: std::clone::Clone` is not satisfied
  --> liblustreapi/src/lib.rs:12:10
   |
12 | #[derive(Copy)]
   |          ^^^^ the trait `std::clone::Clone` is not implemented for `Fid`
```

liblustreapi-sys/build.rs
```diff
@@ -4,10 +4,8 @@
 
 extern crate bindgen;
```
jgrund: This extern is unnecessary.

stratagem-runner/liblustreapi/src/lib.rs
jgrund: Sorry, it does need to be `Clone` to be `Copy`, as `Clone` is the supertrait: https://doc.rust-lang.org/std/marker/trait.Copy.html

You should keep the Copy trait though as it's more efficient than Clone.

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use std::convert::From;
+use std::ffi::{CStr, CString};
+use std::fmt;
+use std::io;
```
jgrund: Won't -1 but I've been compressing common import statements into a single one.

I think there will be a formatter for this in the next version of rustfmt.

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+extern crate liblustreapi_sys as sys;
```
jgrund:
```suggestion
use liblustreapi_sys as sys;
```

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use std::convert::From;
```
jgrund:
```suggestion
use std::{
    convert::From,
    ffi::{CStr, CString},
    fmt, io,
    num::ParseIntError,
    str::FromStr,
};
use liblustreapi_sys as sys;
```

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    pub oid: u32,
+    pub ver: u32,
+}
+impl fmt::Display for Fid {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "[0x{:x}:0x{:x}:0x{:x}]", self.seq, self.oid, self.ver)
+    }
+}
+// @@ There's got to be a better way to do this!
+fn num2u32(num: &str) -> u32 {
```
jgrund: This fn can be removed with change below.

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    if num.starts_with("0x") {
+        u32::from_str_radix(num.trim_start_matches("0x"), 16).unwrap()
+    } else if num.starts_with("0X") {
+        u32::from_str_radix(num.trim_start_matches("0X"), 16).unwrap()
+    } else {
+        num.parse().unwrap()
+    }
+}
+fn num2u64(num: &str) -> u64 {
```
jgrund: This fn can be removed with change below.

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    if num.starts_with("0x") {
+        u64::from_str_radix(num.trim_start_matches("0x"), 16).unwrap()
+    } else if num.starts_with("0X") {
+        u64::from_str_radix(num.trim_start_matches("0X"), 16).unwrap()
+    } else {
+        num.parse().unwrap()
+    }
+}
+impl From<String> for Fid {
```
jgrund: Since this conversion can fail, and it's from a string we can use `FromStr`:

```rust
impl FromStr for Fid {
    type Err = ParseIntError;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let fidstr = s.trim_matches(|c| c == '[' || c == ']');
        let arr: Vec<&str> = fidstr.split(':')
            .map(|num| num.trim_start_matches("0x"))
            .collect();

        Ok(Fid {
            seq: u64::from_str_radix(arr[0], 16)?,
            oid: u32::from_str_radix(arr[1], 16)?,
            ver: u32::from_str_radix(arr[2], 16)?,
        })
    }
}
```

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+        let fidstr = fidstr.trim_matches(|c| c == '[' || c == ']');
+        let arr: Vec<&str> = fidstr.split(':').collect();
+        Fid {
+            seq: num2u64(arr[0]),
+            oid: num2u32(arr[1]),
+            ver: num2u32(arr[2]),
+        }
+    }
+}
+impl From<[u8; 40usize]> for Fid {
```
jgrund: This will need to be updated as well, it contains an `.unwrap()` still.

```rust
impl From<[u8; 40usize]> for Fid {
    fn from(fidstr: [u8; 40usize]) -> Self {
        String::from_utf8_lossy(&fidstr)
            .into_owned()
            .parse::<Fid>()
            .unwrap()
    }
}
```

liblustreapi-sys/Cargo.toml
```diff
@@ -0,0 +1,11 @@
...
+name = "liblustreapi-sys"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+build = "build.rs"
+
+[dependencies]
+
+[build-dependencies]
+bindgen = "0.48.1"
```
johnsonw:
```suggestion
bindgen = "0.49.0"
```

liblustreapi-sys/build.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+fn main() {
+    let out_file = std::env::current_dir().unwrap().join("src").join("bindings.rs");
+
+    // Tell cargo to tell rustc to link the system liblustreapi
+    // shared library.
+    println!("cargo:rustc-link-lib=lustreapi");
```
johnsonw: Should we use the log crate here?

liblustreapi-sys/build.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    if out_file.exists() {
+        return;
```
johnsonw: Might be useful to write information to debug indicating that the file already exists so no processing will take place.

jgrund: You can't, this is in a build script (notice the println! above).

liblustreapi-sys/build.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+fn main() {
+    let out_file = std::env::current_dir().unwrap().join("src").join("bindings.rs");
```
johnsonw:
```suggest
let out_file = std::env::current_dir().expect("Couldn't get current directory.").join("src").join("bindings.rs");
```

liblustreapi-sys/src/lib.rs
```diff
@@ -0,0 +1,12 @@
...
+// license that can be found in the LICENSE file.
+
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+
+include!("bindings.rs");
+
+#[cfg(test)]
+mod tests {}
```
johnsonw: Why include this if there are no tests?

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    let cstr = CString::new(buf);
+    match cstr {
+        Err(x) => Err(io::Error::new(io::ErrorKind::InvalidData, x)),
+        Ok(x) => x
+            .into_string()
+            .map_err(|x| io::Error::new(io::ErrorKind::InvalidData, x)),
+    }
+}
+
+// pub fn obd_statfs(filepath: &String) -> Result<Statfs, io::Error> {
```
johnsonw: dead?

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+//     let mut path = filepath;
+//     sys::llapi_obd_statfs(path.as_ptr(),
+//     Ok(Statfs::from(statfs))
+// }
+
+pub fn rmfid(device: &str, fidlist: impl IntoIterator<Item = String>) -> Result<(), io::Error> {
+    use std::fs; // replace with sys::llapi_rmfid
```
johnsonw: Does this still need to be replaced?

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+    for fidstr in fidlist {
+        let path = fid2path(device, &fidstr)?;
+        fs::remove_file(path)?;
+    }
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
```
johnsonw: If we have any sample data it might be good to put a few scenarios in here unless it's pretty basic.

johnsonw:
Looks really good. A couple of comments.

liblustreapi-sys/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+// license that can be found in the LICENSE file.
+
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+
+include!("bindings.rs");
+
+type lstat_t = libc::stat64;
+pub const IOC_MDC_GETFILEINFO: u32 = 0xc0086916;
```
jgrund: How likely is this to change? Also, why not get this from bindgen?

stratagem-runner/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+name = "stratagem-runner"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+
+[dependencies]
+liblustreapi = { path = "../liblustreapi", version = "0.1.0" }
+csv = "1"
+serde = "1"
+serde_derive = "1"
```
jgrund: Don't need to add `serde_derive` separately, you can enable the derive feature on the serde crate:

`serde = { version = "1", features = ["derive"] }`

stratagem-runner/src/stratagem_warning.rs
```diff
@@ -0,0 +1,83 @@
...
+            let pb: std::path::PathBuf = [&mntpt, &path].iter().collect();
+            let fullpath = pb.to_str().unwrap();
+            let stat = match liblustreapi::mdc_stat(&fullpath) {
+                Ok(stat) => stat,
+                Err(e) => {
+                    eprintln!("Failed to mdc_stat({}) => {:?}", fullpath, e);
+                    return Err(e);
+                }
+            };
+            let user = unsafe {
```
jgrund: unsafe code should not propagate to this level, it should be encapsulated inside the liblustreapi / liblustreapi-sys crates.

Consumers of those crates should not be required to use any unsafe blocks to interact with them.

liblustreapi-sys/build.rs
```diff
@@ -0,0 +1,48 @@
...
+    let out_file = std::env::current_dir()
+        .unwrap()
+        .join("src")
+        .join("bindings.rs");
+
+    // Tell cargo to tell rustc to link the system liblustreapi
+    // shared library.
+    println!("cargo:rustc-link-lib=lustreapi");
+
+    if out_file.exists() && mtime!(&out_file) > mtime!(std::env::current_exe().unwrap()) {
```
jgrund: Checking for mtime as a guard means things like `cargo check` can't run if the header files don't exist.

It may not be obvious, but out_file by itself is a guard that allows compilation for other envs. Here's an example trying to compile this change on mac with mtime:

Screen Shot 2019-04-01 at 9 36 33 AM

liblustreapi-sys/src/lib.rs
```diff
@@ -0,0 +1,15 @@
...
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+#![allow(non_upper_case_globals)]
+#![allow(non_camel_case_types)]
+#![allow(non_snake_case)]
+
+include!("bindings.rs");
+
+type lstat_t = libc::stat64;
```
jgrund: I'm seeing an error here that `stat64` doesn't exist.

Should this be `libc::stat`?

jgrund: Must be because I'm building on mac

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+    convert::From,
+    ffi::{CStr, CString},
+    fmt,
+    io,
+    num::ParseIntError,
+    path::PathBuf,
+    str::FromStr,
+};
+extern crate errno;
+extern crate libc;
```
jgrund:
```suggestion
```

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+use std::{
+    convert::From,
+    ffi::{CStr, CString},
+    fmt,
+    io,
+    num::ParseIntError,
+    path::PathBuf,
+    str::FromStr,
+};
+extern crate errno;
```
jgrund:
```suggestion
```

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+        )
+    };
+    if rc != 0 {
+        return Err(io::Error::from_raw_os_error(rc.abs()));
+    }
+
+    buf2string(buf)
+}
+
+pub fn search_rootpath(fsname: &String) -> Result<String, io::Error> {
```
jgrund:
```suggestion
pub fn search_rootpath(fsname: &str) -> Result<String, io::Error> {
```

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+        );
+        return Err(io::Error::from_raw_os_error(rc.abs()));
+    }
+    buf2string(page)
+}
+
+pub fn mdc_stat(pathname: &str) -> Result<libc::stat64, io::Error> {
+    let page = Vec::with_capacity(PATH_BYTES);
+    let path = PathBuf::from(pathname);
+    let file = path.file_name().unwrap().to_str().unwrap();
```
jgrund: Should we do something other than panic if the path and dir do not exist?

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,222 @@
...
+        return Err(io::Error::from_raw_os_error(rc.abs()));
+    }
+    let statptr: *const libc::stat64 = page.as_ptr();
+    let stat: libc::stat64 = unsafe { *statptr };
+
+    Ok(stat)
+}
+
+pub fn rmfid(device: &String, fidlist: impl IntoIterator<Item = String>) -> Result<(), io::Error> {
+    use std::fs; // replace with sys::llapi_rmfid see LU-12090
```
jgrund: Can we make this into a @TODO?

liblustreapi-sys/build.rs
utopiabound: I'm trying to get it to rebuild if build.rs updated (similar to logic of make)

liblustreapi-sys/src/lib.rs
utopiabound: lstat_t defined to libc::stat64 is more correct and helps clean up a lot of the redefines from bindgen.
IOC_MDC_GETFILEINFO shouldn't change. It would be a Lustre regression if it does (c.f. [LU-8330](https://jira.whamcloud.com/browse/LU-8330)).  I could not get bindgen to cleanly generate the pub const, so this seemed cleaner.

stratagem-runner/Cargo.toml
utopiabound: I'll fix (I pulled this from the [csv documentation](https://docs.rs/csv/1.0.2/csv/))

stratagem-runner/src/stratagem_warning.rs
utopiabound: This is interacting with libc. uid -> username translation.  It doesn't seem to fit in liblustreapi.

liblustreapi-sys/build.rs
jgrund: Understood, my larger point is we need to be able to build this workspace on other OSes than linux.

stratagem-runner/src/stratagem_warning.rs
```diff
@@ -0,0 +1,80 @@
...
+) -> Result<(), StratagemError> {
+    let out = io::stdout();
+    let mntpt = liblustreapi::search_rootpath(&device).map_err(|e| {
+        eprintln!("Failed to find rootpath({}) -> {:?}", device, e);
+        e
+    })?;
+    let mut wtr = csv::Writer::from_writer(out);
+
+    for fid in args {
+        if let Ok(path) = liblustreapi::fid2path(&device, &fid) {
```
jgrund: This will silently discard any errors. Shouldn't we return an error here?

stratagem-runner/src/stratagem_warning.rs
jgrund: Ok, How about hoisting it out to a separate fn.

jgrund: ie:

```rust
            fn uid2user(uid: u32) -> Result<String, StratagemError> {
                let pwent = unsafe { libc::getpwuid(uid) };

                if pwent.is_null() {
                    eprintln!("Failed to getpwuid({})", uid);

                    return Err(std::io::Error::from(std::io::ErrorKind::NotFound).into());
                }

                unsafe { Ok(CStr::from_ptr((*pwent).pw_name).to_str()?.to_string()) }
            }
```

stratagem-runner/src/stratagem_warning.rs
```diff
@@ -0,0 +1,80 @@
...
+            let pb: std::path::PathBuf = [&mntpt, &path].iter().collect();
+            let fullpath = pb.to_str().unwrap();
+            let stat = liblustreapi::mdc_stat(&fullpath).map_err(|e| {
```
jgrund: Since `fullpath` gets converted to a `PathBuf`within this fn, you could just have this fn take a `PathBuf`. Then you won't need to do the potentially fallible conversion to `str`.

utopiabound:
So the Azure tests fail because they don't have lustre-client installed to link against for `cargo test`.

jgrund:
> So the Azure tests fail because they don't have lustre-client installed to link against for cargo test.

This won't build on mac either due to `libc::stat64`.

utopiabound:
> > So the Azure tests fail because they don't have lustre-client installed to link against for cargo test.
> 
> This won't build on mac either due to `libc::stat64`.

Is there a way to do conditional typing? (a la the below C?)
```C
#ifdef N
typedef struct stat64 lstat_t;
# else 
typedef struct stat lstat_t;
#endif
```

jgrund:
> Is there a way to do conditional typing? (a la the below C?)

yes, you can use conditional compilation to do this: https://doc.rust-lang.org/reference/conditional-compilation.html

liblustreapi/src/lib.rs
```diff
@@ -86,11 +86,11 @@ pub fn fid2path(device: &str, fidstr: &str) -> Result<String, LiblustreError> {
     let mut buf: Vec<u8> = vec![0; std::mem::size_of::<u8>() * PATH_BYTES];
     let ptr = buf.as_mut_ptr() as *mut libc::c_char;
 
-    let devptr = device.as_ptr() as *const i8;
-    let fidptr = fidstr.as_ptr() as *const i8;
+    let devptr = CString::new(device)?.into_raw();
+    let fidptr = CString::new(fidstr)?.into_raw();
```
jgrund: You'll want to make sure you call `from_raw` after the ffi call:

https://doc.rust-lang.org/std/ffi/struct.CString.html#method.into_raw

liblustreapi-sys/build.rs
```diff
@@ -0,0 +1,43 @@
...
+fn main() {
+    let out_file = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("src")
+        .join("bindings.rs");
+
+    // Tell cargo to tell rustc to link the system liblustreapi
+    // shared library.
+    println!("cargo:rustc-link-lib=lustreapi");
+
+    if out_file.exists() {
```
jgrund: This return should be before the println!

utopiabound: Nope.  If that happens, it doesn't link correctly:
```$ cargo build
   Compiling liblustreapi-sys v0.1.0 (/home/nclark/integrated-manager-lustre/liblustreapi-sys)
   Compiling liblustreapi v0.1.0 (/home/nclark/integrated-manager-lustre/liblustreapi)
   Compiling stratagem-runner v0.1.0 (/home/nclark/integrated-manager-lustre/stratagem-runner)
error: linking with `cc` failed: exit code: 1
  |
  = note: "cc" "-Wl,--as-needed" "-Wl,-z,noexecstack" "-m64" "-L" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.12kla9cy65x1yhjz.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.1bb95z8b7secl8aw.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.1i2vrvvrgz2hhpp8.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.1pde6m5quu76cyek.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.1xusjp2ruj5tei6a.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.2lo5emoto6tcqzjo.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.2qiobr7i5e3jw5gi.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.321x5b42melgvdba.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.3323k4t45b4gj399.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.3c3erb7g09qz4ley.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.3dmd75i0hchd49d6.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.3e4qndibgd7c01s5.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.3m8eixizit3upxps.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.3yzpe4qsz1ml217q.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.4a7404x5vbjnftwb.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.4bhhf4muakdgaaia.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.4par896ruacl8cpq.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.4uxggcsc2qkdwd36.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.54sde2aehb5z5th6.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.59vsmzkf5zfn0fv7.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.5bwz3kpsenuy2ks5.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.8ds7ev4t7qvds71.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.dk93sk1demxt5ms.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.t5yoibj9w9y38ys.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.tgbbtu59n6xop8s.rcgu.o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.u1zgqzm6avk7yok.rcgu.o" "-o" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4" "/home/nclark/integrated-manager-lustre/target/debug/deps/stratagem_purge-8d9a2b6204dca4d4.12lu82m7lxh0biq8.rcgu.o" "-Wl,--gc-sections" "-pie" "-Wl,-zrelro" "-Wl,-znow" "-nodefaultlibs" "-L" "/home/nclark/integrated-manager-lustre/target/debug/deps" "-L" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib" "-Wl,-Bstatic" "/home/nclark/integrated-manager-lustre/target/debug/deps/libliblustreapi-c9b4156e4beaff68.rlib" "/home/nclark/integrated-manager-lustre/target/debug/deps/libliblustreapi_sys-8312c18576d5bd5e.rlib" "/home/nclark/integrated-manager-lustre/target/debug/deps/liberrno-50bc5fdf875d2ff8.rlib" "/home/nclark/integrated-manager-lustre/target/debug/deps/liblibc-711591a738fe4507.rlib" "-Wl,--start-group" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-0e272be2569672f0.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/libpanic_unwind-369c8b15956ca76c.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/libbacktrace_sys-3c6f04eb2b03d2b3.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/libunwind-c56f1da838c839da.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_demangle-7a93a39a107d6ec4.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/liblibc-1bae7a39c5eff233.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/liballoc-f65a4b83daa5ffd7.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/librustc_std_workspace_core-c56dc9794e055f14.rlib" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcore-139abab598e46a29.rlib" "-Wl,--end-group" "/usr/lib/rustlib/x86_64-unknown-linux-gnu/lib/libcompiler_builtins-806a4ba0b6854dbf.rlib" "-Wl,-Bdynamic" "-lutil" "-lutil" "-ldl" "-lrt" "-lpthread" "-lgcc_s" "-lc" "-lm" "-lrt" "-lpthread" "-lutil" "-lutil"
  = note: /home/nclark/integrated-manager-lustre/target/debug/deps/libliblustreapi-c9b4156e4beaff68.rlib(liblustreapi-c9b4156e4beaff68.15a6bl2hegcj2r7s.rcgu.o): In function `liblustreapi::fid2path::h1bff9ea0b0268d7c':
          /home/nclark/integrated-manager-lustre/liblustreapi/src/lib.rs:95: undefined reference to `llapi_fid2path'
          /home/nclark/integrated-manager-lustre/target/debug/deps/libliblustreapi-c9b4156e4beaff68.rlib(liblustreapi-c9b4156e4beaff68.15a6bl2hegcj2r7s.rcgu.o): In function `liblustreapi::search_rootpath::hae2c5ea8ae94479d':
          /home/nclark/integrated-manager-lustre/liblustreapi/src/lib.rs:126: undefined reference to `llapi_search_rootpath'
          collect2: error: ld returned 1 exit status
          

```

jgrund: hrrm, but if it's not there compilation won't happen at all on mac as it's put in the linker

liblustreapi/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    if rc != 0 {
+        return Err(LiblustreError::os_error(rc.abs()));
+    }
+
+    buf2string(buf)
+}
+
+pub fn search_rootpath(fsname: &str) -> Result<String, LiblustreError> {
+    // @TODO this should do more validation
```
jgrund: What else needs to be validated here?

jgrund:
This needs to be rebased on top of master, and needs to use the `log` crate instead of `eprintln!`. Otherwise this looks good to me.

liblustreapi-sys/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "liblustreapi-sys"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+build = "build.rs"
+
+[dependencies]
+libc = "0.2.50"
```
johnsonw: These things seem to bump daily so I won't -1 but looks like you could bump this to 0.2.51.

liblustreapi/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "liblustreapi"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+
+[dependencies]
+liblustreapi-sys = { path = "../liblustreapi-sys", version = "0.1.0"}
+libc = "0.2.50"
```
johnsonw: This could bump to 0.2.51

stratagem-runner/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+name = "stratagem-runner"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+
+[dependencies]
+liblustreapi = { path = "../liblustreapi", version = "0.1.0" }
+csv = "1"
+serde = { version = "1", features = ["derive"] }
+libc = "0.2.50"
```
johnsonw: Can upgrade to 0.2.51.

PR #834, opened 2019-03-26
Extract useful internal manager messaging

jgrund:
In order to send messages to the IML agent and between IML manager
processes, extract a few fns that can be reused between crates.

Signed-off-by: Joe Grund <jgrund@whamcloud.com>


jgrund:
Landed in: dceaa5a03c3e8407d7dbe25da2db939f277fc5ba

PR #835, opened 2019-03-27
Locks and Actions Base

johnsonw:
This patch pulls out the core python bits from the aggregated locks and actions patches and puts them into a single base patch upon which other patches will live on top of.

### DIFF
--- a/chroma-manager/chroma-manager.conf.template
+++ b/chroma-manager/chroma-manager.conf.template
@@ -3,6 +3,11 @@ map $ssl_client_s_dn $ssl_client_s_dn_cn {
     ~CN=(?<CN>[^,]+) $CN;
 }
 
+map $status $loggable {
+    ~^[23]  0;
+    default 1;
+}
+
 server {
     listen {{HTTPS_FRONTEND_PORT}} ssl http2;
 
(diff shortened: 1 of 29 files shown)


chroma-manager/chroma-manager.conf.template
```diff
@@ -3,14 +3,50 @@ map $ssl_client_s_dn $ssl_client_s_dn_cn {
     ~CN=(?<CN>[^,]+) $CN;
 }
 
+map $status $loggable {
```
jgrund: Drop this loggable.

chroma-manager/chroma-manager.conf.template
```diff
@@ -3,14 +3,50 @@ map $ssl_client_s_dn $ssl_client_s_dn_cn {
...
+
 server {
     listen {{HTTPS_FRONTEND_PORT}} ssl http2;
 
     error_page 497 https://$http_host$request_uri;
 
     proxy_read_timeout 330s;
 
-    include {{APP_PATH}}/mime.types;
+    access_log /var/log/nginx/access.log combined if=$loggable;
```
jgrund: and this.

chroma-manager/chroma-manager.conf.template
johnsonw: removed

PR #863, opened 2019-04-10
Create Actions for Stratagem on Manager side

### LINKED ISSUE #852: Create Actions for Stratagem - Manager Side
Create Commands, Jobs, Steps for Stratagem on the manager side.

## ConfigureStratagem

   - Set the interval, 
   - set report duration 
   - set report duration active
   - set purge duration
   - set purge duration active

  - Creates a systemd timer on the manager that is responsible for calling the following command after either interval has elapsed (if the configuration is active for that particular command):
```
iml stratagem scan <FS_NAME> {report|purge}
```

## UnconfigureStratagem

  - Removes the systemd timer created during ConfigureStratagem.

## RunStratagem

  - Kicks off a run

  - Sends info to each MDS

  - Waits for all results

  - Aggregates results

  - Ingests Metrics

  - Sends relevant results to client to run actions.

  - Reports back results of client run.

johnsonw:
Fixes #852.


chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,66 @@
...
+from chroma_core.lib.cache import ObjectCache
+from chroma_core.lib.job import Step
+from chroma_core.models import Job
+from chroma_core.models import StateChangeJob
+from chroma_help.help import help_text
+
+
+class Stratagem(models.Model):
+    id = models.IntegerField(primary_key=True, default=1)
+    interval = models.IntegerField(help_text="Interval value in which a stratagem run will execute")
```
utopiabound: What unit is the interval and durations in?

johnsonw: I was thinking a unit of days would make the most sense. In addition to that we would probably need to specify a time of day (in UTC I would imagine).

jgrund: We'd probably want it to have a more granular lower bound. We can always scale up to days, but can't scale down from them.

johnsonw: Do you guys have an opinion on adding a units dropdown?

jgrund: I'd be fine with using a more configurable picker for the GUI. 

On the backend, we want to store in granular units, even if we end up using coarser units on the front-end.

johnsonw: yes, absolutely. I'll check into what systemd uses for timers but I was thinking of using minutes.

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1741,3 +1742,31 @@ def update_lnet_configuration(self, lnet_configuration_list):
     @property
     def CommandPlan(self):
         return CommandPlan(self._lock_cache, self._job_collection)
+
+    def configure_stratagem(self, stratagem_data):
+        with self._lock:
+            with transaction.atomic():
+                StratagemConfiguration.objects.all().delete()
```
jgrund: Why delete everything here?

johnsonw: My thought was that anytime we receive a request to configure stratagem we would delete the old configuration and then define the new configuration. I can try to create a new entry where the id is 1 and see if it overwrites the existing values in the db. If that works I should be able to remove the delete.

chroma_api/stratagem.py
```diff
@@ -0,0 +1,32 @@
...
+    class Meta:
+        resource_name = "stratagem_configuration"
+        queryset = StratagemConfiguration.objects.all()
+        authorization = DjangoAuthorization()
+        authentication = AnonymousAuthentication()
+        allowed_methods = ["get", "post"]
+
+    def obj_create(self, bundle, **kwargs):
+        super(StratagemConfigurationResource, self).obj_create(bundle, **kwargs)
+        stratagem_data = bundle.data.get("objects", [bundle.data])
```
jgrund: This logic should be pushed into a validation Class.

chroma_core/services/job_scheduler/job_scheduler.py
jgrund: You can just update the existing object. It's equivalent to an `UPDATE` in SQL.

chroma_help/help.py
```diff
@@ -14,6 +14,9 @@
     "start_copytool": "Start this copytool to begin processing HSM requests.",
     "stop_copytool": "Stop this copytool. In-progress HSM actions will be canceled.",
     "configure_copytool": "Configure this copytool on the worker.",
+    "configure_stratagem": "Configure's Stratagem on the selected node.",
+    "configure_stratagem_long": "Configure's Stratagem on the selected node.",
+    "configure_stratagem_description": "Configure Stratagem description",
```
jgrund: Fixme

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+# Use of this source code is governed by a MIT-style
+# license that can be found in the LICENSE file.
+
+import logging
+
+from django.db import models
+
+from chroma_core.lib.cache import ObjectCache
+from chroma_core.models.jobs import StatefulObject
+from chroma_core.lib.job import Step, job_log, DependOn, DependAll, DependAny
```
jgrund: Cruft here.

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+        default=False, help_text="Indicates if the purge should execute at the given interval"
+    )
+
+    states = ["unconfigured", "configured"]
+    initial_state = "unconfigured"
+
+    # def set_state(self, state, intentional=False):
+    #     job_log.debug("configure_stratagem.set_state %s %s" % (state, intentional))
+    #     super(StratagemConfiguration, self).set_state(state, intentional)
+    #     StratagemUnconfiguredAlert.notify_warning(self, self.state == "unconfigured")
```
jgrund: Cruft

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+
+    class Meta:
+        app_label = "chroma_core"
+        ordering = ["id"]
+
+    # def get_deps(self, state=None):
```
jgrund: Cruft

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+    #     return DependOn(ObjectCache.get_one(ServerProfile, lambda p: p.name == "stratagem_server"), True)
+
+    # def get_available_states(self, begin_state):
```
jgrund: Cruft

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+            alert_item=self.alert_item.primary_host,
+            alert=self,
+            severity=logging.INFO,
+        )
+
+    def affected_targets(self, affect_target):
+        affect_target(self.alert_item)
+
+
+class ConfigureSystemdTimerStep(Step):
```
jgrund: This have a name that includes `Stratagem`.

i.e. `ConfigureStratagemTimerStep`

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+    def run(self, kwargs):
+        job_log.debug("Create systemd time Step kwargs: {}".format(kwargs))
```
jgrund: mention stratagem here.

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+    display_order = 10
+
+    requires_confirmation = False
+    state_verb = "Configure Stratagem"
+
+    class Meta:
+        app_label = "chroma_core"
+        ordering = ["id"]
+
+    # def __init__(self, *args, **kwargs):
```
jgrund: Cruft

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def description(self):
+        return help_text["configure_stratagem_description"]
+
+    def get_steps(self):
+        steps = [(ConfigureSystemdTimerStep, {})]
+
+        return steps
+
+    # def get_deps(self, state=None):
```
jgrund: Cruft

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+    #     return DependOn(ObjectCache.get_one(ServerProfile, lambda p: p.name == "stratagem_server").managed, True)
+
+    # def create_locks(self):
```
jgrund: Cruft

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,122 @@
...
+    #     locks = super(ConfigureStratagemJob, self).create_locks()
+
+    #     # Take a write lock on mtm objects related to this target
```
jgrund: Cruft

chroma_core/models/stratagem.py
johnsonw: Looks like you can specify units for systemd timers (eg. 15min). I'm starting to think seconds might be a good unit for the backend. What do you guys think?

jgrund: Seconds should be more than granular enough.

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: Updated to the following:
```
configuration_data = {
                    'state': "unconfigured",
                    'id': 1,
                    'interval': stratagem_data.get('interval'),
                    'report_duration': stratagem_data.get('report_duration'),
                    'report_duration_active': stratagem_data.get('report_duration_active'),
                    'purge_duration': stratagem_data.get('purge_duration'),
                    'purge_duration_active': stratagem_data.get('purge_duration_active')
                }

                if StratagemConfiguration.objects.exists() == True:
                    stratagem_configuration = StratagemConfiguration.objects.update(**configuration_data)
                else:
                    stratagem_configuration = StratagemConfiguration.objects.create(**configuration_data)
```

chroma_help/help.py
johnsonw: Do we know what these values should be?

chroma_core/models/stratagem.py
johnsonw: removed

chroma_core/models/stratagem.py
johnsonw: removed

chroma_core/models/stratagem.py
johnsonw: removed

chroma_core/models/stratagem.py
johnsonw: removed

chroma_core/models/stratagem.py
johnsonw: removed

chroma_core/models/stratagem.py
johnsonw: removed

chroma_core/models/stratagem.py
johnsonw: removed

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,174 @@
...
+                                "argument": "smaller_than_1M",
+                            },
+                            {
+                                "action": "LAT_COUNTER_INC",
+                                "expression": "&& >= size 1048576 < size 1048576000",
+                                "argument": "not_smaller_than_1M_and_smaller_than_1G",
+                            },
+                            {
+                                "action": "LAT_COUNTER_INC",
+                                "expression": ">= size 1048576000",
```
utopiabound: Should this be bounded between 1GB and 1TB?  
Are there any requirements for striations?

jgrund:
@johnsonw the rust-python bridge patch has landed, you can rebase this on master now.

johnsonw:
Thanks @jgrund will do this shortly.

chroma_core/models/stratagem.py
johnsonw: I took this from the baked json. Is this not correct?

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,214 @@
...
+        affect_target(self.alert_item)
+
+
+class ConfigureStratagemTimerStep(Step):
+    def run(self, kwargs):
+        job_log.debug("Create stratagem timer step kwargs: {}".format(kwargs))
+        # Create systemd timer
+
+
+class ConfigureStratagemJob(StateChangeJob):
```
jgrund: One thing to note is that configuration needs to be bounded per filesystem. This means there must be a separate configuration / timer for each FS managed by IML.

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,214 @@
...
+        result = json.loads(result)
+
+        if "Err" in result:
+            self.log("Error Scanning {}: \n{}".format(path, json.dumps(result["Err"], indent=2)))
+        else:
+            self.log("Successfully scanned {}:\n{}".format(path, json.dumps(result["Ok"], indent=2)))
+
+
+class RunStratagemJob(Job):
+    mdt_id = models.IntegerField()
```
jgrund: Again, this will need to be bounded per-filesystem.

We don't want to scan one filesystem and then try to use it's results on another.

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -867,9 +868,11 @@ def notify(self, content_type, object_id, time_serialized, update_attrs, from_st
 
     def run_jobs(self, job_dicts, message):
         with self._lock:
-            result = self.CommandPlan.command_run_jobs(job_dicts, message)
-            self.progress.advance()
-            return result
+            with transaction.atomic():
```
jgrund: Why the atomic here?

chroma_api/stratagem.py
```diff
@@ -0,0 +1,62 @@
...
+
+        command_id = JobSchedulerClient.configure_stratagem(stratagem_data[0])
+
+
+class RunStratagemValidation(Validation):
+    def is_valid(self, bundle, request=None):
+        # Need to verify that stratagem is configured.
+        return {}
```
jgrund: Validate that an existing fs with the correct profiles has been passed.

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: I noticed it being used in other places, such as "set_host_profile" so I thought I should probably add it here as well. Will this cause an issue?

jgrund: It really should only be added if there is a clear reason to do so.

chroma_core/models/stratagem.py
johnsonw: I'll handle this by passing in the filesystem id and only retrieve mdt's associated with that filesystem.

jgrund: Sounds good. In addition, you need to make sure the servers the MDTs reside on have the proper profile installed.

johnsonw: 👍

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: ok, i'll remove it.

chroma_api/stratagem.py
```diff
@@ -59,7 +59,7 @@ def is_valid(self, bundle, request=None):
 
 
 class StratagemConfigurationResource(ChromaModelResource):
-    filesystem_id = fields.CharField(attribute="filesystem_id", max_length=8, null=False)
+    filesystem_id = fields.CharField(attribute="filesystem_id", null=False)
```
jgrund: Why `CharField`?

johnsonw:
The parser now takes stratagem results, converts them into influxdb points, and the points are then submitted to influxdb in step 2 of the process. After submitting I checked the db with the following results:

```
> select * from stratagem_scan
name: stratagem_scan
time       classify_attr_type count counter_name                       expression group_name
----       ------------------ ----- ------------                       ---------- ----------
1560852083 uid                3     uid_0                                         user_distribution
1560852083                    0     less_than_1m                                  size_distribution
1560852083                    3     classify_of_rule_0                 1          user_distribution
1560852083                    0     greater_than_equal_1m_less_than_1g            size_distribution
1560852083                    0     greater_than_equal_1t                         size_distribution
1560852083                    0     greater_than_equal_1g                         size_distribution
1560852202                    0     greater_than_equal_1t                         size_distribution
1560852202                    0     greater_than_equal_1g                         size_distribution
1560852202                    8     classify_of_rule_0                 1          user_distribution
1560852202 uid                8     uid_0                                         user_distribution
1560852202                    0     less_than_1m                                  size_distribution
1560852202                    0     greater_than_equal_1m_less_than_1g            size_distribution
1560854797                    0     less_than_1m                                  size_distribution
1560854797                    0     greater_than_equal_1m_less_than_1g            size_distribution
1560854797                    0     greater_than_equal_1g                         size_distribution
1560854797                    0     greater_than_equal_1t                         size_distribution
1560854797 uid                3     uid_0                                         user_distribution
1560854797                    3     classify_of_rule_0                 1          user_distribution
```

johnsonw:
Mount is created before streaming to client.
mounts

chroma_core/lib/job.py
```diff
@@ -84,6 +84,7 @@ def satisfied(self):
             depended_object = self.get_stateful_object()
         except:
             self.stateful_object.__class__._base_manager.get(pk=self.stateful_object.pk)
+
```
jgrund: Can you extract the changes in this file to it's own patch? They can land separately from this work.

chroma_help/help.py
jgrund: Can you describe where you are using the values above?

iml-manager-cli/src/main.rs
```diff
@@ -129,17 +129,24 @@ fn start_spinner(msg: &str) -> impl FnOnce(Option<String>) -> () {
     }
 }
 
-fn display_cmd_state(cmd: &Command) {
+fn display_cmd_state(cmd: &Result<Command, iml_manager_client::RunStratagemValidationError>) {
```
jgrund: These changes should not live in this patch.

iml-manager-client/src/lib.rs
```diff
@@ -2,16 +2,21 @@
 // Use of this source code is governed by a MIT-style
 // license that can be found in the LICENSE file.
 
-use futures::{Future, IntoFuture as _};
-use reqwest::{header, r#async::Client, Url};
+use futures::{Future, IntoFuture as _, Stream as _};
```
jgrund: These changes should not live in this patch.

settings.py
```diff
@@ -61,6 +61,8 @@
 
 INFLUXDB_STRATAGEM_SCAN_DB = "iml_stratagem_scans"
 
+INFLUXDB_PORT = 8086
```
jgrund: This is odd because the port here does not actually configure anything, it's just reflecting the default value of influxdb.

In other words changing this would not actually change the config of influx.

iml-manager-client/src/lib.rs
```diff
@@ -2,16 +2,21 @@
 // Use of this source code is governed by a MIT-style
```
jgrund: This should not be part of this patch.

iml-manager-cli/src/main.rs
```diff
@@ -129,17 +129,24 @@ fn start_spinner(msg: &str) -> impl FnOnce(Option<String>) -> () {
     }
 }
 
-fn display_cmd_state(cmd: &Command) {
```
jgrund: This should not be part of this patch.

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -872,8 +874,9 @@ def notify(self, content_type, object_id, time_serialized, update_attrs, from_st
     def run_jobs(self, job_dicts, message):
         with self._lock:
             result = self.CommandPlan.command_run_jobs(job_dicts, message)
-            self.progress.advance()
-            return result
+
+        self.progress.advance()
```
jgrund: Why is this moved outside the lock?

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1748,106 @@ def update_lnet_configuration(self, lnet_configuration_list):
     @property
     def CommandPlan(self):
         return CommandPlan(self._lock_cache, self._job_collection)
+
+    def configure_stratagem(self, stratagem_data):
+        with self._lock:
```
jgrund: What is the lock doing here?

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1748,106 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+                "filesystem_id": stratagem_data.get("filesystem"),
+                "interval": stratagem_data.get("interval"),
+                "report_duration": stratagem_data.get("report_duration"),
+                "report_duration_active": stratagem_data.get("report_duration_active"),
+                "purge_duration": stratagem_data.get("purge_duration"),
+                "purge_duration_active": stratagem_data.get("purge_duration_active"),
+            }
+
+            # The filesystem_id may come in as the fs name or the fs id. In terms of storing information in the database, the fs id should always be used.
+            fs_identifier = str(configuration_data.get("filesystem_id"))
```
jgrund: What happens if the dict doesn't have this value?

jgrund: In other words:

```
>>> str(None)
'None'
```

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1748,106 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+            fs_id = filter(
+                lambda x: str(x.get("id")) == fs_identifier or str(x.get("name")) == fs_identifier,
```
jgrund:
```suggestion
                lambda x: fs_identifier in [str(x.get("id")), str(x.get("name"))],
```

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1748,106 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+                ManagedFilesystem.objects.values("id", "name"),
+            )[0].get("id")
```
jgrund: This is making an assumption that the fs being looked up will always exist.

chroma_core/services/job_scheduler/job_scheduler.py
jgrund: Actually, This and the fs lookup below could be replaced by using `Q`.

```python
fs_id = configuration_data.get("filesystem_id")
fs = ManagedFilesystem.objects.filter(Q(id=fs_id) | Q(name=fs_id)).first()
# Check if fs is None here and return error?
configuration_data["filesystem_id"] = fs.id
```

jgrund: This also has the advantage of not returning the whole table, and the SQL generated is not bad:

```sql
SELECT "chroma_core_managedfilesystem"."id" FROM "chroma_core_managedfilesystem" WHERE ("chroma_core_managedfilesystem"."not_deleted" = True  AND ("chroma_core_managedfilesystem"."id" = 3  OR "chroma_core_managedfilesystem"."name" = fs )) ORDER BY "chroma_core_managedfilesystem"."id" ASC
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+    Command,
+)
+
+from chroma_api.chroma_model_resource import ChromaModelResource
+
+
+class RunStratagemValidation(Validation):
+    def is_valid(self, bundle, request=None):
+        purge_duration = bundle.data.get("purge_duration")
+        report_duration = bundle.data.get("report_duration")
```
jgrund: Check if purge / report duration is `None` and return error early.

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if "filesystem" not in bundle.data:
+            return {"code": "filesystem_required", "message": "filesystem required when running stratagem."}
```
jgrund:
```suggestion
            return {"code": "filesystem_required", "message": "filesystem required."}
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        elif isinstance(purge_duration, int) and isinstance(report_duration, int) and report_duration >= purge_duration:
```
jgrund: What happens if either of these are not ints? is that being validated? Perhaps cast both to ints (if they exist) and return an error if they are not.

jgrund:
```suggestion
        elif isinstance(purge_duration, int) and isinstance(report_duration, int) and report_duration > purge_duration:
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+            return {"code": "duration_order_error", "message": "Report duration must be less than purge duration."}
```
jgrund:
```suggestion
            return {"code": "duration_order_error", "message": "Report time must be less than purge time."}
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+
+        fs_identifier = str(bundle.data.get("filesystem"))
+
+        if not any(
```
jgrund:
```suggestion
if not ManagedFilesystem.objects.filter(Q(id= fs_identifier) | Q(name=fs_identifier)).exists():
  return {
                "code": "filesystem_does_not_exist",
                "message": "Filesystem {} does not exist.".format(fs_identifier),
            }
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        ):
+            return {
+                "code": "filesystem_does_not_exist",
+                "message": "Filesystem {} does not exist.".format(fs_identifier),
+            }
+
+        fs_id = filter(
+            lambda x: str(x.get("id")) == fs_identifier or str(x.get("name")) == fs_identifier,
+            ManagedFilesystem.objects.values("id", "name"),
+        )[0].get("id")
```
jgrund: This and the check about could be done with a single query (excuse poor formatting, you'll need to run black):

```python
fs_identifier = bundle.data.get("filesystem")
fs = ManagedFilesystem.objects.filter(Q(id=fs_id) | Q(name=fs_id)).values('id').first()

if fs is None:
 return {
                "code": "filesystem_does_not_exist",
                "message": "Filesystem {} does not exist.".format(fs_identifier),
            }

fs_id = fs.get(''id")
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Each MDT associated with the fielsystem must be installed on a server that has the stratagem profile installed
```
jgrund:
```suggestion
        # Each MDT associated with the filesystem must be installed on a server that has the stratagem profile installed
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        target_mount_ids = map(lambda mdt: mdt.active_mount_id, ManagedMdt.objects.filter(filesystem_id=fs_id))
```
jgrund:
```suggestion
        target_mount_ids = ManagedMdt.objects.filter(filesystem_id=1).values('active_mount_id')
```

jgrund: Note: the `active_mount_id`s may be `None`

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        host_ids = map(
```
jgrund:
```suggestion
        host_ids =  ManagedTargetMount.objects.filter(id__in=target_mount_ids).values('host_id')
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+            lambda target_mount: target_mount.host_id, ManagedTargetMount.objects.filter(id__in=target_mount_ids)
+        )
+        host_ids = pipe(host_ids, set, list)
+        installed_profiles = map(lambda host: host.server_profile_id, ManagedHost.objects.filter(id__in=host_ids))
```
jgrund:
```suggestion
        installed_profiles =  ManagedHost.objects.filter(id__in=host_ids).values_list('server_profile_id', flat=True)
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+            "purge_duration",
+            "purge_duration_active",
+        ]
+
+        difference = set(required_args) - set(bundle.data.keys())
+
+        if not difference:
+            return super(StratagemConfigurationValidation, self).is_valid(bundle, request)
+
+        return {"__all__": "Required fields are missing: {}".format(", ".join(difference))}
```
jgrund: Isn't `interval` the only required arg?

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+                "code": "stratagem_server_profile_not_installed",
+                "message": "'stratagem_servers' profile must be installed on all MDT servers.",
+            }
+
+        return {}
+
+
+class StratagemConfigurationValidation(RunStratagemValidation):
+    def is_valid(self, bundle, request=None):
+        required_args = [
```
jgrund: missing filesystem_name / fs_id?

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        if not difference:
+            return super(StratagemConfigurationValidation, self).is_valid(bundle, request)
+
+        return {"__all__": "Required fields are missing: {}".format(", ".join(difference))}
+
+
+class StratagemConfigurationResource(ChromaModelResource):
+    filesystem = fields.CharField(attribute="filesystem_id", null=False)
+    interval = fields.IntegerField(attribute="interval", null=False)
+    report_duration = fields.IntegerField(attribute="report_duration", null=False)
```
jgrund: Can't durations be null?

If they were null, doesn't that mean we don't need the active fields?

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        object_class = dict
+        validation = RunStratagemValidation()
+
+    def get_resource_uri(self, bundle=None, url_name=None):
+        return Resource.get_resource_uri(self)
+
+    @validate
+    def obj_create(self, bundle, **kwargs):
+        fs_identifier = str(bundle.data.get("filesystem"))
+        fs_id = filter(
```
jgrund:
```python
fs_id = ManagedFilesystem.objects.filter(Q(id=fs_identifier) | Q(name=fs_identifier)).values_list('id', flat=True).first()
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        authentication = AnonymousAuthentication()
+        object_class = dict
+        validation = RunStratagemValidation()
+
+    def get_resource_uri(self, bundle=None, url_name=None):
+        return Resource.get_resource_uri(self)
+
+    @validate
+    def obj_create(self, bundle, **kwargs):
+        fs_identifier = str(bundle.data.get("filesystem"))
```
jgrund: str cast not necessary.

chroma_api/stratagem.py
```diff
@@ -0,0 +1,145 @@
...
+        fs_id = filter(
+            lambda x: str(x.get("id")) == fs_identifier or str(x.get("name")) == fs_identifier,
+            ManagedFilesystem.objects.values("id", "name"),
+        )[0].get("id")
+
+        mdts = map(lambda mdt: mdt.id, ManagedMdt.objects.filter(filesystem_id=fs_id))
```
jgrund:
```suggestion
        mdts = ManagedMdt.objects.filter(filesystem_id=fs_id).values_list('id', flat=True)
```

chroma_core/lib/stratagem.py
```diff
@@ -0,0 +1,171 @@
...
+    "size >= 1t": "greater_than_equal_1t",
+}
+
+
+filter_out_other_counter = partial(filter, lambda counter: counter.get("name").lower() != "other")
+flatten = lambda xs: [item for l in xs for item in l]
+
+
+def tuple_to_equals(xs):
+    return "=".join((str(xs[0]), str(xs[1])))
```
jgrund: Couldn't this be a format string?

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1748,106 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+
+            configuration_data["filesystem_id"] = fs_id
+
+            with transaction.atomic():
```
jgrund: Since you are only doing a single db alteration in each branch, why do you need the atomic?

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1748,106 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+                matches = StratagemConfiguration.objects.filter(filesystem_id=fs_id)
```
jgrund: cnt = StratagemConfiguration.objects.filter(filesystem_id=fs_id).count()

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1748,106 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+        if not client_mount_exists:
+            self._create_client_mount(client_host, filesystem, mountpoint)
+
+        client_mount = LustreClientMount.objects.get(host_id=client_host.id)
+        client_mount.state = "unmounted"
+        client_mount.mountpoint = mountpoint
+        client_mount.filesystem_id = filesystem.id
+        client_mount.save()
+
+        run_stratagem_list.append(
```
jgrund: Does this need to be appended if the mount exists?

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,399 @@
...
+    StorageResourceRecord,
+)
+
+
+class StratagemConfiguration(StatefulObject):
+    filesystem_id = models.IntegerField(
+        help_text="The filesystem id associated with the stratagem configuration", null=False
+    )
+    interval = models.IntegerField(help_text="Interval value in seconds between each stratagem execution", null=False)
+    report_duration = models.IntegerField(
```
jgrund: Can't these be null? If so, do we need the active flags?

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,399 @@
...
+class RunStratagemStep(Step):
+    def run(self, args):
+        host = args["host"]
+        path = args["path"]
+        target_name = args["target_name"]
+        report_duration = args["report_duration"]
+        purge_duration = args["purge_duration"]
+
+        def calc_purge_duration(report_duration, purge_duration):
+            if isinstance(report_duration, int) and isinstance(purge_duration, int):
```
jgrund: What happens if they are not ints? should you cast after checking for existence?

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,399 @@
...
+                return "(atime < sys_time - {} && atime > sys_time - {})".format(report_duration, purge_duration)
```
jgrund: This needs to be in prefix notation

chroma_core/models/stratagem.py
```diff
@@ -0,0 +1,399 @@
...
+
+class RunStratagemStep(Step):
+    def run(self, args):
+        host = args["host"]
+        path = args["path"]
+        target_name = args["target_name"]
+        report_duration = args["report_duration"]
+        purge_duration = args["purge_duration"]
+
+        def calc_purge_duration(report_duration, purge_duration):
```
jgrund: Shouldn't this be calculating report duration not purge duration?

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: I've abstracted this logic as it's used in a few spots and i've updated it to no longer make this assumption. It will now return back the filesystem id or None.

chroma_api/stratagem.py
johnsonw: I attempt to convert these to ints and if it fails I return an error now.

johnsonw: I thought we determined they couldn't be the same. Did this change?

jgrund: Yes, you are correct.

chroma_api/stratagem.py
johnsonw: I abstracted this logic out from a previous comment so I will apply the enhancement there.

johnsonw: Actually, I tried using this and it fails with this:
```
ManagedFilesystem.objects.filter(Q(id=fs_identifier) | Q(name=fs_identifier))
*** ValueError: invalid literal for int() with base 10: 'fs18'
```
`fs_identifier` is 'fs18' but `id` expects an integer so it throws an exception.

johnsonw: any easy way around that?

chroma_api/stratagem.py
johnsonw: I added:
```
[x for x in target_mount_ids if x is not None]
```
To filter out any `None` values.

chroma_api/stratagem.py
jgrund: I guess you will need to try and parse to an int,  and handle the error if parsing fails.

chroma_api/stratagem.py
jgrund: What happens if all targets are unmounted?

chroma_api/stratagem.py
johnsonw: I'll update. At the time I thought these were the required fields.

chroma_api/stratagem.py
johnsonw: sure, works for me.

chroma_api/stratagem.py
johnsonw: I'm just think about this a bit more... If the target is unmounted then it's basically not going to be able to run stratagem. Would it be acceptable to say that if "None" is returned for any of these then return an error indicating that all mdt targets must be mounted?

chroma_api/stratagem.py
johnsonw: Ok, i've updated:
```
def get_fs_id_from_identifier(fs_identifier):
    try:
        id = int(str(fs_identifier), 10)
        return ManagedFilesystem.objects.filter(id=id).values("id").first()
    except ValueError:
        return ManagedFilesystem.objects.filter(name=fs_identifier).values("id").first()
```

chroma_api/stratagem.py
johnsonw: I've been casting this way since i'm trying to parse to an int. Basically:
```
(Pdb) int("7", 10)
7
(Pdb) int(7, 10)
*** TypeError: int() can't convert non-string with explicit base
(Pdb) int(str(7), 10)
7
```

johnsonw: but i've moved the cast so i'll remove it from here.

chroma_core/lib/stratagem.py
johnsonw:
```
"{}={}".format(str(xs[0]), str(xs[1]))
```
?

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: It doesn't make sense. I'll remove.

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: Yes, we want it to always try to mount on each run. If it's already mounted then it will come back and simply update the state.

chroma_core/models/stratagem.py
johnsonw: yes, updated

chroma_core/models/stratagem.py
johnsonw: Actually, I really just need to check if they are null or not. It's ok if they are strings.

chroma_core/models/stratagem.py
johnsonw: I feel like we discussed this not too long ago and decided on this format. What should I change it to?

jgrund: Does this actually work? If so then yeah, it's fine. I was under the impression that rules needed to be infix but perhaps I'm mistaken

chroma_core/models/stratagem.py
johnsonw: From our conversation on June 21'st:
```
(atime < sys_time - 5 && atime > sys_time - 10)
```
Maybe we can discuss again?

chroma_core/lib/stratagem.py
jgrund: Won't just passing to the format fn convert to a str representation?

chroma_core/models/stratagem.py
jgrund: Perhaps I'm mistaken but I think it's the report time that needs the upper bound.

`|--Report--||--Purge---|`

iml-manager-client/src/lib.rs
johnsonw: I had rebased on these changes earlier but changed it back to master. Not sure why it's in here.

chroma_core/lib/job.py
johnsonw: https://github.com/whamcloud/integrated-manager-for-lustre/pull/1025

settings.py
johnsonw: I was using it in three places in https://github.com/whamcloud/integrated-manager-for-lustre/blob/GH-852/chroma_core/lib/stratagem.py#L93

johnsonw: I can hardcode it if you prefer

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: I put it inside the lock.

johnsonw: I probably moved it out of the lock initially because in other spots, such as `test_host_contact`, the `advance` was moved outside the lock.

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: I think in every other place where we are adding to the cache it is being done inside a lock. Should I remove the lock?

settings.py
jgrund: Not commenting on DRYness but the fact that this setting is not configuring influxdb which is different than all the other settings in this file.

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: This is now using the `get_fs_id_from_identifier` so if fs_identifier was `None` it would simply return `None` for fs_id and the function would return. Here is output from a test I ran:

```
69 	            # The filesystem_id may come in as the fs name or the fs id. In terms of storing information in the database, the fs id should always be used.
1770 	            fs_identifier = str(configuration_data.get("filesystem_id"))
1771 	            from remote_pdb import RemotePdb
1772 	            RemotePdb('127.0.0.1', 4444).set_trace()
1773 ->	            fs_id = get_fs_id_from_identifier(fs_identifier)
1774 	
1775 	            if not fs_id:
1776 	                return None
1777 	
1778 	            configuration_data["filesystem_id"] = fs_id
(Pdb) fs_identifier
'fs23'
(Pdb) fs_identifier = None
(Pdb) fs_identifier
(Pdb) n
> /usr/share/chroma-manager/chroma_core/services/job_scheduler/job_scheduler.py(1775)configure_stratagem()
-> if not fs_id:
(Pdb) fs_id
```

However, i'm not sure that this scenario is possible as the validator would ensure the filesystem exists before creating the bundle data and invoking the JobSchedulerClient.configure_stratagem function.

chroma_api/stratagem.py
johnsonw: Based on our discussion i've updated this to the following:
```
        target_mount_ids = (
            ManagedMdt.objects.filter(filesystem_id=fs_id, active_mount_id__isnull=False)
            .values_list("active_mount_id", flat=True)
            .distinct()
        )
        mdt0 = ManagedMdt.objects.filter(filesystem_id=fs_id, name__contains="MDT0000").first()

        if mdt0 is None:
            return {"code": "mdt0_not_found", "message": "MDT0 could not be found."}

        if mdt0.active_mount_id not in target_mount_ids:
            return {"code": "mdt0_not_mounted", "message": "MDT0 must be mounted in order to run stratagem."}
```

chroma_core/lib/stratagem.py
johnsonw: I tried a few things but not sure i'm following. 
```
>>> "{}={}".format(x)
  File "<stdin>", line 1, in <module>
IndexError: tuple index out of range
```

chroma_core/models/stratagem.py
johnsonw: I've change this to: 
```
return "(&& < atime - sys_time {} > atime - sys_time {})".format(report_duration, purge_duration)
```

chroma_core/models/stratagem.py
johnsonw: correct, i've upated.

chroma_core/lib/stratagem.py
jgrund: `"{}={}".format(*x)`

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: Will raise error.

settings.py
johnsonw: will hardcode

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -872,8 +880,10 @@ def notify(self, content_type, object_id, time_serialized, update_attrs, from_st
     def run_jobs(self, job_dicts, message):
         with self._lock:
             result = self.CommandPlan.command_run_jobs(job_dicts, message)
+
             self.progress.advance()
```
jgrund: Does this need to be within the lock?

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1755,104 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+                "purge_duration": stratagem_data.get("purge_duration"),
+            }
+
+            # The filesystem_id may come in as the fs name or the fs id. In terms of storing information in the database, the fs id should always be used.
+            fs_identifier = str(configuration_data.get("filesystem_id"))
+            fs_id = get_fs_id_from_identifier(fs_identifier)
+
+            if not fs_id:
+                from tastypie.exceptions import NotFound
+                raise NotFound("No matching filesystem for {}".format(fs_identifier))
```
jgrund: This should be a generic error, not a Tastypie one. What happens if this is called outside the API?

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: took outside of the block.

johnsonw:
rust files got moved in during the rebase. They should no longer be in there.

chroma_help/help.py
```diff
@@ -14,6 +14,10 @@
     "start_copytool": "Start this copytool to begin processing HSM requests.",
     "stop_copytool": "Stop this copytool. In-progress HSM actions will be canceled.",
     "configure_copytool": "Configure this copytool on the worker.",
+    "configure_stratagem": "Configure's Stratagem on the selected node.",
```
jgrund:
```suggestion
    "configure_stratagem": "Configures Stratagem for a future scan.",
```

chroma_help/help.py
```diff
@@ -14,6 +14,10 @@
...
+    "configure_stratagem_long": "Configure's Stratagem on the selected node.",
```
jgrund:
```suggestion
    "configure_stratagem_long": "Configures Stratagem for a future scan.",
```

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1745,3 +1755,103 @@ def update_lnet_configuration(self, lnet_configuration_list):
...
+            # The filesystem_id may come in as the fs name or the fs id. In terms of storing information in the database, the fs id should always be used.
+            fs_identifier = str(configuration_data.get("filesystem_id"))
+            fs_id = get_fs_id_from_identifier(fs_identifier)
+
+            if not fs_id:
+                raise Exception("No matching filesystem for {}".format(fs_identifier))
+
+            configuration_data["filesystem_id"] = fs_id
+
+            matches = StratagemConfiguration.objects.filter(filesystem_id=fs_id)
```
jgrund: I think this would be clearer as

```python
try:
    stratagem_configuration = StratagemConfiguration.objects.get(filesystem_id=fs_id)
    stratagem_configuration.update(**configuration_data)
except StratagemConfiguration.DoesNotExist:
    stratagem_configuration = StratagemConfiguration.objects.create(**configuration_data)
```

chroma_api/stratagem.py
```diff
@@ -0,0 +1,155 @@
...
+        try:
+            report_duration = bundle.data.get("report_duration") and get_bundle_int_val(
+                bundle.data.get("report_duration")
+            )
+        except ValueError:
+            return {"code": "invalid_argument", "message": "Report duration must be an integer value."}
+
+        if "filesystem" not in bundle.data:
+            return {"code": "filesystem_required", "message": "Filesystem required."}
+        elif purge_duration and report_duration and report_duration >= purge_duration:
```
jgrund: Isn't it that the report time must be larger than the warn time?

chroma_core/services/job_scheduler/job_scheduler.py
johnsonw: In the try, shouldn't we update and then pull the data again to have the new object before adding it to cache?

jgrund: Does that make a difference? I would have expected .save() to be called, but didn't see it happening before.

johnsonw: Maybe something like:
```
try:
    stratagem_configuration = StratagemConfiguration.objects.get(filesystem_id=fs_id)
    stratagem_configuration.update(**configuration_data)
    stratagem_configuration = StratagemConfiguration.objects.get(filesystem_id=fs_id)
except StratagemConfiguration.DoesNotExist:
    stratagem_configuration = StratagemConfiguration.objects.create(**configuration_data)
```

PR #897, opened 2019-04-29
Fix discarding of target mount events

AlexTalker:
Consider that by any reason an `OST`(or any other part of `Lustre` which has a mount point) is unmounted on `primary` node and mounted on `secondary`(i.e. a `fail-over` performed).
In monitored mode, from time to time, target is left in `Offline` state.
This seems to be because selection of updates is based on the time when `device_plugins/lustre.py` was run on the agent side. However, performance of this code or the time it run at is not synchronized in any way.
Thus, it is possible when event of `fail-over` discovered on `passive` node during run which started earlier than on `primary` node. In this case, if event from `primary` node arrived first, discovered information is discarded. To prevent it, since handling of `Lustre` information is handled by a `single-thread` application and a queue, order of updates must be defined by order in which data is arrived to server, i.e. in order it were processed on server-side.


AlexTalker:
@jgrund @utopiabound Would you kindly review this?

jgrund:
test this please

jgrund:
test this please

jgrund:
test this please

jgrund:
test this please

AlexTalker:
In context of https://github.com/whamcloud/integrated-manager-for-lustre/pull/917 , and after improving speed of `Shell.run`, this problem doesn't reproduce anymore so this might be an overhead. However, I suggest you to consider that starting time of scanner on agent site might differentiate and this way lead to the situation of dropping necessary data. Interval in `10` minutes before forced update is quite high.

jgrund:
Jenkins: trigger a test run

AlexTalker:
After https://github.com/whamcloud/iml-common/pull/60 this change seems to be unnecessary.
It is better to focus on https://github.com/whamcloud/integrated-manager-for-lustre/pull/917

PR #936, opened 2019-05-10
Load power-control service after rabbitmq-server.service

johnsonw:
- Load power-control service after rabbitmq-server.service

### DIFF
--- a/iml-power-control.service
+++ b/iml-power-control.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=IML Power Control Service
 PartOf=iml-manager.target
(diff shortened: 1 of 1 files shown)


johnsonw:
I'll add that in and test locally.

johnsonw:
After rebooting the node the iml-power-control.service started up correctly, but the iml-update-handler had a cycle issue. This was caused by https://github.com/whamcloud/device-scanner/pull/262 and https://github.com/whamcloud/srcmap-reverse/pull/84. Once these were fixed and the node was rebooted all services loaded as expected.

PR #939, opened 2019-05-13
Override rabbitmq-server.service

jgrund:
so that `NotifyAccess=all`


jgrund:
> 10-rabbitmq-server.service needs to be added to MANIFEST.in

How come? It's being added  as a `SOURCE`

PR #943, opened 2019-05-20
Fix volumes not appearing on volumes page

johnsonw:
Fixes GUI#201

Description:

When the volumes page loads it makes an asynchronouse call to https://localhost:7443/api/volume?limit=0&category=unused.
The problem is that nginx proxies this request over to gunicorn, where
it then generates a 301 to http://localhost:7443/api/volume/?limit=0&category=unused. This would normally be ok as when it
get sent back to nginx it would 302 to https://localhost:7443/api/volume/?limit=0&category=unused, but because the url changed
from https to http the csp is not satisfied and the request is then
canceled. I belive there are two things that need to be done to solve
this issue:

1. redirect api calls to always have a trailing slash before being
proxied to gunicorn
2. Update existing api calls that do not have a trailing slash

This patch addresses the first item.

### DIFF
--- a/chroma-manager.conf.template
+++ b/chroma-manager.conf.template
@@ -149,6 +149,8 @@ server {
         proxy_set_header Host $http_host;
         proxy_set_header X-Forwarded-Proto $scheme;
         proxy_set_header X-Forwarded-Server $host;
+
+        rewrite ^(.*[^/])$ $1/ permanent;
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_pass {{HTTP_API_PROXY_PASS}}/api;
     }


jgrund:
Can you explain why this happens in Docker but not outside containers?

johnsonw:
@jgrund , I'm putting documentation together now that will hopefully explain this.

johnsonw:
Not the absolute core cause but I do have more information as to why this happens in docker and not vagrant. Running a curl request in both the nginx and gunicorn containers clearly shows that gunicorn receives the message but the 301 (generated by gunicorn) points to http, not https. In vagrant, the `Location` header points to https in the 301. This is why we don't see the error in vagrant, but I still need to find out why the 301 doesn't point to https in docker. Here is some output and the commands I ran both in the docker containers and the vagrant node:

curl commands:

```
# curling api endpoint from nginx
curl -k -i -L \
-H "accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3" \
-H "accept-encoding: gzip, deflate, br" \
-H "accept-language: en-US,en;q=0.9" \
-H "cache-control: no-cache" \
-H "upgrade-insecure-requests: 1" \
-H "X-Forwarded-Proto: https" \
https://127.0.0.1:443/api/volume?limit=0&category=unused

# Curling api endpoint from gunicorn
curl -k -i -L \
-H "accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3" \
-H "accept-encoding: gzip, deflate, br" \
-H "accept-language: en-US,en;q=0.9" \
-H "cache-control: no-cache" \
-H "upgrade-insecure-requests: 1" \
-H "X-Forwarded-Proto: https" \
http://127.0.0.1:8001/api/volume?limit=0&category=unused
```

Docker containers:
```
From nginx:

/ # HTTP/2 301 
server: nginx/1.15.12
date: Mon, 20 May 2019 20:56:32 GMT
content-type: text/html; charset=utf-8
location: http://nginx:7443/api/volume/?limit=0

HTTP/1.1 302 Moved Temporarily
Server: nginx/1.15.12
Date: Mon, 20 May 2019 20:56:32 GMT
Content-Type: text/html
Content-Length: 146
Connection: close
Location: https://nginx:7443/api/volume/?limit=0

HTTP/2 200 
server: nginx/1.15.12
date: Mon, 20 May 2019 20:56:32 GMT
content-type: application/json
vary: Accept, Cookie
cache-control: no-cache

From gunicorn:

 HTTP/1.1 301 MOVED PERMANENTLY
Server: gunicorn/19.9.0
Date: Mon, 20 May 2019 20:57:56 GMT
Connection: keep-alive
Transfer-Encoding: chunked
Content-Type: text/html; charset=utf-8
Location: http://gunicorn:8001/api/volume/?limit=0

HTTP/1.1 200 OK
Server: gunicorn/19.9.0
Date: Mon, 20 May 2019 20:57:56 GMT
Connection: keep-alive
Transfer-Encoding: chunked
Vary: Accept, Cookie
Content-Type: application/json
Cache-Control: no-cache
```

Vagrant:

```
From nginx:

HTTP/1.1 301 MOVED PERMANENTLY
Server: nginx/1.12.2
Date: Mon, 20 May 2019 21:16:01 GMT
Content-Type: text/html; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Location: https://127.0.0.1/api/volume/?limit=0

HTTP/1.1 200 OK
Server: nginx/1.12.2
Date: Mon, 20 May 2019 21:16:01 GMT
Content-Type: application/json
Transfer-Encoding: chunked
Connection: keep-alive
Vary: Accept, Cookie
Cache-Control: no-cache


From gunicorn:
[root@adm conf.d]# HTTP/1.1 301 MOVED PERMANENTLY
Server: gunicorn/18.0
Date: Mon, 20 May 2019 21:13:45 GMT
Connection: keep-alive
Transfer-Encoding: chunked
Content-Type: text/html; charset=utf-8
Location: https://127.0.0.1:8001/api/volume/?limit=0

curl: (35) Encountered end of file
```

johnsonw:
It's also worth noting that on vagrant gunicorn appears to listen to the `-H "X-Forwarded-Proto: https" \` header. This means that the 301 gets routed to https. In docker, it seems to ignore this header and redirect to http.

jgrund:
@johnsonw So that shows that it *is* happening, but still doesn't explain why one environment gets a 301 and the other gets a 302.

Given that, It seems like this is more of a configuration problem. Have you tried matching the versions on both gunicorn and nginx?

johnsonw:
@jgrund I'm recording the environment variables so I can compare them across both environments. Once I'm done comparing them I will check the versions.

johnsonw:
Just a quick update that I updated the gunicorn container to use 18.0 as is used in vagrant and still see the same issue.

jgrund:
Going to investigate difference between docker stack / compose and non-containerized environments

Issue #895, opened 2019-04-26
SSI: VolumeNode matching query does not exist


jgrund:
In SSI testing, we see the following issue:

```
description: Update managed target mounts for target testfs-OST0000
class_name: UpdateManagedTargetMount
backtrace: Traceback (most recent call last):
  File "/usr/share/chroma-manager/chroma_core/services/job_scheduler/job_scheduler.py", line 326, in _run
    result = step.run(args)
  File "/usr/share/chroma-manager/chroma_core/models/target.py", line 1443, in run
    expected_exception_classes=[VolumeNode.DoesNotExist],
  File "/usr/lib/python2.7/site-packages/iml_common/lib/util.py", line 220, in wait_for_result
    return lambda_expression()
  File "/usr/share/chroma-manager/chroma_core/models/target.py", line 1440, in <lambda>
    lambda: VolumeNode.objects.get(host=host, path=filesystem.mount_path(target.name)),
  File "/usr/lib/python2.7/site-packages/django/db/models/manager.py", line 151, in get
    return self.get_queryset().get(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/django/db/models/query.py", line 310, in get
    self.model._meta.object_name)
DoesNotExist: VolumeNode matching query does not exist.
```

Digging into this a bit, the `VolumeNode` *does* exist, but it has been marked deleted:

```
chroma=> select * from chroma_core_volumenode where host_id = 3 and id = 24;
 id | volume_id | host_id |                       path                       | storage_resource_id | primary | use | not_deleted
----+-----------+---------+--------------------------------------------------+---------------------+---------+-----+-------------
 24 |         3 |       3 | /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_target2 |                     | t       | t   |
```

Further the device-aggregator knows about this record when the bug occurs, which points to the bug being within the resource manager.

![Screen Shot 2019-04-26 at 4 15 54 PM](https://user-images.githubusercontent.com/458717/56834411-4e748c00-683f-11e9-8e86-42254b22cb5b.png)

jgrund:
Temporary workaround if seen outside test:

```
systemctl restart iml-manager.target
```

Causes a re-read of devices

AlexTalker:
@jgrund I am working on this problem right now.
On my installation it can be reproduced in monitoring-only mode.
[...]

PR #946, opened 2019-05-21
Turn up SSI debugging

utopiabound:
Turn on extra debugging for Issue #895 

COPR Module: managerforlustre/device-scanner-devel

### DIFF
--- a/settings.py
+++ b/settings.py
@@ -240,7 +240,7 @@ CRYPTO_FOLDER = "/var/lib/chroma"
 
 GUNICORN_PID_PATH = "/var/run/gunicorn.pid"
 
-LOG_LEVEL = logging.INFO
+LOG_LEVEL = logging.DEBUG
 
 EMAIL_HOST = None
 EMAIL_SUBJECT_PREFIX = "[Chroma Server]"


settings.py
```diff
@@ -234,7 +234,7 @@
 
 GUNICORN_PID_PATH = "/var/run/gunicorn.pid"
 
-LOG_LEVEL = logging.INFO
```
jgrund: You should add this as a value under `local_settings.py` so it doesn't get persisted to production.

utopiabound: This is for debug, not meant to actually land

jgrund: ah, ok. That said, it may not be a bad thing to have this always be there in `local_settings.py`

utopiabound: Actually we already do: https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/tests/framework/integration/shared_storage_configuration/full_cluster/cluster_setup#L81-L85

And it's present and shows up via `python manage.py diffsettings|grep LOG`

But I'm not seeing DEBUG logging from all plugins/jobs.

jgrund: It think this only applies for the first test though. At some point this file is getting removed, I think when we re-install the manager between tests?

Issue #702, opened 2018-08-03
Add MLNX_OFED managed ldiskfs profile


jgrund:
Add support for MLNX_OFED in a managed ldiskfs capacity.

This will involve the following changes on IML side.

  - Add metapackage to https://github.com/whamcloud/lustre-ldiskfs-zfs/blob/81f1287d2a80631b2f6893c624039b56677abcc4/lustre-ldiskfs-zfs.spec.in that will install the proper packages for a MLNX_OFED ldiskfs setup.
  - Add the repo for prebuilt RPMs to https://github.com/whamcloud/integrated-manager-for-lustre/blob/9da817af17623d1092067f06f3453de79795a616/storage_server.repo.in
  - Ensure that we enable / disable the correct repos at install time dependent on MLNX being selected or not.

Caveats for first version:

  - We are using pre-built rpms.
  - Only supporting ldiskfs in managed mode.

@kbuggenhout @e73kiel, following on from #651, will this be useful for you?

PR #955, opened 2019-05-23
Add Managed MLNX Profile

utopiabound:
This adds a profile that will support Mellanox OFED.

Fixes Issue #702 


utopiabound:
Manual testing on VMs has this install correctly.

PR #958, opened 2019-05-25
Make docker publish manual process

jgrund:
Make publishing docker containers a manual process, so we don't build
new containers for every master landing.


.travis.yml
```diff
@@ -109,5 +107,7 @@ stages:
   - test
   - name: cd
```
utopiabound: Does this need to be defined as a stage somewhere?

jgrund: It is on: https://github.com/whamcloud/integrated-manager-for-lustre/pull/958/files#diff-354f30a63fb0907d4ad57269548329e3R71

just hidden by default by the github review UI

PR #967, opened 2019-05-30
Add initial rust manager cli

jgrund:
This is the bare bones needed for a rust cli against the manager API.

It uses API key auth so no password config file is needed.

As a POC I've implemented a tabular list for the host endpoint.


jgrund:
Screen Shot 2019-05-30 at 3 55 58 PM

PR #984, opened 2019-06-06
Wait on jobs

johnsonw:
There is currently no way for the job indicator to indicate that a
specific job should wait on other jobs within the same command. This
patch adds the ability to indicate that a particular job should wait on
other jobs that are being defined within the command by their index. An
example would be:

```
run_stratagem_list = map(lambda mdt_id: {
    "class_name": "RunStratagemJob",
    "args": {
        "mdt_id": mdt_id,
        "uuid": unique_id
    }
}, mdts)

run_stratagem_list.append({
    "class_name": "SendStratagemResultsToClientJob",
    "args": {
        "depends_on_job_range": str(range(len(mdts)))
    }
})
```


chroma_core/models/jobs.py
```diff
@@ -331,6 +331,9 @@ def __hash__(self):
     states = ("pending", "tasked", "complete")
     state = models.CharField(max_length=16, default="pending", help_text="One of %s" % (states,))
 
+    depends_on_job_range = models.CharField(max_length=64, default="[]", help_text="Range of jobs in list that must\
```
jgrund: Why does this need to be saved to the db?

johnsonw: Looking at it further I think I can do it without storing the data. I'll update.

johnsonw:
Yes, addressing these now.

johnsonw:
This should pass now.

jgrund:
In SSI I'm seeing: `AssertionError: {u'jobs': [u'/api/job/9/', u'/api/job/10/', u'/api/job/11/', u'/api/job/12/', u'/api/job/13/', u'/api/job/14/', u'/api/job/15/', u'/api/job/16/', u'/api/job/17/', u'/api/job/18/'], u'complete': True, u'created_at': u'2019-06-11T03:14:46.629075', u'message': u'Setup managed host vm5', u'cancelled': True, u'errored': True, u'resource_uri': u'/api/command/9/', u'id': 9, u'logs': u'\n\n\n\n\n'}
`

It looks like complete, cancelled and errored are all true

PR #986, opened 2019-06-10
Add support for `LAT_ATTR_CLASSIFY` rules

jgrund:
This patch adds support for parsing and displaying
`LAT_ATTR_CLASSIFY` rules.

### DIFF
--- a/iml-agent/src/action_plugins/stratagem/server.rs
+++ b/iml-agent/src/action_plugins/stratagem/server.rs
@@ -324,6 +324,7 @@ pub fn stream_fidlists(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::collections::HashMap;
 
     #[test]
     fn test_get_fid_dirs() {
--- a/iml-agent/src/action_plugins/stratagem/server.rs
+++ b/iml-agent/src/action_plugins/stratagem/server.rs
@@ -324,7 +324,6 @@ pub fn stream_fidlists(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::collections::HashMap;
 
     #[test]
     fn test_get_fid_dirs() {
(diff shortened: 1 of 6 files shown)


jgrund:
New Content:

Screen Shot 2019-06-10 at 12 28 53 PM

iml-agent/src/action_plugins/stratagem/server.rs
```diff
@@ -26,71 +28,153 @@ impl StratagemGroup {
...
+///
+/// Exposes their common properties as trait methods.
+pub trait Counter {
+    fn name(&self) -> &str;
+    fn count(&self) -> u64;
     fn have_flist(&self) -> bool;
 }
 
-impl HaveFlist for StratagemCounter {
+impl Counter for &StratagemCounter {
```
johnsonw: One thing we could consider doing here is using a macro (perhaps this is overkill but i'll mention it any way). As an example:

```
macro_rules! impl_Counter {
    ($T:ident) => {
        impl Counter for $T {
            fn name(&self) -> &str {
                &self.name
            }

            fn count(&self) -> u64 {
                self.count
            }
        }
    }
}
```

Then call the macro to define the implementation for each counter type:

```
impl_Counter!(&StratagemCounter)
impl_Counter!(&StratagemClassifyCounter)
```

I thought this might be a way to dry it up a bit but maybe it's overkill. What are your thoughts?

jgrund: I'd prefer to stay away from implementing our own macros unless there is a really clear benefit to doing so. I think they have a number of downsides.

johnsonw: ok, i'll keep that in mind.

Issue #988, opened 2019-06-11
Setup Grafana


jgrund:
Setup grafana

https://grafana.com/docs/installation/rpm/

utopiabound:
Link to grafana embeded charts:
https://grafana.com/docs/reference/sharing/

PR #996, opened 2019-06-14
Grafana

utopiabound:
Related to issue #988 

Does the following:
* Ensure grafana install
* Setup grafana
* Adjust setup of influx to ensure future migration of postgres data into influx works
* Adjust postgres so grafana can use postgres backing (seperate from postgres datasource)

### DIFF
--- a/grafana-iml.ini
+++ b/grafana-iml.ini
@@ -1,6 +1,6 @@
 [database]
 type = postgres
-host = localhost
+host = /var/run/postgresql/
 name = grafana
 user = chroma
 ssl_mode = disable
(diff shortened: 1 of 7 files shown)


grafana-iml.ini
```diff
@@ -0,0 +1,31 @@
+[database]
+type = postgres
+host = localhost
```
jgrund: possible to connect via socket?

utopiabound: I found it: https://github.com/grafana/grafana/issues/2553
I'll update, and pull out all the pgsql changes

PR #999, opened 2019-06-18
Update influx db name for stratagem scans

johnsonw:
Separate databases will be required in order to manage retention between
stats and stratagem scans. Create a unique database for stratagem scans.

### DIFF
--- a/chroma_core/lib/service_config.py
+++ b/chroma_core/lib/service_config.py
@@ -313,7 +313,7 @@ class ServiceConfig(CommandLine):
         )
 
         log.info("Creating InfluxDB databse...")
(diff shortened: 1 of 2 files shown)


chroma_core/lib/service_config.py
```diff
@@ -313,7 +313,7 @@ def _setup_influxdb(self):
         )
 
         log.info("Creating InfluxDB databse...")
-        self.try_shell(["influx", "-execute", "CREATE DATABASE iml"])
+        self.try_shell(["influx", "-execute", "CREATE DATABASE {}".format(settings.INFLUXDB_STRATAGEM_SCAN_DB)])
```
jgrund: Don't we want both?

PR #1003, opened 2019-06-19
Add Grafana reverse proxy to nginx

utopiabound:
Issue #988 

Add /api/auth that retuns true (200) or false (401) if user is logged in.
Add proxy via nginx for grafana
Add influx datasource reader to grafana on install


chroma_core/lib/service_config.py
```diff
@@ -595,6 +635,8 @@ def _setup_database(self, check_db_space):
 
         else:
             log.info("Postgres already accessible")
+            self._config_pgsql_auth(database)
```
johnsonw: Would this be databases["default"]?

PR #1008, opened 2019-06-20
Ensure ntpd is running

jgrund:
with the advent of `chronyd`, `ntpd`, which IML currently relies on may
be inactive.

Make sure we stop and disable chronyd on the manager.

### DIFF
--- a/chroma_core/lib/service_config.py
+++ b/chroma_core/lib/service_config.py
@@ -471,7 +471,7 @@ class ServiceConfig(CommandLine):
             return error_msg
 
     def _restart_pgsql(self):
-        postgresql_service = ServiceControlEL7.create("postgresql")
+        postgresql_service = ServiceControl.create("postgresql")
         postgresql_service.restart()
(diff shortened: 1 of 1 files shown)


chroma_core/lib/service_config.py
```diff
@@ -467,7 +471,7 @@ def _check_db_space(self, required_space_gigabytes):
             return error_msg
 
     def _restart_pgsql(self):
-        postgresql_service = ServiceControl.create("postgresql")
+        postgresql_service = ServiceControlEL7.create("postgresql")
```
jgrund:
```suggestion
        postgresql_service = ServiceControl.create("postgresql")
```

johnsonw: Curious as to why this was reverted?

jgrund: It was causing an error in EFS.

johnsonw: 👍

PR #1020, opened 2019-06-26
Update command to account for stratagem status

johnsonw:
Update command to account for stratagem status


iml-manager-cli/src/main.rs
```diff
@@ -3,12 +3,15 @@
...
 use futures::Future;
+
 use iml_manager_cli::{
     api_utils,
     manager_cli_error::{DurationParseError, ImlManagerCliError},
 };
+
 use iml_wire_types::{ApiList, Command, Host};
 use prettytable::{Row, Table};
+use reqwest as _;
```
jgrund: Dead code?

johnsonw: I think it's being used as my compiler isn't issuing a warning.

jgrund: I don't see how it could be... it's not a trait.

iml-manager-cli/Cargo.toml
```diff
@@ -19,6 +19,7 @@ structopt = "0.2"
 termion = "1"
 tokio = "0.1"
 exitcode = "1.1"
+reqwest = { version = "0.9", features = ["default-tls", "tls"] }
```
jgrund: Do you need this dep?

johnsonw: I may not any more, i'll remove and re-compile.

iml-manager-cli/src/main.rs
johnsonw: I removed it and I don't see any warnings or errors.

iml-manager-cli/src/main.rs
```diff
@@ -3,12 +3,15 @@
 // license that can be found in the LICENSE file.
 
 use futures::Future;
+
 use iml_manager_cli::{
     api_utils,
     manager_cli_error::{DurationParseError, ImlManagerCliError},
 };
+
```
jgrund:
```suggestion
```

iml-manager-cli/src/main.rs
```diff
@@ -3,12 +3,15 @@
 // license that can be found in the LICENSE file.
 
 use futures::Future;
+
```
jgrund:
```suggestion

```

iml-manager-cli/src/main.rs
```diff
@@ -172,20 +191,60 @@ fn main() {
                             purge_duration: pd
                         }),
                     )
+                    .and_then(|resp| iml_manager_client::concat_body(resp))
+                    .map(|(resp, chunk)| {
```
jgrund:
```suggestion
                    .map(|(resp, body)| {
```

iml-manager-cli/src/main.rs
```diff
@@ -148,6 +151,22 @@ struct CmdWrapper {
     command: Command,
 }
 
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
+#[serde(rename_all = "snake_case")]
+enum RunStratagemCommandResult {
```
jgrund: Why does this not use the Error trait?

iml-manager-cli/src/main.rs
```diff
@@ -172,20 +191,60 @@ fn main() {
...
+                        stop_spinner(None);
 
-                display_cmd_state(&command);
+                        display_cmd_state(&command);
 
-                exit(exitcode::OK)
+                        exit(exitcode::OK)
+                    }
+                    Err(validation_message) => {
+                        let command = Command {
```
jgrund: Instead of constructing a virtual command just to display the message, you should refactor `display_cmd_state` so that it calls different message fns for success / failure etc...

iml-manager-cli/src/main.rs
```diff
@@ -172,20 +191,60 @@ fn main() {
                             purge_duration: pd
                         }),
                     )
+                    .and_then(|resp| iml_manager_client::concat_body(resp))
+                    .map(|(resp, chunk)| {
+                        let status = resp.status();
+                        match status.as_u16() {
```
jgrund: Instead of matching over status code ranges, this would be more clear if you used the methods exposed on `StatusCode` with if ... else. https://docs.rs/reqwest/0.9.18/reqwest/struct.StatusCode.html#method.is_success, https://docs.rs/reqwest/0.9.18/reqwest/struct.StatusCode.html#method.is_client_error

iml-manager-cli/src/main.rs
johnsonw: I'll add

jgrund: This should be nested inside `ImlManagerClientError` as well.

iml-manager-cli/src/main.rs
johnsonw: sure

iml-manager-cli/src/main.rs
jgrund: Instead of using `map` here, using `and_then` would allow the Error to propagate to the error side of the channel.

Then the Result is lifted up to the future.

johnsonw:
Some output:

```
[root@adm x86_64]#  iml stratagem scan --filesystem fs1 -p5d -r0d
✗ Filesystem None does not exist. errored
[root@adm x86_64]#  iml stratagem scan --filesystem fs17 -p0d -r5d
✗ Report duration must be less than purge duration. errored
[root@adm chroma-manager]#  iml stratagem scan --filesystem fs17 -p5d -r0d
✗ Internal server error. errored
[root@adm chroma-manager]#  iml stratagem scan --filesystem fs17 -p5d -r0d
✗ 'stratagem_servers' profile must be installed on all MDT servers. errored
[root@adm chroma-manager]#  iml stratagem scan --filesystem fs18 -p5d -r0d
✔ Stratagem: Scanning all MDT's successful
```

johnsonw:
I've addressed all but the last comment. I'd like to discuss this one a bit if possible.

iml-manager-cli/src/main.rs
```diff
@@ -129,17 +129,24 @@ fn start_spinner(msg: &str) -> impl FnOnce(Option<String>) -> () {
     }
 }
 
-fn display_cmd_state(cmd: &Command) {
+fn display_cmd_state(cmd: &Result<Command, iml_manager_client::ImlManagerClientError>) {
```
jgrund: This fn shouldn't be changed, A cmd is not going to be wrapped in an Error.

iml-manager-cli/src/main.rs
```diff
@@ -172,20 +179,51 @@ fn main() {
                             purge_duration: pd
                         }),
                     )
+                    .and_then(|resp| iml_manager_client::concat_body(resp))
+                    .and_then(|(resp, body)| {
+                        let status = resp.status();
+                        if status.is_success() {
+                            Ok(serde_json::from_slice::<CmdWrapper>(&body).expect("Couldn't parse CmdWrapper."))
```
jgrund: Why unwrap here?

iml-manager-cli/src/main.rs
```diff
@@ -172,20 +179,51 @@ fn main() {
...
+                        } else if status.is_client_error() {
+                            let validation_message: iml_manager_client::RunStratagemValidationError =
+                                    serde_json::from_slice(&body).expect("Could not parse chunk.");
```
jgrund: Why unwrap here?

iml-manager-cli/src/main.rs
```diff
@@ -172,20 +179,51 @@ fn main() {
...
+
+                            Err(validation_message)?
+                        } else if status.is_server_error() {
+                            Err(iml_manager_client::RunStratagemValidationError {
+                                code: iml_manager_client::RunStratagemCommandResult::ServerError,
+                                message: "Internal server error.".to_string(),
```
jgrund:
```suggestion
                                message: "Internal server error.".into(),
```

iml-manager-cli/src/main.rs
```diff
@@ -172,20 +179,51 @@ fn main() {
...
+                            })?
+                        } else {
+                            Err(iml_manager_client::RunStratagemValidationError {
+                                code: iml_manager_client::RunStratagemCommandResult::UnknownError,
+                                message: "Unknown error.".to_string(),
```
jgrund:
```suggestion
                                message: "Unknown error.".into(),
```

iml-manager-client/src/lib.rs
```diff
@@ -20,6 +26,8 @@ impl std::fmt::Display for ImlManagerClientError {
             ImlManagerClientError::Reqwest(ref err) => write!(f, "{}", err),
             ImlManagerClientError::InvalidHeaderValue(ref err) => write!(f, "{}", err),
             ImlManagerClientError::UrlParseError(ref err) => write!(f, "{}", err),
+            ImlManagerClientError::SerdeJsonError(ref err) => write!(f, "{}", err),
+            ImlManagerClientError::RunStratagemValidationError(ref err) => write!(f, "{}", err),
```
jgrund: I think this should live under `ImlManagerCliError` instead actually. It seems to specific for the client at this point.

iml-manager-cli/src/main.rs
johnsonw: Maybe I didn't understand. The previous comment said to refactor:
> Instead of constructing a virtual command just to display the message, you should refactor display_cmd_state so that it calls different message fns for success / failure etc...

iml-manager-cli/src/main.rs
johnsonw: removed.

iml-manager-cli/src/main.rs
johnsonw: removed

iml-manager-client/src/lib.rs
johnsonw: I'm not sure I understand. Do I need to remove the `RunStratagemValidationError` type?

iml-manager-cli/src/main.rs
johnsonw: I reverted this function and created a new function to handle the validation error case.

jgrund: Instead of reusing `display_cmd_state` for things not command related, take the `println`s and make separate functions out of them:

`display_error(...)`, `display_success(...)`

Then when handling an API error that's not related to a command, you can call those fns directly and not conflate that error with a command.

iml-manager-client/src/lib.rs
jgrund: Move the validation error type into `ImlManagerCliError`

PR #1025, opened 2019-07-01
Update invoke_rust_agent_expect_result to handle various exceptions

johnsonw:


chroma_core/lib/job.py
```diff
@@ -200,7 +200,15 @@ def invoke_rust_agent(self, host, command, args={}):
     def invoke_rust_agent_expect_result(self, host, command, args={}):
         from chroma_core.services.job_scheduler.agent_rpc import AgentException
 
-        result = json.loads(self.invoke_rust_agent(host, command, args))
+        try:
```
jgrund: I think the only thing inside this try should be 

`result = self.invoke_rust_agent(host, command, args)`

Then the json parsing should be in a separate try afterwards.

johnsonw: sounds good 👍

chroma_core/lib/job.py
```diff
@@ -200,7 +200,21 @@ def invoke_rust_agent(self, host, command, args={}):
...
+            result = self.invoke_rust_agent(host, command, args)
+        except RustAgentCancellation as e:
+            raise AgentException(host, command, args, "Cancelled: {}".format(e))
+        except ValueError as e:
```
johnsonw: We definitely need the ValueError below when parsing json but i'm wondering if I need it here as well?

jgrund: Nope.

chroma_core/lib/job.py
```diff
@@ -200,7 +200,21 @@ def invoke_rust_agent(self, host, command, args={}):
...
+            raise AgentException(host, command, args, result)
+        except Exception as e:
+            raise AgentException(host, command, args, "Unexpected error: {}".format(e))
+
+        try:
+            result = json.loads(result)
+        except ValueError as e:
+            raise AgentException(host, command, args, "Error parsing json: {}".format(e))
+        except Exception as e:
+            raise AgentException(host, command, args, "Unexpected error while parsing json: {}".format(e))
```
jgrund: Is there any other possible error besides a ValueError?

chroma_core/lib/job.py
johnsonw: ok great, i'll remove then.

jgrund: To be clear, we shouldn't need the `ValueError` case.

chroma_core/lib/job.py
johnsonw: I think we just nee ValueError. https://github.com/simplejson/simplejson/blob/v3.0.5/simplejson/decoder.py#L33 shows the the encoder exception inherits from ValueError. Should I remove the general exception underneath?

jgrund: Yes... unless there is a possibility Exception can be raised in addition to `ValueError`

johnsonw: I did a quick search and din't see anything.

PR #1030, opened 2019-07-02
Update manager cli to parse purge and report durations correctly

johnsonw:
Currently, the cli converts hours to seconds and days to seconds.
Stratagem uses milliseconds instead of seconds so the conversion needs
to be updated.

- Convert hours to milliseconds
- Convert days to milliseconds

### DIFF
--- a/iml-agent/src/cli.rs
+++ b/iml-agent/src/cli.rs
@@ -53,8 +53,10 @@ fn parse_duration(src: &str) -> Result<u32, io::Error> {
         .map_err(|_| invalid_input_err(&format!("Could not parse {} to u32", val)))?;
 
     match unit {
-        Some('h') => Ok(val * 3_600),
-        Some('d') => Ok(val * 86_400),
+        Some('h') => Ok(val * 3_600_000),
+        Some('d') => Ok(val * 86_400_000),
+        Some('m') => Ok(val * 60_000),
+        Some('s') => Ok(val * 1_000),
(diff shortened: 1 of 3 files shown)


iml-manager-cli/src/main.rs
```diff
@@ -51,8 +51,8 @@ fn parse_duration(src: &str) -> Result<u32, ImlManagerCliError> {
     let val = val.parse::<u32>()?;
 
     match unit {
-        Some('h') => Ok(val * 3_600),
-        Some('d') => Ok(val * 86_400),
+        Some('h') => Ok(val * 3_600_000),
+        Some('d') => Ok(val * 86_400_000),
```
johnsonw: Should I add an option for minutes as well?

jgrund: May as well add seconds too, so it's easier for debugging.

johnsonw: 👍

iml-agent/src/cli.rs
```diff
@@ -53,8 +53,10 @@ fn parse_duration(src: &str) -> Result<u32, io::Error> {
         .map_err(|_| invalid_input_err(&format!("Could not parse {} to u32", val)))?;
```
jgrund: Parse here needs updating to u64 as well.

johnsonw: Thanks, I meant to include this in the same commit and forgot to add it.

PR #1036, opened 2019-07-05
Temporary solution for race between plugin_runner and device-aggregator

AlexTalker:
# Setup
4 servers, aka 2 DC. first pair holds data(active-active), second - metadata(active-passive)
1 IML server.
Network between Lustre client and servers is `IPoIB` via simple switch.
Everything goes in monitored mode.

# Problem

Sometimes, in logs seen message:
```
[2019-07-04 08:44:49,214: ERROR/plugin_runner] iml-device-aggregator is not providing expected data, ensure iml-device-scanner package is installed and relevant services are running on storage servers (No JSON object could be decoded)
```

Usually it happens:
1. On restart/reboot of IML server
2. Randomly

The reason seems to be a race condition between `plugin_runner` and `device-aggregator-daemon`.
First of all, there's no dependency declared between these services in `systemd` unit, and connection goes via `HTTP`.
Second, since connection proxies by nginx, it returns 502 if service didn't started yet.
Third, if connection from agent is initiated, it doesn't mean device information has already arrive in parallel. Thus, it is useful to wait for the incoming data. 

Signed-off-by: Alex Talker <alextalker@ya.ru>


jgrund:
Jenkins: trigger a test run

chroma_core/plugins/linux.py
```diff
@@ -211,12 +211,22 @@ def agent_session_start(self, host_id, data, initial_scan=True):
             initiate_device_poll = False
             reported_device_node_paths = []
 
-            fqdn = ManagedHost.objects.get(id=host_id).fqdn
-            devices = get_devices(fqdn)
+            host = ManagedHost.objects.get(id=host_id)
+            fqdn = host.fqdn
+            devices = get_devices(fqdn, timeout=5.0)
 
             # use info from IML 4.0
```
jgrund:
```suggestion
```

chroma_core/plugins/linux.py
```diff
@@ -211,12 +211,22 @@ def agent_session_start(self, host_id, data, initial_scan=True):
...
-            if not devices and data:
-                devices = data
+            if not devices:
+                if data:
```
jgrund:
```suggestion
                # use info from IML 4.0
                if data:
```

chroma_core/plugins/block_devices.py
```diff
@@ -12,23 +12,51 @@
...
+
+
+def get_devices(fqdn, timeout=0):
+    summary = 0.0
+    timeout = timeout if timeout >= 0 else 0
+
+    while summary <= timeout:
+        try:
+            _data, elapsed = _fetch_aggregator(timeout - summary)
+            summary += elapsed
```
jgrund: What's the point of incrementing summary here?

If we've reached this spot we are exiting the loop.

chroma_core/plugins/block_devices.py
```diff
@@ -12,23 +12,51 @@
...
+            return _data[fqdn]
+        except (KeyError, TypeError) as e:
+            # This error might be caused by fact that device-aggregator
+            # hasn't yet received device information update on startup
+            log.error(
+                "iml-device-aggregator is not providing expected data, ensure "
+                "iml-device-scanner package is installed and relevant "
+                "services are running on storage servers (%s)" % e
```
jgrund: I know you didn't write this error message, but I think the wording here does not reflect what the probable causes are.

Something like:

`iml-device-aggregator is not providing expected data for fqdn {}. Waited: {}, Trying again for: {}, Error: {}.format(fqdn, summary, timeout - summary, e)`

chroma_core/plugins/block_devices.py
```diff
@@ -12,23 +12,51 @@
...
-        return _data[fqdn]
-    except Exception as e:
-        log.error(
-            "iml-device-aggregator is not providing expected data, ensure "
-            "iml-device-scanner package is installed and relevant "
-            "services are running on storage servers (%s)" % e
-        )
-        return {}
+    summary = 0.0
+    timeout = timeout if timeout >= 0 else 0
```
jgrund: If someone passes in a negative number, is it better to change it to 0 than raise an error?

jgrund:
Jenkins: trigger a test run

chroma_core/plugins/block_devices.py
```diff
@@ -12,23 +12,51 @@
...
+    while summary <= timeout:
+        resp = requests.get(settings.DEVICE_AGGREGATOR_URL)
+        summary += resp.elapsed.seconds
+        summary += resp.elapsed.microseconds / 10.0 ** 6
+
+        try:
+            return resp.json()
```
jgrund: This variant is not returning a tuple, which is causing testing to fail.

```suggestion
            return resp.json(), summary
```

chroma_core/plugins/block_devices.py
AlexTalker: If we have reached this, we still might get `KeyError` but we need to keep time knowledge intact

chroma_core/plugins/block_devices.py
AlexTalker: I think so. At least, from a point of backward compatibility.
If you wish, you can suggest a proper exception name.

jgrund:
Jenkins: trigger a test run

jgrund:
Jenkins: trigger a test run

jgrund:
Jenkins: trigger a test run

jgrund:
Jenkins: trigger a test run

jgrund:
Refreshed on latest master

jgrund:
Jenkins: trigger a test run

jgrund:
Jenkins: trigger a test run

PR #1041, opened 2019-07-09
Add lock-indicator component

jgrund:
The lock indicator is used to show any active locks on ongoing jobs.
It is associated with a cooresponding record.


jgrund:
Screen Shot 2019-07-09 at 4 16 20 PM

PR #1044, opened 2019-07-10
Add Grafana Dashboard Provisioning

johnsonw:
Grafana allows for privisioning of dashboards as well as data sources.
This patch will provision grafana with a stratagem dashboard that will
be built and ready to go as soon as installation completes. This will
allow us to easily embed an iframe and create the chart component.
    
### DIFF
--- a/chroma_core/lib/stratagem.py
+++ b/chroma_core/lib/stratagem.py
@@ -2,6 +2,8 @@ import requests
 import settings
 import operator
 import json
+import calendar
+import time
 
 from toolz.functoolz import pipe, partial
 from requests.exceptions import ConnectionError
@@ -23,13 +25,6 @@ labels = {
     "greater_than_equal_1t": ">\\\=\\\ 1\\\ TiB",
 }
 
-distribution_weight = {
-    "less_than_1m": 0,
-    "greater_than_equal_1m_less_than_1g": 1,
-    "greater_than_equal_1g": 2,
-    "greater_than_equal_1t": 3,
-}
-
 filter_out_other_counter = partial(filter, lambda counter: counter.get("name").lower() != "other")
 flatten = lambda xs: [item for l in xs for item in l]
 
@@ -38,13 +33,13 @@ def tuple_to_equals(xs):
     return "{}={}".format(*xs)
 
 
-def create_stratagem_influx_point(measurement, tags, fields, time):
+def create_stratagem_influx_point(measurement, tags, fields):
     return "{},{} {} {}".format(
-        measurement, ",".join(map(tuple_to_equals, tags)), ",".join(map(tuple_to_equals, fields)), time or ""
+        measurement, ",".join(map(tuple_to_equals, tags)), ",".join(map(tuple_to_equals, fields)), ""
     ).rstrip()
 
 
-def parse_size_distribution(measurement, distribution_weight, labels, counters):
+def parse_size_distribution(measurement, labels, counters):
     return pipe(
         counters,
         filter_out_other_counter,
@@ -55,12 +50,10 @@ def parse_size_distribution(measurement, distribution_weight, labels, counters):
                 measurement,
                 [
                     ("group_name", "size_distribution"),
-                    ("distribution_weight", distribution_weight.get(x.get("name"))),
                     ("counter_name", x.get("name")),
                     ("label", labels.get(x.get("name"))),
                 ],
                 [("count", x.get("count"))],
-                None,
             ),
         ),
     )
@@ -83,7 +76,6 @@ def parse_user_distribution(measurement, counters):
                     ("counter_name", x.get("name")),
                 ],
                 [("count", x.get("count"))],
-                None,
             ),
         ),
     )
@@ -91,7 +83,7 @@ def parse_user_distribution(measurement, counters):
 
 def parse_stratagem_results_to_influx(measurement, stratagem_results_json):
     parse_fns = {
-        "size_distribution": partial(parse_size_distribution, measurement, distribution_weight, labels),
+        "size_distribution": partial(parse_size_distribution, measurement, labels),
         "user_distribution": partial(parse_user_distribution, measurement),
     }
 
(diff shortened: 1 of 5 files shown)


johnsonw:
Upon installation I logged into IML and then attempted to load the grafana dashboard. The stratagem dashboard was loaded and ready:

image

jgrund:
> Upon installation I logged into IML and then attempted to load the grafana dashboard. The stratagem dashboard was loaded and ready.

On a clean install?

johnsonw:
For this test I removed IML and grafana and then re-installed. I will destroy the admin node and do a full install to perform a full test.

jgrund:
> I will destroy the admin node and do a full install to perform a full test. 

No need, I was just surprised you would see data on a clean install.

chroma_core/lib/stratagem.py
```diff
@@ -16,6 +16,19 @@
     "size >= 1t": "greater_than_equal_1t",
 }
 
+labels = {
```
jgrund: It's too bad we need to do this.

jgrund:
Is there a way to select the light theme in the conf file?

chroma_core/lib/stratagem.py
```diff
@@ -16,6 +16,19 @@
...
+    "less_than_1m": "<\\\ 1\\\ Mib",
+    "greater_than_equal_1m_less_than_1g": ">\\\=\\\ 1\\\ Mib\\\,\\\ <\\\ 1\\\ GiB",
+    "greater_than_equal_1g": ">\\\=\\\ 1\\\ GiB",
+    "greater_than_equal_1t": ">\\\=\\\ 1\\\ TiB",
+}
+
+distribution_weight = {
```
jgrund:
```suggestion
order = {
```

`distribution_weight` makes it sound like these have different weights when the only purpose for this is to enforce ordering.

Also, since is just a k v dict of increasing order, just use a list instead.

chroma_core/lib/stratagem.py
```diff
@@ -40,7 +53,12 @@ def parse_size_distribution(measurement, counters):
             map,
             lambda x: create_stratagem_influx_point(
                 measurement,
-                [("group_name", "size_distribution"), ("counter_name", x.get("name"))],
+                [
+                    ("group_name", "size_distribution"),
+                    ("distribution_weight", distribution_weight.get(x.get("name"))),
```
jgrund: use order as name here as well.

chroma_core/lib/stratagem.py
```diff
@@ -150,9 +168,11 @@ def submit_aggregated_data(measurement, aggregated):
     points = map(
         lambda point: (
             [
-                ("counter_name", point.get("counter_name")),
                 ("classify_attr_type", point.get("classify_attr_type")),
                 ("group_name", point.get("group_name")),
+                ("distribution_weight", point.get("distribution_weight")),
```
jgrund: order

grafana/dashboards/stratagem-dashboard-1.json
```diff
@@ -0,0 +1,168 @@
+{
+  "annotations": {
```
jgrund: What is the purpose of annotations?

grafana/dashboards/stratagem-dashboard-1.json
```diff
@@ -0,0 +1,168 @@
...
+        {
+          "alias": "$tag_label",
+          "groupBy": [
+            {
+              "params": ["label"],
+              "type": "tag"
+            }
+          ],
+          "measurement": "stratagem_scan",
+          "orderByTime": "ASC",
```
jgrund: Is this needed? there is only a single point available

grafana/dashboards/stratagem-dashboard-1.json
```diff
@@ -0,0 +1,168 @@
...
+        }
+      ],
+      "yaxis": {
+        "align": false,
+        "alignLevel": null
+      }
+    }
+  ],
+  "schemaVersion": 18,
+  "style": "dark",
```
jgrund:
```suggestion
  "style": "light",
```

guessing light provides the white background.

grafana/dashboards/stratagem-dashboard-1.json
```diff
@@ -0,0 +1,168 @@
...
+  "tags": ["stratagem"],
+  "templating": {
+    "list": []
+  },
+  "time": {
+    "from": "now-6h",
```
jgrund: Time is irrelevant in this chart. We want an infinite time scale so we don't miss older scans.

-6h would miss most data.

johnsonw:
I'll look and see. Also, transparency is an option. Do you think it should be transparent?

chroma_core/lib/stratagem.py
johnsonw: I'm monitoring https://community.grafana.com/t/how-can-i-map-x-axis-values-to-user-friendly-labels-in-a-series-graph-using-influx/19140. No one has responded yet unfortunately.

grafana/dashboards/stratagem-dashboard-1.json
johnsonw: This was generated by grafana. From their documentation: 
> Annotations provide a way to mark points on the graph with rich events. When you hover over > an annotation you can get event description and event tags. The text field can include links to > other systems with more detail.

grafana/dashboards/stratagem-dashboard-1.json
johnsonw: This was generated by grafana.

grafana/dashboards/stratagem-dashboard-1.json
johnsonw: 👍

grafana/dashboards/stratagem-dashboard-1.json
johnsonw: Yes, I've tried removing this by editing the query manually but it always adds it back. I think there is another place to modify this at the dashboard level so I will try there.

johnsonw:
After reprovisioning the admin node I did a clean install and grafana loads the chart as expected, without data:
image

grafana/dashboards/stratagem-dashboard-1.json
johnsonw: There doesn't seem to be a way to remove this so I set the starting date to the epoch (1970-01-01T00:00:00.000Z).

chroma_core/lib/stratagem.py
```diff
@@ -2,6 +2,8 @@
 import settings
 import operator
 import json
+import calendar
+import time
```
jgrund: Are these imports still needed?

PR #1064, opened 2019-07-11
Fixup RPM build

### LINKED ISSUE #1063: Copr build broken after grafana dashboard landing
The Copr build for IML devel is not completing. It appears this is because the grafana dir was provided as an unpacked source. We have 19 sources in our spec, and we could have far less. Fix the unpacked source issue and also do the following:

- Remove .service files from the manifest.in, those should not be present for non RPM installs.
- Create a source1 that is a compressed tarball of sources 1-19.

jgrund:
Fixes #1063.


python-iml-manager.spec
```diff
@@ -226,27 +210,27 @@ mkdir -p $RPM_BUILD_ROOT%{_sysconfdir}/rabbitmq
...
+install -m 644 iml-gunicorn.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-http-agent.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-job-scheduler.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-lustre-audit.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-manager.target $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-plugin-runner.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-power-control.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-settings-populator.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-stats.service $RPM_BUILD_ROOT%{_unitdir}/
+install -m 644 iml-syslog.service $RPM_BUILD_ROOT%{_unitdir}/
```
utopiabound: Can we install `*.service` instead of enumerating them?

jgrund: Done

jgrund: Just kidding. 

These files need to enumerated here (or in the Makefile task) so we don't pull in the rust .service files as well.

I did remove the enumeration from the `%files` block though.

python-iml-manager.spec
```diff
@@ -15,30 +15,13 @@ BuildRequires: systemd
 Name:           python-%{pypi_name}
 Version:        %{version}
 # Release Start
-Release:    1%{?dist}
+Release:    1.%{?dist}
```
jgrund:
```suggestion
Release:    1%{?dist}
```

jgrund:
> I feel like tarring it all up as an "extra" tarball feels kludgy, and we should be putting it in the source tarball, but it's better than it was I guess.

The reason for the extra tarball is we don't need things like .service files in docker containers, ideally we should be able to put the python package directly into the container without the extra cruft (admittedly we're not there yet).

If you have a solution that allows for this approach, I'd be willing to implement it.

PR #1065, opened 2019-07-11
GH-1056 - Create nginx integration with influx

### LINKED ISSUE #1056: Create nginx integration with influx
Create a simple integration with influxdb + nginx so it can be utilized via the GUI.

We can use the existing nginx auth check as a gate for this endpoint.

johnsonw:
Fixes #1056.

Description:
Since influxdb provides an http api, the reverse proxy feature of nginx
should be leveraged to allow a query to be run from the application.
This will allow components to be given a stream coming directly from the
database rather than having to go through realtime to get the data. The
end point should be restricted to GET only and should follow
authentication rules. Queries can be made in the form:

https://localhost:8443/influx/<dbname>/<query>

### DIFF
--- a/chroma-manager.conf.template
+++ b/chroma-manager.conf.template
@@ -152,7 +152,17 @@ server {
         proxy_set_header X-WEBAUTH-USER "admin";
         rewrite  ^/grafana/(.*)  /$1 break;
         proxy_pass http://localhost:3000/;
-   }
+    }
+
+    location /influx {
+        limit_except GET { deny all; }
+        auth_request /auth;
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Server $host;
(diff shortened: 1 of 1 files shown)


chroma-manager.conf.template
```diff
@@ -152,7 +152,17 @@ server {
...
-   }
+    }
+
+    location /influx {
+        limit_except GET { deny all; }
+        auth_request /auth;
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Forwarded-Server $host;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        rewrite ^/influx/(\w+)/(.*)$ /query?db=$1&q=$2 break;
```
jgrund: Is this necessary?

PR #1066, opened 2019-07-11
GH-1062 - Stratagem duratiion models should be bigint

### LINKED ISSUE #1062: Stratagem duration models should be bigint
Any stratagem model field that needs a ms accuracy should be represented by a bigint in the database.

johnsonw:
Fixes #1062.

Since milleseconds will be used for duration and it's possible that a
number of years could be specified, the fields should be of the
BigInteger type instead of Integer. This patch will ensure that:
1. The duration fields on the models use BigInteger
2. The API will handle hydrating and dehydrating data from String to
BigInteger such that data is stored as a u64 in the database and
retrieved as a long instead of a string (Tastypie does not support
    BigInteger in resources so it must be converted to a long)
3. Validation is added to ensure that no value > MAX_SAFE_INTEGER (9007199254740991)
  can be used for either duration.

### DIFF
--- a/chroma_api/stratagem.py
+++ b/chroma_api/stratagem.py
@@ -31,6 +31,29 @@ MAX_SAFE_INTEGER = 9007199254740991
 
 class RunStratagemValidation(Validation):
     def is_valid(self, bundle, request=None):
+        def check_duration(duration_key, bundle):
+            duration_type = duration_key.split("_")[0].capitalize()
+            try:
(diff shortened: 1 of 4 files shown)


johnsonw:
```
[root@adm vagrant]#  iml stratagem scan --filesystem fs -r9007199254740991s
✗ Report duration cannot be larger than 9007199254740991.
```

PR #1068, opened 2019-07-12
GH-1062 - Stratagem duration models should be bigint

### LINKED ISSUE #1062: Stratagem duration models should be bigint

johnsonw:
Fixes #1062.

GH-1062 - Stratagem duratiion models should be bigint
    
    Since milleseconds will be used for duration and it's possible that a
    number of years could be specified, the fields should be of the
    BigInteger type instead of Integer. This patch will ensure that:
    1. The duration fields on the models use BigInteger
    2. The API will handle hydrating and dehydrating data from String to
    BigInteger such that data is stored as a u64 in the database and
    retrieved as a long instead of a string (Tastypie does not support
        BigInteger in resources so it must be converted to a long)
    3. Validation is added to ensure that no value > MAX_SAFE_INTEGER (9007199254740991)
      can be used for either duration.
    
### DIFF
--- a/chroma_api/stratagem.py
+++ b/chroma_api/stratagem.py
@@ -31,6 +31,29 @@ MAX_SAFE_INTEGER = 9007199254740991
 
 class RunStratagemValidation(Validation):
     def is_valid(self, bundle, request=None):
+        def check_duration(duration_key, bundle):
+            duration_type = duration_key.split("_")[0].capitalize()
+            try:
+                duration = bundle.data.get(duration_key) and get_bundle_int_val(bundle.data.get(duration_key))
+                if duration > MAX_SAFE_INTEGER:
+                    return {
+                        "code": "{}_too_big".format(duration_key),
+                        "message": "{} duration cannot be larger than {}.".format(duration_type, MAX_SAFE_INTEGER),
+                    }
+            except ValueError:
+                return {
+                    "code": "invalid_argument",
+                    "message": "{} duration must be an integer value.".format(duration_type),
+                }
+
+        purge_check = check_duration("purge_duration", bundle)
+        if purge_check:
+            return purge_check
+
+        report_check = check_duration("report_duration", bundle)
+        if report_check:
+            return report_check
+
         try:
             purge_duration = bundle.data.get("purge_duration") and get_bundle_int_val(bundle.data.get("purge_duration"))
             if purge_duration > MAX_SAFE_INTEGER:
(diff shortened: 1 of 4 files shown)


chroma_api/stratagem.py
```diff
@@ -27,71 +27,158 @@
 
 get_bundle_int_val = compose(partial(flip, int, 10), str)
 
+MAX_SAFE_INTEGER = 9007199254740991
```
jgrund: probably worth adding a comment about why this is needed.

chroma_api/stratagem.py
```diff
@@ -27,71 +27,158 @@
...
+                partial(validate_filesystem, bundle),
+                partial(validate_target_mount, bundle),
+                partial(validate_mdt_profile, bundle),
+                partial(validate_client_profile, bundle),
+            )
+            or {}
+        )
 
 
 class StratagemConfigurationValidation(RunStratagemValidation):
```
jgrund: How is this error being accounted for on the Rust side:

`return {"__all__": "Required fields are missing: {}".format(", ".join(difference))}`

chroma_api/stratagem.py
```diff
@@ -27,71 +27,158 @@
...
+        return {
+            "code": "stratagem_client_profile_not_installed",
+            "message": "A client must be added with the 'Stratagem Client' profile to run this command.",
+        }
+
+
+class RunStratagemValidation(Validation):
+    def is_valid(self, bundle, request=None):
+        return (
+            pipe(
```
jgrund: It's non-obvious what's happening here.

Instead of doing a pipe chain that takes a partial and forwards each result to the next, just use an or statement:

```python
return validate_duration(bundle) or validate_filesystem(bundle) or validate_target_mount(bundle) or validate_mdt_profile(bundle) or validate_client_profile(bundle) or {}
```

By doing this, you can remove the repeated early return of:

```python
	    if result:
        return result
```

from each validation. I understand the appeal of wanting to use functional programming concepts, but for python (which favors simpler code and lacks useful static typing) it tends to obfuscate the intent of the code, especially for outside contributors who may not be used to this style.

chroma_api/stratagem.py
johnsonw: Good catch. Updated to:
```
{
       "code": "required_fields_missing",
       "message": "Required fields are missing: {}".format(", ".join(difference)),
}
```

PR #1073, opened 2019-07-17
Skip All MDTs not MDT0 for filesystem "unavailable" state

### LINKED ISSUE #1023: IML FS moves to unavailable state if one MDT is unmounted
If Lustre is setup to have multiple MDTs, it seems that the filesystem should not move to the unavailable state unless MDT0 has been unmounted.

utopiabound:
Fixes #1023

### DIFF
--- a/chroma_core/lib/util.py
+++ b/chroma_core/lib/util.py
@@ -50,6 +50,17 @@ def sizeof_fmt_detailed(num):
     return int(num)
 
 
+def target_label_split(label):
+    """
+    Split a target label into a tuple of it's parts: (fsname, target type, index)
+    """
+    a = label.split("-")
+    if len(a) == 1:
+        # MGS
+        return (None, a[0][0:3], None)
+    return (a[0], a[1][0:3], int(a[1][3:], 16))
+
+
 class timeit(object):
     def __init__(self, logger):
         self.logger = logger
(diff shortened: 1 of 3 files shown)


chroma_core/models/filesystem.py
```diff
@@ -299,6 +302,9 @@ def description(self):
     def get_deps(self):
         deps = []
         for t in ObjectCache.get_targets_by_filesystem(self.filesystem_id):
+            # Report filesystem available if MDTs other than 0 are unmounted
+            if t.target_type().lower() == "mdt" and not t.get_label().ends_with("MDT0000"):
```
johnsonw: This logic looks to be the same, would it be worth abstracting it? Also, i'm not sure the `continue` is necessary. Maybe something like:
```
def target_is_mdt0(t):
    return t.target_type().lower() == "mdt" and t.get_label().ends_with("MDT0000")
```
Then:
```
if target_is_mdt0(t):
    deps.append(DependOn(t, "mounted", fix_state="unavailable"))
```

johnsonw: by the way, i'm ok with it the way it is, just thought this might be a bit cleaner than using a continue.

chroma_core/lib/util.py
```diff
@@ -50,6 +50,17 @@ def sizeof_fmt_detailed(num):
     return int(num)
 
 
+def target_label_split(label):
+    """
+    Split a target label into a tuple of it's parts: (fsname, target type, index)
```
johnsonw: Should `target type` be `label`? I asked based on the line below:
```
(_, label, index) = target_label_split(t.get_label())
```

utopiabound: I guess the `label` variable should be something like `target_type`

PR #1085, opened 2019-07-24
Add (Un)Configure stratagem timer

### LINKED ISSUE #1048: Implement systemd timer creation for stratagem scans
Implement a local command that will configure stratagem via a new IML API endpoint.

In Scope:

  - Create an endpoint that when called will create / update a systemd.timer with given params.
  - Add a dependency such that when all stratagem servers are removed / the timer is removed as well.

utopiabound:
Create/destroy systemd service and timer.

Fixes #1048


utopiabound:
This now creates and destroys timer/service correctly in manual testing

Issue #926, opened 2019-05-08
Client side - Use rmfid in liblustreapi


utopiabound:
### Description 

Use rmfid to do mass delete without doing fid2path

This is blocked on [LU-12090](https://jira.whamcloud.com/browse/LU-12090) Change [#34449](https://review.whamcloud.com/34449)

PR #1088, opened 2019-07-26
Optionally use llapi_rmfids if it's available

utopiabound:
Issue #926

### DIFF
--- a/liblustreapi/src/lib.rs
+++ b/liblustreapi/src/lib.rs
@@ -313,7 +313,6 @@ impl Llapi {
         Ok(())
     }
 
-    // Does llapi_rmfid() or returns back fidlist as a Vec<String> in Err()
     fn llapi_rmfid(
         &self,
         mntpt: &str,
--- a/liblustreapi/src/lib.rs
+++ b/liblustreapi/src/lib.rs
@@ -313,11 +313,7 @@ impl Llapi {
         Ok(())
     }
 
-    fn llapi_rmfid(
-        &self,
-        mntpt: &str,
-        fidlist: &Vec<String>,
-    ) -> Result<(), ()> {
+    fn llapi_rmfid(&self, mntpt: &str, fidlist: &Vec<String>) -> Result<(), ()> {
         let func: lib::Symbol<extern "C" fn(*const libc::c_char, *const RmfidsArg) -> libc::c_int> =
             match unsafe { self.lib.get(b"llapi_rmfid\0") } {
                 Err(e) => {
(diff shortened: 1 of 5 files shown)


utopiabound:
Manual testing shows this working as expected.
- Calls `llapi_rmfid` if present
- Falls back to `unlink` if it isn't or if llapi_rmfid fails for some other reason (for instance error 524 LUSTRE_ENOTSUPP if rmfid not supported by server)

liblustreapi/src/lib.rs
```diff
@@ -265,29 +313,84 @@ impl Llapi {
...
+
+        let devptr = match CString::new(mntpt) {
+            Ok(c) => c.into_raw(),
+            Err(_) => return Err(flist),
+        };
+
+        let mut fids: Vec<RmfidsArg> = Vec::with_capacity(flist.len() + 1);
+        fids.push(FidArrayHdr::new(0).into());
+        let flist2 = flist.clone();
+        for fidstr in flist {
```
johnsonw: One thing to keep in mind is that for loops can also be written as fold's: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold

johnsonw:
Just a small comment.

liblustreapi/src/lib.rs
```diff
@@ -70,6 +71,53 @@ impl From<types::lu_fid> for Fid {
...
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub union RmfidsArg {
+    pub hdr: FidArrayHdr,
+    pub fid: Fid,
+}
+
+impl From<FidArrayHdr> for RmfidsArg {
+    fn from(hdr: FidArrayHdr) -> Self {
+        Self { hdr: hdr }
```
johnsonw:
```
Self {
   hdr
}
```

liblustreapi/src/lib.rs
```diff
@@ -70,6 +71,53 @@ impl From<types::lu_fid> for Fid {
...
+    }
+}
+
+impl From<Fid> for RmfidsArg {
+    fn from(fid: Fid) -> Self {
+        Self { fid: fid }
```
johnsonw:
```
Self {
   fid
}
```

liblustreapi/src/lib.rs
```diff
@@ -265,29 +313,83 @@ impl Llapi {
...
+        if rc == 0 {
+            Ok(())
+        } else {
+            log::info!(
+                "Call to llapi_rmfids({}, [{}, ...]) failed: {}",
+                &mntpt,
+                unsafe { fids[0].hdr.nr },
+                rc
+            );
+            Err(())
```
johnsonw: Should we return information in the Err object?

utopiabound: No, because we ignore the error and use unlink instead.

PR #1104, opened 2019-07-29
Restart grafana during chroma-config setup

### LINKED ISSUE #1095: grafana.service should restart on `chroma-config setup`
The `grafana.service` should be restarted within `chroma-config setup` so any provisioning changes made can take effect within grafana.

utopiabound:
Fixes #1095


chroma_core/lib/service_config.py
```diff
@@ -293,7 +293,7 @@ def _setup_rabbitmq_credentials(self):
         self.try_shell(sudo + ["rabbitmqctl", "set_user_tags", settings.AMQP_BROKER_USER, "management"])
 
     def _setup_influxdb(self):
-        influx_service = ServiceControlEL7("influxdb")
+        influx_service = ServiceControl("influxdb")
```
johnsonw: Should we also restart influx when restarting chroma? I was looking at postgres and I see the following:

```
 def _restart_pgsql(self):
        postgresql_service = ServiceControl.create("postgresql")
        if postgresql_service.running:
            postgresql_service.reload()
        else:
            postgresql_service.start()
        postgresql_service.enable()
```

utopiabound: Influx does restart during setup

johnsonw: ok guess I missed that.

jgrund:
@utopiabound Looks like there is a different bug now.

PR #1108, opened 2019-07-30
Statagem timer re-add removes old timer

### LINKED ISSUE #1099: Updating stratagem config should reconfigure the systemd timer
When the Stratagem config is updated, the associated systemd timer and service should also be updated with a new interval and warn / purge durations if needed. Right now, the data is only being persisted to the db.

Create an `UpdateStratagemJob` that will update the systemd timer / service with the new values and reload it. The Command should be triggered on a `POST` to the stratagem endpoint.

In addition, add Rust cli support on the manager for updating.

utopiabound:
If you re-add a timer, the old timer should be removed.  This allows "updating" a timer via running add a second time.

Fixes #1099

### DIFF
--- a/chroma_core/models/stratagem.py
+++ b/chroma_core/models/stratagem.py
@@ -67,6 +67,18 @@ class StratagemConfiguration(StatefulObject):
         ordering = ["id"]
 
 
+def unit_name(fid):
+    return "iml-stratagem-{}".format(fid)
+
+
+def timer_file(fid):
+    return "/etc/systemd/system/{}.timer".format(unit_name(fid))
+
+
+def service_file(fid):
+    return "/etc/systemd/system/{}.service".format(unit_name(fid))
+
+
 class ConfigureStratagemTimerStep(Step, CommandLine):
     def run(self, kwargs):
         job_log.debug("Configure stratagem timer step kwargs: {}".format(kwargs))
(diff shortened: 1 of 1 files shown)


utopiabound:
> This approach is basically duplicating the logic in `UnconfigureStratagemTimerStep`.
> 
It's actually lighter weight than a full `UnconfigureStratagemTimerStep` because it doesn't do a daemon-reload

> Instead, why don't you conditionally add `UnconfigureStratagemTimerStep` if the timer already exists here:
> 
> https://github.com/whamcloud/integrated-manager-for-lustre/blob/d13e4df2260cb4ae45631760ba4d81568b02ee58/chroma_core/models/stratagem.py#L153-L156
> 
> That will be more explicit what is happening in the cmd output and will reuse the existing logic.

It seems like an improper separation of layers to check the file existance in the Job, vs in the Step

jgrund:
> It's actually lighter weight than a full UnconfigureStratagemTimerStep because it doesn't do a daemon-reload

I don't think there's much cost to running `systemctl daemon-reload`, is there?

> It seems like an improper separation of layers to check the file existence in the Job, vs in the Step

It's certainly a strange way to do it, but it's how It's done within IML. The advantage of doing it this way is it's more explicit what steps were taken when reviewing later on from a user perspective. 

In other words, but using an explicit step the user will see that there was an unconfigure step taken in the GUI before configuring again.

utopiabound:
window
This is what shows in the command status output for subsequent runs of configure add.

PR #1112, opened 2019-08-01
Fixup Stratagem bugs and enhancements on the backend and wasm components

### LINKED ISSUE #1075: Add foreign key for filesystem in Stratagem Configuration model
Currently the stratagem configuration contains a reference to the filesystem id it is configured with. It does not, however, enforce referential integrity by using a foreign key. After discussing this further we cannot think of a case in which a stratagem configuration should remain after a filesystem is removed. Therefore, we should set the foreign key constraint on the filesystem_id column in the stratagem configuration table.

### LINKED ISSUE #1094: When forgetting the filesystem in Monitored mode, also forget Stratagem
When the user clicks the forget action on a monitored fs:

Is Stratagem is configured for the fs, it should have it's `UnconfigureStratagemJob` run as part of the fs removal. 

An open question is what should happen if someone attempts to remove the fs while a scan is occuring, but this can probably be handled in another issue.

### LINKED ISSUE #1096: Top inode users table should be truncated after n users
The top inode users table should be truncated after n users to deal with the case where there may be many users writing data to a client.

### LINKED ISSUE #1100: Enable / Update / Delete stratagem config should show command modal
Clicking the Enable / Update or Delete buttons within Stratagem config should show the command modal. This way the user can be notified of progress.

In addition, these commands should take a write-lock on the stratagem object, which we can use to disable the button.

### LINKED ISSUE #1105: Create Separate Scan Now Modal
Create a new modal that can run a stratagem scan if all preconditions are met. Ensure it is separate from the timer configuration.

### LINKED ISSUE #1106: Kick off the inode data fetch when the stratagem configuration is received.
The inode table currently should kick off the fetch when the stratagem configuration is received.

### LINKED ISSUE #1107: Fix table formatting on fs detail page
The length of the volume name in each table is different, which shifts the entries in the table a bit. We should fix the widths in the table cells to ensure the tables columns are aligned properly.

johnsonw:
Fixes #1100.
Fixes #1107.
Fixes #1106.
Fixes #1075.
Fixes #1096.
Fixes #1105.
Fixes #1094.


chroma_core/models/stratagem.py
```diff
@@ -104,10 +131,15 @@ def run(self, kwargs):
...
-                "ExecStart={}\n".format(config.filesystem_id, iml_cmd)
+                "ExecStart={}\n".format(config.filesystem.id, iml_cmd)
             )
         self.try_shell(["systemctl", "daemon-reload"])
         self.try_shell(["systemctl", "enable", "--now", "{}.timer".format(name)])
+        
+        # This step happens so quickly that sometimes the modal doesn't have an opportunity to appear. 
+        # Adding a short sleep will ensure that the command modal can be launched. This is temporary 
+        # for the demo and until we can find a better solution.
+        time.sleep(1)
```
jgrund: This is a front-end problem. Why is there a possibility that the modal can be missed?

jgrund:
Closing as this patch has been obsoleted by breaking it into smaller patches.

PR #1118, opened 2019-08-05
Start timer after interval from unit activation

### LINKED ISSUE #1117: iml-stratagem timer should not use OnBootSec to calculate next run
The current timer uses `OnBootSec` and `OnUnitActiveSec` as determinations of when to run the scan next. 

`OnBootSec` appears to take precedence and schedules the next run relative to how long the system has been up.

Ex with a timer of 30 days:

Instead of having `OnBootSec` I think we can just use `OnUnitActiveSec` (haven't tested if that works correctly on reboot)

utopiabound:
Only start timer after unit start, not based on IML boot time.

Fixes #1117

### DIFF
--- a/chroma_core/models/stratagem.py
+++ b/chroma_core/models/stratagem.py
@@ -86,7 +86,7 @@ class ConfigureStratagemTimerStep(Step, CommandLine):
                 "\n[Unit]\n"
                 "Description=Start Stratagem run on {}\n"
                 "\n[Timer]\n"
-                "OnBootSec={}\n"
+                "OnActiveSec={}\n"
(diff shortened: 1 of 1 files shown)


chroma_core/models/stratagem.py
```diff
@@ -86,7 +86,7 @@ def run(self, kwargs):
                 "\n[Unit]\n"
                 "Description=Start Stratagem run on {}\n"
                 "\n[Timer]\n"
-                "OnBootSec={}\n"
+                "OnActiveSec={}\n"
```
jgrund: Why do we need both of these?

PR #1129, opened 2019-08-08
Stratagem API checks

### LINKED ISSUE #1128: Stratagem API endpoint should validate duration lower-bound
When sending a config to stratagem, we should validate that all incoming numbers are non-negative (in addition to the other current checks) and return an error describing the issue if they are `< 0`.

utopiabound:
Check durations are non-negative

Fixes #1128

### DIFF
--- a/chroma_api/stratagem.py
+++ b/chroma_api/stratagem.py
@@ -63,6 +63,12 @@ def check_duration(duration_key, bundle):
             ),
         }
 
+    if duration < 0:
+        return {
+            "code": "{}_too_small".format(duration_key),
+            "message": "{} duration cannot be negative.".format(get_duration_type(duration_key)),
+        }
+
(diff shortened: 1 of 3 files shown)


chroma_api/stratagem.py
```diff
@@ -63,6 +63,12 @@ def check_duration(duration_key, bundle):
             ),
         }
 
+    if duration < 0:
+        return {
+            "code": "{}_too_small".format(duration_key),
```
jgrund: The way this is currently designed means you also need to add new variants to this enum:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/465e8fb076d3ac083d6f1dc4c4ec3a1c5d1cd402/iml-manager-cli/src/manager_cli_error.rs#L12-L29

PR #1132, opened 2019-08-08
Fix table formatting

### LINKED ISSUE #1107: Fix table formatting on fs detail page

johnsonw:
Fixes #1107.

### DIFF
--- a/iml-wasm-components/iml-fs/src/fs_detail_page.rs
+++ b/iml-wasm-components/iml-fs/src/fs_detail_page.rs
@@ -419,20 +419,22 @@ fn target_table(
             thead![tr![
                 th!["Name"],
                 th!["Volume"],
-                th!["Primary Server"],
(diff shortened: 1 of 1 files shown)


johnsonw:
image

image

jgrund:
Is the Volume label for the MGS being truncated?

johnsonw:
I intentionally removed a good portion of the string to show that the tables are still in alignment.

johnsonw:
The second image shows that the primary and secondary columns are not being displayed.

jgrund:
According to mdn, break-word is deprecated: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

johnsonw:
Looks like I can use "break-all"

PR #1145, opened 2019-08-13
Stratagem scan attempts to send fid and purge lists to client, even if

### LINKED ISSUE #1137: Stratagem scan attempts to send fid and purge lists to client, even if they don't exist
If stratagem has been configured to perform scans and purges, it will always try to send both a warn list and a purge list to the client after a scan.

However, if there were no fids found for a particular rule, the fid list will not be sent to the manager via mailbox at all.

Instead of always trying to send both, the job should check that the files in question exist on the fs before sending. If they do not exists, it's a noop.

johnsonw:
they don't exist

Fixes #1137.

- Make sure values are not None before checking if they are negative in
validation
- Filter out action items if the mailbox file does not exist

### DIFF
--- a/chroma_api/stratagem.py
+++ b/chroma_api/stratagem.py
@@ -53,7 +53,7 @@ def get_duration(duration_key, bundle):
 def check_duration(duration_key, bundle):
     duration = get_duration(duration_key, bundle)
 
-    if isinstance(duration, dict):
+    if duration is None or isinstance(duration, dict):
         return duration
 
     if duration > MAX_SAFE_INTEGER:
@@ -64,7 +64,7 @@ def check_duration(duration_key, bundle):
             ),
         }
 
-    if duration is not None and duration < 0:
+    if duration < 0:
         return {
             "code": "{}_too_small".format(duration_key),
             "message": "{} duration cannot be negative.".format(get_duration_type(duration_key)),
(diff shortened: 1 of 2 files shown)


chroma_api/stratagem.py
```diff
@@ -64,7 +64,7 @@ def check_duration(duration_key, bundle):
             ),
         }
 
-    if duration < 0:
+    if duration is not None and duration < 0:
```
jgrund: Shouldn't this just return early at the top of this fn if duration is None?

chroma_core/models/stratagem.py
```diff
@@ -529,6 +531,8 @@ def run(self, args):
             if duration is not None
         ]
 
+        action_list = filter(lambda (label, (mnt_pt, mailbox_file)): path.exists("/var/spool/iml/mailbox/{}".format(mailbox_file)), action_list)
```
jgrund: Utilize `MAILBOX_PATH` in settings. It has this path hardcoded.

chroma_api/stratagem.py
johnsonw: sure, i'll move the condition up.

chroma_core/models/stratagem.py
johnsonw: ah, thanks

PR #1150, opened 2019-08-15
Add support for returning sizes of file counts.

### LINKED ISSUE #1139: Integrate file size usage into file-size-distribution chart
There is an ask to display total space used per buckets.

This should be fairly straightforward to implement.

1. Enable the new field `summarize_size:true` to the top level of the config.
1. Utilize the `size` field returned in each counter result. It will need to be persisted into the influx db alongside the existing measurement.
1. Display the size in the most appropriate value. for the File Size Distribution chart it should be displayed on top of each bar.
1. For the Inode users table, there should be a new space-used column that reports the size in a human readable format.

jgrund:
Fixes #1139.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,7 +72,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "arc-swap"
-version = "0.4.1"
+version = "0.4.2"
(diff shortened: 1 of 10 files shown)


jgrund:
stratagem-v1-gui

PR #1158, opened 2019-08-20
Create influx measurements and dynamic grafana dashboards for each fi…

### LINKED ISSUE #1111: Create influx measurements and dynamic grafana dashboards for each filesystem
Each filesystem should have its own measurement in the influx db. When stratagem scans, the results should be written to the appropriate measurement within the db.

johnsonw:
…lesystem

Fixes #1111.

Each filesystem should have its own measurement in the influx db. When stratagem scans, the results should be written to the appropriate measurement within the db.

### DIFF
--- a/iml-wasm-components/iml-grafana-chart/src/lib.rs
+++ b/iml-wasm-components/iml-grafana-chart/src/lib.rs
@@ -12,13 +12,14 @@ pub static GRAFANA_DASHBOARD_NAME: &str = "stratagem";
 pub fn grafana_chart<T>(
     dashboard_id: &str,
     dashboard_name: &str,
+    fs_name: &str,
     refresh: &str,
     panel_id: u16,
     width: &str,
     height: &str,
 ) -> Node<T> {
     iframe![attrs! {
-       At::Src => format!("{}d-solo/{}/{}?orgId=1&refresh={}&panelId={}", grafana_root(), dashboard_id, dashboard_name, refresh, panel_id),
+       At::Src => format!("{}d-solo/{}/{}?orgId=1&var-fs_name={}&refresh={}&panelId={}", grafana_root(), dashboard_id, dashboard_name, fs_name, refresh, panel_id),
        At::Width => width,
        At::Height => height,
        "frameborder" => 0
--- a/iml-wasm-components/iml-grafana-chart/src/lib.rs
+++ b/iml-wasm-components/iml-grafana-chart/src/lib.rs
@@ -3,25 +3,38 @@
 // license that can be found in the LICENSE file.
 
 use iml_environment::grafana_root;
-use seed::{attrs, iframe, prelude::*};
+use seed::{attrs, iframe, p, prelude::*};
 
 pub static GRAFANA_DASHBOARD_ID: &str = "OBdCS5IWz";
 pub static GRAFANA_DASHBOARD_NAME: &str = "stratagem";
+
+#[derive(Debug, Default, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GrafanaChartData<'a> {
+    pub org_id: u16,
+    #[serde(rename(serialize = "var-fs_name"))]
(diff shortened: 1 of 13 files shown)


iml-wasm-components/iml-grafana-chart/src/lib.rs
```diff
@@ -12,13 +12,14 @@ pub static GRAFANA_DASHBOARD_NAME: &str = "stratagem";
...
     dashboard_name: &str,
+    fs_name: &str,
     refresh: &str,
     panel_id: u16,
     width: &str,
     height: &str,
 ) -> Node<T> {
     iframe![attrs! {
-       At::Src => format!("{}d-solo/{}/{}?orgId=1&refresh={}&panelId={}", grafana_root(), dashboard_id, dashboard_name, refresh, panel_id),
+       At::Src => format!("{}d-solo/{}/{}?orgId=1&var-fs_name={}&refresh={}&panelId={}", grafana_root(), dashboard_id, dashboard_name, fs_name, refresh, panel_id),
```
jgrund: After this change, the `grafana_chart fn` will have a var specific to filesystems. Is there a way we can pass extra vars?

What if instead of passing the qs params individually you pass in something that implements `Serialize` and use https://docs.rs/serde_urlencoded/0.6.1/serde_urlencoded/fn.to_string.html

iml-wasm-components/iml-stratagem/src/lib.rs
```diff
@@ -68,6 +69,7 @@ impl Default for Model {
         Model {
```
jgrund: Just curious, why do we need a default implementation for this?

johnsonw: I was actually thinking about that as well. Maybe I should remove this altogether.

iml-wasm-components/iml-grafana-chart/src/lib.rs
johnsonw: I like the idea. I'll update.

johnsonw:
Filesystem 1:
image

Filesystem 2:
image

iml-wasm-components/iml-grafana-chart/src/lib.rs
```diff
@@ -3,25 +3,38 @@
...
 /// Create an iframe that loads the specified stratagem chart
 pub fn grafana_chart<T>(
     dashboard_id: &str,
     dashboard_name: &str,
-    fs_name: &str,
-    refresh: &str,
-    panel_id: u16,
     width: &str,
     height: &str,
+    chart_data: GrafanaChartData,
```
jgrund: This still has the same problem from earlier. This should just be a param that implements Serialize, then you won't need to care about the param values at all.

iml-wasm-components/iml-stratagem/src/lib.rs
```diff
@@ -51,7 +53,7 @@ pub struct ActionResponse {
     command: iml_wire_types::Command,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
```
jgrund: This is now derived, but why do we need it at all?

iml-wasm-components/iml-grafana-chart/src/lib.rs
johnsonw: Perhaps I misunderstood? I put all of the query parameters into the GrafanaChartData struct. The other items being passed in are not query parameters. Are those supposed to be in the struct as well?

iml-wasm-components/iml-stratagem/src/lib.rs
johnsonw: Will update this in a separate patch.

iml-wasm-components/iml-grafana-chart/src/lib.rs
johnsonw: Changed to `impl serde::Serialize` as discussed.

Issue #1157, opened 2019-08-20
Show legend for file-size and space distribution charts


jgrund:
There may be cases where the variance between buckets is so large that the bars are basically invisible. By adding a legend we can see the values in these cases.

<img width="937" alt="Screen Shot 2019-08-20 at 9 43 07 AM" src="https://user-images.githubusercontent.com/458717/63352565-5779f380-c32f-11e9-878b-41c3e0dab025.png">

PR #1164, opened 2019-08-21
GH-1157 - Show legend for file-size and space distribution charts

johnsonw:
There may be cases where the variance between buckets is so large that the bars are basically invisible. By adding a legend we can see the values in these cases.

### DIFF
--- a/grafana/dashboards/stratagem-dashboard-1.json
+++ b/grafana/dashboards/stratagem-dashboard-1.json
@@ -20,7 +20,7 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "iteration": 1566309516649,
+  "iteration": 1566404748526,
   "links": [],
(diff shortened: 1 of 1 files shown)


johnsonw:
Once the other patch lands we can rebase on master and it will only contain the legend portion.

grafana/dashboards/stratagem-dashboard-1.json
```diff
@@ -42,11 +42,14 @@
         "alignAsTable": false,
         "avg": false,
         "current": false,
+        "hideEmpty": false,
```
jgrund: What do these params do? Any reason they are not on the other chart?

johnsonw: It's probably from setting them to look at the difference and then unchecking them. I'll just take them out. they aren't needed.

PR #1194, opened 2019-09-02
[WIP] IML-1178  remove stratagem scans from MDS nodes after transfer

bm13kk:
https://github.com/whamcloud/integrated-manager-for-lustre/issues/1178

Removing each directory after mail send.

[ ] choose solution
1 collaps sync
2 create global counter for last transmission

### DIFF
--- a/iml-agent/src/action_plugins/stratagem/server.rs
+++ b/iml-agent/src/action_plugins/stratagem/server.rs
@@ -6,6 +6,7 @@ use crate::{agent_error::ImlAgentError, cmd::cmd_output_success, http_comms::mai
 use futures::{future, stream, Future, IntoFuture, Stream as _};
 use iml_fs::{read_file_to_end, stream_dir_lines, write_tempfile};
 use std::{convert::Into, path::PathBuf};
+use tokio::fs::remove_dir;
 use uuid::Uuid;
 
 /// The device that is scanned for matching rules.
@@ -362,8 +363,14 @@ pub fn trigger_scan(
 /// Streams output for all given mailbox files
 ///
 /// This fn will stream all files in parallel and return once they have all finished.
+///
+/// Issues:
+/// * mailbox_files - actually directories
+/// * Parent dir and files - should be one contract (or struct in Rust terms).
+///     This directory used only for listed files and should not used anyone else.
+/// * Used one tuple param (instead two params) to meet mk_callback() reqs.
 pub fn stream_fidlists(
-    mailbox_files: MailboxFiles,
+    (parent_dir, mailbox_files): (String, MailboxFiles),
 ) -> impl Future<Item = (), Error = ImlAgentError> {
     let mailbox_files = mailbox_files.into_iter().map(|(file, address)| {
         stream_dir_lines(file)
@@ -377,7 +384,9 @@ pub fn stream_fidlists(
             })
     });
 
-    future::join_all(mailbox_files).map(|_| ())
+    future::join_all(mailbox_files)
+        .and_then(|_| remove_dir(parent_dir).from_err())
+        .map(drop)
 }
 
 #[cfg(test)]
--- a/iml-agent/src/action_plugins/stratagem/server.rs
+++ b/iml-agent/src/action_plugins/stratagem/server.rs
@@ -363,12 +363,6 @@ pub fn trigger_scan(
 /// Streams output for all given mailbox files
 ///
 /// This fn will stream all files in parallel and return once they have all finished.
-///
-/// Issues:
-/// * mailbox_files - actually directories
-/// * Parent dir and files - should be one contract (or struct in Rust terms).
-///     This directory used only for listed files and should not used anyone else.
-/// * Used one tuple param (instead two params) to meet mk_callback() reqs.
 pub fn stream_fidlists(
     (parent_dir, mailbox_files): (String, MailboxFiles),
 ) -> impl Future<Item = (), Error = ImlAgentError> {
--- a/iml-agent/src/action_plugins/stratagem/server.rs
+++ b/iml-agent/src/action_plugins/stratagem/server.rs
@@ -4,9 +4,8 @@
 
 use crate::{agent_error::ImlAgentError, cmd::cmd_output_success, http_comms::mailbox_client};
 use futures::{future, stream, Future, IntoFuture, Stream as _};
-use iml_fs::{read_file_to_end, stream_dir_lines, write_tempfile};
+use iml_fs::{read_file_to_end, remove_dir_all, stream_dir_lines, write_tempfile};
 use std::{convert::Into, path::PathBuf};
-use tokio::fs::remove_dir;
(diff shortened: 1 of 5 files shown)


utopiabound:
You should run `cargo fmt` on your changes to fix the rust style and submit a patch to this branch

jgrund:
In order to effectively remove the parent dir we can use the parent dir passed in the previous step. 

On the manager side we can update the code here:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/5c488c1e3166177acfd14fe58cb9448840c79663/chroma_core/models/stratagem.py#L397-L406

The ignored assignment on line 397 is actually the parent dir so that needs to be assigned and then passed as an additional parameter to `invoke_rust_agent_expect_result` on line 406.

iml-agent/src/action_plugins/stratagem/server.rs
```diff
@@ -377,7 +384,12 @@ pub fn stream_fidlists(
             })
     });
 
-    future::join_all(mailbox_files).map(|_| ())
+    future::join_all(mailbox_files)
+        .and_then(|_| {
+            remove_dir(parent_dir);
```
jgrund: This needs to be returned from the `and_then` combinator, otherwise it won't get polled to completion.

Something like:

```rust
        .and_then(|_| {
            remove_dir(parent_dir).from_err()
         })
```

iml-agent/src/action_plugins/stratagem/server.rs
```diff
@@ -377,7 +384,12 @@ pub fn stream_fidlists(
...
+            Ok(())
+        })
+        .map(|_| ())
```
jgrund:
```suggestion
        .map(drop)
```

iml-agent/src/action_plugins/stratagem/server.rs
```diff
@@ -362,8 +363,14 @@ pub fn trigger_scan(
 /// Streams output for all given mailbox files
 ///
 /// This fn will stream all files in parallel and return once they have all finished.
+///
+/// Issues:
+/// * mailbox_files - actually directories
```
jgrund: Perhaps rename the type alias instead of commenting on this?

iml-agent/src/action_plugins/stratagem/server.rs
```diff
@@ -362,8 +363,14 @@ pub fn trigger_scan(
...
+/// * Parent dir and files - should be one contract (or struct in Rust terms).
```
jgrund: Not sure what this referring to, the tuple below?

iml-agent/src/action_plugins/stratagem/server.rs
```diff
@@ -362,8 +363,14 @@ pub fn trigger_scan(
...
+///     This directory used only for listed files and should not used anyone else.
+/// * Used one tuple param (instead two params) to meet mk_callback() reqs.
```
jgrund: This comment is not necessary as all "multi arg" action-plugins need to be inside a tuple.

iml-fs/src/remove_dir_all.rs
```diff
@@ -0,0 +1,67 @@
...
+    type Item = ();
+    type Error = io::Error;
+
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        blocking_io(|| fs::remove_dir_all(&self.path))
+    }
+}
+
+/// tokio-fs.lib
+fn blocking_io<F, T>(f: F) -> Poll<T, io::Error>
```
jgrund: This is unnecessary as blocking_io is already part of tokio 0.1.

Take a look at https://github.com/tokio-rs/tokio/blob/59fb5b9a7d89ee84389b8f8a039e874ef8971dbd/tokio-fs/src/create_dir.rs for example of how this can be done.

iml-fs/src/remove_dir_all_tokio.rs
```diff
@@ -0,0 +1,75 @@
+use futures::{
```
jgrund: Any reason why the patch that goes to tokio can't be the same as the remove_dir_all file? `std::task, std::pin are not being used in 0.1.x code: https://github.com/tokio-rs/tokio/blob/v0.1.x/tokio-fs/src/create_dir.rs

iml-fs/src/remove_dir_all.rs
bm13kk: But this function is private

jgrund: Ah, that's too bad. 

So, Instead of carrying tokio internals locally, I'd argue we should create a local version that has similar semantics but is much simpler.

```
pub fn remove_dir_all<P: AsRef<Path>>(p: P) -> impl Future<Item = (), Error = io::Error> {
    poll_fn(move || {
        blocking(|| fs::remove_dir_all(&p)).map_err(|_| panic!("the threadpool shut down"))
    })
    .and_then(std::convert::identity)
}
```

We should still submit an upstream patch to tokio 0.1.x, and once that lands we should remove this fn.

iml-fs/src/remove_dir_all_tokio.rs
```diff
@@ -0,0 +1,75 @@
...
+    Async,
+    Poll as FeaturePoll,
+};
+use std::{
+    fs,
+    future::Future,
+    io,
+    path::Path,
+    pin::Pin,
```
jgrund: This is not used in 0.1.x

iml-fs/src/remove_dir_all_tokio.rs
```diff
@@ -0,0 +1,75 @@
...
+{
+    type Output = io::Result<()>;
+
+    fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
+        blocking_io(|| fs::remove_dir_all(&self.path))
+    }
+}
+
+/// tokio-fs.lib
+fn blocking_io<F, T>(f: F) -> Poll<io::Result<T>>
```
jgrund: This won't be needed for the tokio PR.

iml-fs/src/remove_dir_all_tokio.rs
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+impl<P> Future for RemoveDirAllFuture<P>
+where
+    P: AsRef<Path>,
+{
+    type Output = io::Result<()>;
+
+    fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
```
jgrund: tokio 0.1.x implementations don't use `Pin`

jgrund:
Superseded by #1329

utopiabound:
For aging out old mailboxes you can update `iml-mailbox.conf` (see man page tmpfiles.d(5)).
I think you need to add an age to the `d` line in that file.

jgrund:
Closing as this is superseded by:#1329

PR #1199, opened 2019-09-05
Add check_ha() to iml-agent

### LINKED ISSUE #1190: Create `check_ha` Action Plugin
Create `check_ha` Action Plugin

utopiabound:
Returns list of resources configured to run on a given host.

Fixes #1190


iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -0,0 +1,91 @@
...
+            agent: ResourceAgentType::new(
+                elem.get_attr("class").unwrap().to_string(),
+                elem.get_attr("provider").map(|s| s.to_string()),
+                elem.get_attr("type").unwrap().to_string(),
+            ),
+            group: group,
+            id: elem.get_attr("id").unwrap().to_string(),
+            args: match elem.find("instance_attributes") {
+                None => vec![],
+                Some(e) => e
```
jgrund: Args is a `HashMap`. 

Instead of collecting the empty vec and transforming it, you can just return an empty hashmap directly:

```rust
args: match elem.find("instance_attributes") {
                None => HashMap::new(),
                Some(e) => e
                    .find_all("nvpair")
                    .map(|nv| {
                        (
                            nv.get_attr("name").unwrap().to_string(),
                            nv.get_attr("value").unwrap().to_string(),
                        )
                    })
                    .collect(),
            },
```

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -0,0 +1,91 @@
...
+            }
+            .iter()
+            .cloned()
+            .collect(),
+        }
+    }
+}
+
+pub fn check_ha(_: ()) -> impl Future<Item = Vec<AgentInfo>, Error = ImlAgentError> {
+    cmd_output_success("cibadmin", &["--query", "--xpath", "//resources"]).and_then(|output| {
```
jgrund: nit: you can get rid of the extra wrappings / unwrapping with something like:

```rust
    cmd_output_success("cibadmin", &["--query", "--xpath", "//resources"])
        .and_then(|output| Element::from_reader(output.stdout.as_slice()).map_err(|e| e.into()))
        .map(|elem| {
            elem.find_all("group")
                .flat_map(|g| {
                    let name = g.get_attr("id").unwrap().to_string();
                    g.find_all("primitive")
                        .map(move |p| AgentInfo::new(p, Some(name.clone())))
                })
                .chain(elem.find_all("primitive").map(|p| AgentInfo::new(p, None)))
                .collect()
        })
```

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub group: Option<String>,
+    pub id: String,
+    pub args: HashMap<String, String>,
+}
+
+impl AgentInfo {
+    pub fn new(elem: &Element, group: Option<String>) -> AgentInfo {
+        AgentInfo {
+            agent: ResourceAgentType::new(
+                elem.get_attr("class").unwrap().to_string(),
```
jgrund: Should we replace these `unwraps` with a conversion to a specific error type? If so, then we can use a type similar to:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/0aa188ef4c4d954784a9144c6b4923f3cdbfa751/iml-agent/src/agent_error.rs#L10-L11

Where we can put the name of the attribute that was missing in the field.

That way, we avoid panicing and can return the missing field name to the caller.

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -0,0 +1,91 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::{agent_error::ImlAgentError, cmd::cmd_output_success};
```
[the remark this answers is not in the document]
utopiabound: That's probably a good idea.  I can open a new issue as technical debt.

iml-agent/src/action_plugins/check_ha.rs
utopiabound: This doesn't work because Element does not support the Send trait.

iml-agent/src/cli.rs
```diff
@@ -305,5 +309,13 @@ fn main() {
                 };
             }
         },
+        App::CheckHA => match check_ha::check_ha(()).wait() {
+            Ok(v) => {
+                for e in v {
```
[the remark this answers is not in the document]
utopiabound: I would rather have it as json, for debugging.

iml-agent/src/action_plugins/check_ha.rs
utopiabound: I could not coerce this into working correctly.  Trying to use "?" from within the `flat_map()` in `check_ha()` would throw an error:
```
error[E0277]: the `?` operator can only be used in a function that returns `Result` or `Option` (or another type that implements `std::ops::Try`)
  --> iml-agent/src/action_plugins/check_ha.rs:88:67
   |
88 |                         .chain(elem.find_all("primitive").map(|p| AgentInfo::create(p, None)?))
   |                                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot use the `?` operator in a function that returns `action_plugins::check_ha::AgentInfo`
   |
   = help: the trait `std::ops::Try` is not implemented for `action_plugins::check_ha::AgentInfo`
   = note: required by `std::ops::Try::from_error`
```

iml-agent/src/action_plugins/check_ha.rs
jgrund: Hrrm, I tried this earlier and it compiled. Perhaps I had some other changes with it

PR #1202, opened 2019-09-09
Update to Django 1.11.23

jgrund:
This patch updates Django from 1.6 to 1.11. 

The update has many changes, probably the most significant is the removal of south and the introduction of a new migration system that looks at the model schema for migrations. https://docs.djangoproject.com/en/2.2/releases/1.7/#schema-migrations.

As part of this patch, We've removed all 40 something south migrations, they have been replaced with 3 migrations in the new Django migration system.

 In addition, Tastypie has been bumped from 0.12 to 0.14.


chroma_core/management/__init__.py
```diff
@@ -68,5 +67,3 @@ def all_subclasses(cls):
 # before our hook, so that Permission objects will be there
 # by the time we are called.
 import django.contrib.auth.management
-
-post_migrate.connect(setup_groups)
```
utopiabound: How does `setup_groups` get called now?

jgrund: It gets called here:

https://github.com/whamcloud/integrated-manager-for-lustre/pull/1202/files?file-filters%5B%5D=#diff-ea98e1400611bc4bbb2e0d030794f7d1R18

post migration hooks are now only safe to call within the ready block of an `AppConfig`

chroma_core/models/host.py
```diff
@@ -46,33 +46,14 @@
 from chroma_core.services.job_scheduler import job_scheduler_notify
 from iml_common.lib.util import ExceptionThrowingThread
 from chroma_core.models.sparse_model import VariantDescriptor
+from django.contrib.postgres.aggregates import BoolOr
```
johnsonw: I might be missing it, but how is this being used?

iml-wire-types/wire-types.rs
```diff
@@ -542,7 +542,7 @@ pub struct Volume {
     pub kind: String,
     pub label: String,
     pub resource_uri: String,
-    pub size: Option<String>,
+    pub size: Option<i64>,
```
johnsonw: Is there ever a case in which size could be negative?

python-iml-manager.spec
```diff
@@ -95,48 +94,14 @@ Requires:       rust-iml-cli
 Requires:       influxdb
 Requires:       grafana
 
-Conflicts:      chroma-agent
+Conflicts:       chroma-agent
```
johnsonw: Do we need the extra space?

jgrund: Just making it line up with the other lines.

iml-wire-types/wire-types.rs
jgrund: I hope not, it would not be a very useful storage device :)

chroma_core/models/host.py
jgrund: It's being used here:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/52bc9f87b7c23a840f2a2a6129ecb7e577dad340/chroma_core/models/host.py#L392

chroma_api/chroma_model_resource.py
```diff
@@ -40,25 +40,10 @@ def build_schema(self):
 
         return data
 
-    def obj_update(self, bundle, **kwargs):
```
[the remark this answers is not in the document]
jgrund: This was being done to override an older version of tastypie. The new version of tastypie doesn't need this.

`obj_update` is called on the superclass now.

iml-wire-types/wire-types.rs
johnsonw: Should it be a u64?

python-iml-manager.spec
johnsonw: ah ok it looked like it was spaced over in the pr.

iml-wire-types/wire-types.rs
jgrund: Ah, right. I remember why I did this. The volume size is stored in the database as a `BIGINT` which has a max value of `9223372036854775807` Which corresponds to the max_value of an i64 in rust: https://doc.rust-lang.org/std/primitive.i64.html#method.max_value. If we were to use a u64 instead, it would have the potential to overflow, as it's max value is `18446744073709551615`

I'm in the middle of redoing device persistence in IML, I guess I should use https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.DecimalField for the update.

johnsonw: Ok that makes sense 👍

jgrund:
This had a full SSI pass with: https://build.whamcloud.com/job/integration-tests-shared-storage-configuration/2693/

jgrund:
Landed in 988b68277078edb694162e24f8b25ee538c6b19f

Issue #1191, opened 2019-08-28
Create `check_power_control` Action Plugin


johnsonw:
Create `check_power_control` Action Plugin

utopiabound:
Proposed API:
```check_stonith() -> (configured: Boolean, explination: String)```

PR #1205, opened 2019-09-11
Add check_stonith

utopiabound:
Call to check_stonith() to see if stonith is enabled and configured.
Return true/false plus explanitory text.

Fixes Issue #1191 

### DIFF
--- a/iml-agent/src/action_plugins/check_stonith.rs
+++ b/iml-agent/src/action_plugins/check_stonith.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use elementtree::Element;
 use futures::Future;
+use iml_wire_types::{ComponentState, ConfigState, ServiceState};
 use std::process::Command;
 use tracing::{span, Level};
 use tracing_futures::Instrument;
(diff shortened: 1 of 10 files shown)


iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,64 @@
...
+        // @@TODO: This should check for node name in
+        // instance_attributes:pcmk_host_list if
+        // pcmk_host_check=static-list
+
+        // @@TODO: This should check meta_attributes:target-role is
+        // not Stopped (disabled agent)
+    }
+}
+
+pub fn check_stonith(_: ()) -> impl Future<Item = (bool, String), Error = ImlAgentError> {
```
johnsonw: This looks fine to me but one thing we may want to consider is to move the `cmd_output` outside of this function and instead let it take the future that is returned. This would allow us to write a unit test with a sample cibadmin output.

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    #[test]
+    fn test_stonith_unconfigured_fence_chroma() -> agent_error::Result<()> {
+        let testxml = r#"<primitive class="stonith" id="st-fencing" type="fence_chroma"/>"#;
+        assert_eq!(
+            _check_stonith(&testxml.as_bytes()).unwrap(),
+            (false, "fence_chroma - unconfigured".to_string())
+        );
+
+        Ok(())
```
johnsonw: Why do we need to return a result of type unit here?

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,174 @@
...
+            None => Ok((false, "fence_chroma - unconfigured".to_string())),
+            Some(_) => Ok((true, "fence_chroma".to_string())),
+        },
+        Some(t) => Ok((true, t.to_string())),
+        // @@TODO: This should check for node name in
+        // instance_attributes:pcmk_host_list if
+        // pcmk_host_check=static-list
+
+        // @@TODO: This should check meta_attributes:target-role is
+        // not Stopped (disabled agent)
```
jgrund: Is there any reason these can't be implemented now?

iml-agent/src/action_plugins/check_stonith.rs
utopiabound: I guess it doesn't, will fix

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,272 @@
...
+};
+use elementtree::Element;
+use futures::Future;
+use iml_wire_types::{ComponentState, ConfigState, ServiceState};
+use std::process::Command;
+use tracing::{span, Level};
+use tracing_futures::Instrument;
+
+fn inst_attr_problem(elem: &Option<&Element>, node: &String) -> Option<String> {
+    match elem {
```
jgrund: Have you tried using the `?` operator here? Since you are dealing with a bunch of matches / nesting, and returning an `Option`, it would dedent this code a bit.

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,272 @@
...
+                    _ => (),
+                }
+            }
+            None
+        }
+    }
+}
+
+fn meta_attr_problem(elem: &Option<&Element>) -> Option<String> {
+    match elem {
```
jgrund: Same as above. The `?` can reduce nesting.

johnsonw:
I think it looks good. I agree with Joe that we could trim the nesting up a bit using the `?` operator.

jgrund:
This can be rebased on master now.

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,314 @@
...
+        Some(instance) => {
+            let mut hc: Option<bool> = None;
+            let mut hl: Option<bool> = None;
+            for e in instance.children() {
+                match e.get_attr("name") {
+                    Some("pcmk_host_list") => match e.get_attr("value") {
+                        None => return None,
+                        Some(val) => {
+                            if val.contains(node) {
+                                match hc {
```
johnsonw: I might be reading this incorrectly but it looks like  `hc` will always be `None` right?

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,314 @@
...
+use std::default::Default;
+use std::process::Command;
+use tracing::{span, Level};
+use tracing_futures::Instrument;
+
+fn inst_attr_problem(elem: &Option<&Element>, node: &String) -> Option<String> {
+    match elem {
+        None => None,
+        Some(instance) => {
+            let mut hc: Option<bool> = None;
```
johnsonw: This feels really busy. Is there any way this could be cleaned up? We might be able to pull some of the logic into functions. What are your thoughts?

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,314 @@
...
+                                    }
+                                    None => hl = Some(true),
+                                }
+                            } else {
+                                hl = Some(false);
+                            }
+                        }
+                    },
+                    Some("pcmk_host_check") => match e.get_attr("value") {
+                        Some("static-list") => match hl {
```
johnsonw: Isn't `hl` equal to `None`?

johnsonw:
I'm seeing files from a different branch mixed in.

iml-agent/src/action_plugins/check_stonith.rs
utopiabound: No, I get:
```
error[E0277]: the `?` operator can only be applied to values that implement `std::ops::Try`
  --> iml-agent/src/action_plugins/check_stonith.rs:63:14
   |
63 |     for e in elem?.from_err().children() {
   |              ^^^^^ the `?` operator cannot be applied to type `&std::option::Option<&elementtree::Element>`
   |
   = help: the trait `std::ops::Try` is not implemented for `&std::option::Option<&elementtree::Element>`
   = note: required by `std::ops::Try::into_result`
```
I can flatten this more though

iml-agent/src/action_plugins/check_stonith.rs
utopiabound: No, It's a big set of match statements, hc and hl can be set in either order, I'll add a comment to make it clearer

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -8,74 +8,78 @@ use crate::{
...
-                            Some(true) => return None,
-                            Some(false) => {
-                                return Some(format!("{} missing from pcmk_host_list", node))
-                            }
-                            None => hc = Some(true),
-                        },
-                        _ => return None,
-                    },
-                    _ => (),
+    if let Some(instance) = elem {
```
johnsonw: This feels a little awkward to me because an option will either return something or nothing. In this case, the something will always be an error message and None is returned on success. It seems like a `Result` type would fit better here. You could then return Ok(()) when successful or Err with the error message if there is a problem.

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -8,74 +8,78 @@ use crate::{
...
     }
+    None
 }
 
+/// This processes <meta_attributes> section of a stonith <primitive>
+/// from the pacemaker cib
+///
+/// Returns: Some("ERROR MESSAGE") on an issue, or None for no issues
+///
 fn meta_attr_problem(elem: &Option<&Element>) -> Option<String> {
```
johnsonw: Same here, I think the result type might fit better.

PR #1207, opened 2019-09-14
Ntp daemon

johnsonw:
Ntp daemon

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1207)


iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,536 @@
...
+            .into();
+
+        assert_eq!(
+            get_chrony_time_offset(s),
+            Some("0.000026767 seconds".into())
+        );
+    }
+
+    #[test]
+    fn test_session_with_ntp_configured_by_iml() {
```
johnsonw: I'm trying to find a better way to do this. The idea is that I implement the SideEffects trait for the TestNtp struct. This works but unfortunately I have to re-implement the DaemonPlugin trait. I was hoping I could somehow wrap the Ntp struct, implement the SideEffect trait, and then I wouldn't need to re-implement DaemonPlugin. This seems to be more difficult than I anticipated though.

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,536 @@
...
+use iml_util::Flatten;
+use iml_wire_types::AgentResult;
+use lazy_static::lazy_static;
+use parking_lot::RwLock;
+use regex::Regex;
+use serde_json::json;
+use std::time::{SystemTime, UNIX_EPOCH};
+use std::{path::Path, process::Command};
+use std::{process, sync::Arc};
+use tokio_process::CommandExt;
```
jgrund: Lot's of unused imports here.

iml-agent/src/daemon_plugins/ntp.rs
jgrund: Why do we need to implement an entire `SideEffects` trait? I.E. What do we gain by testing DaemonPlugin instead of testing individual calls?

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,536 @@
...
+pub struct TimeStatus {
+    synced: TimeSync,
+    offset: Option<String>,
+}
+
+fn is_ntp_synced(output: String) -> TimeSync {
+    match output.as_ref() {
+        "b true" => TimeSync::Synced,
+        "b false" => TimeSync::Unsynced,
+        _ => TimeSync::Unsynced,
```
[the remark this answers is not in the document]
johnsonw: Yes of course. The manager plugin will be monitoring the results and if it gets Unsynced it will generate a warning on the manager side. I'm working on piece now.

iml-ntp/main.rs
```diff
@@ -0,0 +1,147 @@
...
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    env_logger::init();
+
+    loop { 
+        log::info!("Entering loop and connecting to rabbit.");
+        let client = connect_to_rabbit().compat().await?;
+
+        tokio::spawn(async move {
```
jgrund: likewise I don't think you should be spawning to a new thread here. 

You'll basically want to pull from the service queue in the main thread.

Cargo.toml
```diff
@@ -15,5 +15,6 @@ members = [
     'iml-manager-client',
     'iml-fs',
     'iml-util',
-    'iml-postgres'
+    'iml-postgres',
```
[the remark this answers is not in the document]
johnsonw: I will move it back. We were discussing where to put the services so I moved it outside until we finalized the location. I'll move it back.

iml-ntp/main.rs
```diff
@@ -0,0 +1,147 @@
...
+use iml_postgres::connect;
+use tokio_postgres::{Client};
+use iml_rabbit::connect_to_rabbit;
+use iml_services::service_queue::{consume_service_queue, data_only, into_deserialized};
+use iml_wire_types::{ntp::{TimeSync, TimeStatus}, Fqdn};
+use std::io;
+
+/// This async function will retrieve the content type id of the specified model. A future
+/// containing the content type id of the model is returned.
+async fn get_managed_host_content_type_id(
```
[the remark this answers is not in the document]
johnsonw: sure

iml-ntp/main.rs
```diff
@@ -0,0 +1,147 @@
...
+        .map_err(|e: tokio_postgres::error::Error| NtpError::IoError(io::Error::new(io::ErrorKind::Other, e.to_string())))
+        .await?
+        .map_err(|e: tokio_postgres::error::Error| NtpError::IoError(io::Error::new(io::ErrorKind::Other, e.to_string())))
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    env_logger::init();
+
+    loop {
```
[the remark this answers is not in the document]
johnsonw: So this confused me as I didn't want to put the loop here but without it I get an error. Also, the docs show that they are using a loop as well: https://docs.rs/tokio/0.2.0-alpha.5/tokio/index.html

iml-ntp/main.rs
johnsonw: ok, maybe that's why i'm getting an access error. I'll update.

Cargo.toml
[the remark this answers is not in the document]
jgrund: I'll repush my patch stack, I've moved https://github.com/whamcloud/integrated-manager-for-lustre/pull/1220 on top of the django bump.

iml-ntp/main.rs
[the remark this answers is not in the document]
jgrund: yeah, the loop is not necessary, it's just there as it's useful for their example.

johnsonw:
This update shows that when the ntpd service is stopped on one of the server nodes the manager will pick up the state and display an alert. Likewise, when the ntpd service is started again and is in sync, the alert will be removed from the gui.

ntp_alert_gui

iml-agent/Cargo.toml
```diff
@@ -41,6 +41,7 @@ url = "1.7"
 v_hist = "0.1.1"
 uuid = { version = "0.7", features = ["v4"] }
 parking_lot = "0.9"
+regex = "1.3.1"
```
jgrund: What are we using this for?

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,346 @@
...
+}
+
+fn is_ntp_synced(output: &str) -> TimeSync {
+    let x = match output {
+        "b true" => TimeSync::Synced,
+        "b false" => TimeSync::Unsynced,
+        _ => TimeSync::Unsynced,
+    };
+
+    info!("is_ntp_synced: {:?}; equal? {}", x, output == "b true");
```
jgrund: This should be `debug`. We don't want a message for each tick.

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,346 @@
...
+
+fn is_ntp_configured_by_iml() -> impl Future<Item = bool, Error = ImlAgentError> {
+    iml_fs::stream_file_lines(Path::new(NTP_CONFIG_FILE))
+        .filter(|l| l.find("# IML_EDIT").is_some())
+        .from_err()
+        .fold(vec![], |mut acc, x| {
+            acc.push(x);
+            future::ok::<_, ImlAgentError>(acc)
+        })
+        .and_then(|x: Vec<String>| future::ok::<_, ImlAgentError>(!x.is_empty()))
```
jgrund: This may be more succinct by using `into_future`:


```
.into_future()
.map(|(x, _)| x.is_some())
```

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,346 @@
...
+    F4: Future<Item = TimeStatus, Error = ImlAgentError>,
+>(
+    is_ntp_configured_by_iml: fn() -> F1,
+    get_time_sync_services: fn() -> F2,
+    ntpd_synced: fn() -> F3,
+    chronyd_synced: fn() -> F4,
+) -> impl Future<Item = Output, Error = ImlAgentError> {
+    is_ntp_configured_by_iml()
+        .and_then(move |configured| {
+            info!("Configured: {:?}", configured);
```
jgrund: This should be `debug!`.

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,346 @@
...
+            get_time_sync_services().map(move |tss| {
+                info!("time_sync_service shows: {:?}", tss);
```
jgrund: This should be `debug!`.

iml-services/iml-ntp/cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+
+[dependencies]
+lapin-futures = "0.18.0"
+tokio = "0.2.0-alpha.4"
+tokio-timer = "0.2.11"
+tokio-postgres = "0.4.0-rc.3"
+env_logger = "0.6.1"
+log = "0.4.6"
```
jgrund: use tracing, tracing-subscriber instead.

iml-services/iml-ntp/cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+serde = { version = "1", features = ["derive"] }
+serde_json = "1.0"
+failure = "0.1.5"
```
jgrund: Let's not use failure, it adds a bunch of overhead.

iml-services/iml-ntp/cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+warp = "0.1.16" 
+iml-wire-types = { path = "../../iml-wire-types", version = "0.2" }
+iml-rabbit = { path = "../../iml-rabbit", version = "0.1.0" }
+iml-postgres = { path = "../../iml-postgres", version = "0.1.0" }
+iml-manager-env = { path = "../../iml-manager-env", version = "0.1.0" }
+iml-service-queue = { path = "../iml-service-queue", version = "0.1.0" }
+tokio-runtime-shutdown = { path = "../../tokio-runtime-shutdown", version = "0.1.0" }
+parking_lot = "0.8"
```
jgrund: Are you using this?

iml-services/iml-ntp/src/error.rs
```diff
@@ -0,0 +1,37 @@
+use std::io;
```
jgrund: copyright.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+#![feature(async_await)]
```
jgrund: Not needed.

iml-services/iml-ntp/cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "iml-ntp"
+version = "0.1.0"
+authors = ["IML Team <iml@whamcloud.com>"]
+edition = "2018"
+
+[dependencies]
```
jgrund: Audit the deps / dev-deps, I don't think you're using many of these.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+
+mod error;
```
jgrund: probably better to have a lib.rs with this in it.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+        .await?
+        .map(|row| row.get::<_, i32>(0))
+        .ok_or_else(|| NtpError::IoError(io::Error::new(io::ErrorKind::NotFound, "managedhost")))
+}
+
+/// This async function will retrieve the managed host id for a given fqdn. A future
+/// containing the managed host id is returned.
+async fn get_managed_host_id(fqdn: &str, client: &mut Client) -> Result<i32, NtpError> {
+    client
+        .prepare("SELECT id FROM chroma_core_managedhost WHERE fqdn=$1;")
```
jgrund: You should use the [Name trait](https://docs.rs/iml-wire-types/0.2.0/iml_wire_types/db/struct.ManagedHostRecord.html#impl-Name) instead of hardcoding this. 

That way if we change table names we can do that in one spot.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+};
+use std::io;
+use tokio_postgres::Client;
+use tracing_subscriber::{fmt::Subscriber, EnvFilter};
+
+/// This async function will retrieve the content type id of the specified model. A future
+/// containing the content type id of the model is returned.
+async fn get_managed_host_content_type_id(client: &mut Client) -> Result<i32, NtpError> {
+    client
+        .prepare("SELECT id FROM django_content_type WHERE model='managedhost';")
```
jgrund: Should create a [`Name` trait](https://docs.rs/iml-wire-types/0.2.0/iml_wire_types/db/trait.Name.html) for this in wire-types.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+            NtpError::IoError(io::Error::new(
+                io::ErrorKind::NotFound,
+                "Error getting host id",
+            ))
+        })
+}
+
+async fn get_active_alert_for_fqdn(fqdn: &str, client: &mut Client) -> Result<i32, NtpError> {
+    client
+        .prepare(r#"SELECT S.id FROM chroma_core_alertstate AS S LEFT JOIN chroma_core_managedhost AS MH
```
jgrund: Use  [`Name` traits](https://docs.rs/iml-wire-types/0.2.0/iml_wire_types/db/trait.Name.html)  for the table names.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+        .try_next()
+        .await?
+        .map(|row| row.get::<_, i32>(0))
+        .ok_or_else(||  NtpError::IoError(io::Error::new(io::ErrorKind::NotFound, "Couldn't retrieve active alert id.")))
+}
+
+async fn set_alert_inactive(id: i32, client: &mut Client) -> Result<u64, NtpError> {
+    client
+        .prepare(&format!(
+            r#"UPDATE chroma_core_alertstate
```
jgrund: Use Name traits for the table names.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+SET active=Null,
+    "end"=Now()
+WHERE
+    id=$1;"#
+        ))
+        .compat()
+        .and_then(move |s| client.execute(&s, &[&id]).compat())
+        .map(Ok)
+        .map_err(|e: tokio_postgres::error::Error| {
+            NtpError::IoError(io::Error::new(io::ErrorKind::Other, e.to_string()))
```
jgrund: The error should be preserved here instead of converted to a string.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+        })
+        .await?
+        .map_err(|e: tokio_postgres::error::Error| {
+            NtpError::IoError(io::Error::new(io::ErrorKind::Other, e.to_string()))
```
jgrund: The error should be preserved here instead of converted to a string.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+    client
+        .prepare(
+            &format!(r#"INSERT INTO chroma_core_alertstate
+                (record_type, variant, alert_item_id, alert_type, begin, message, active, dismissed, severity, alert_item_type_id) values
+                ($1, $2, $3, $4, Now(), $5, $6, $7, $8, $9);"#
+            )
+        )
+        .compat()
+        .and_then(move |s| {
+            client.execute(&s, &[&"NtpOutOfSyncAlert", &"{}", &alert_item_id, &"NtpOutOfSyncAlert", &format!("Ntp is not in sync on server {}", fqdn).as_str(), &true, &false, &40, &alert_item_type_id]).compat()
```
jgrund:
```suggestion
            client.execute(&s, &[&"NtpOutOfSyncAlert", &"{}", &alert_item_id, &"NtpOutOfSyncAlert", &format!("Ntp is out of sync on server {}", fqdn).as_str(), &true, &false, &40, &alert_item_type_id]).compat()
```

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+        })
+        .map(Ok)
+        .map_err(|e: tokio_postgres::error::Error| NtpError::IoError(io::Error::new(io::ErrorKind::Other, e.to_string())))
```
jgrund: The error should be preserved here instead of converted to a string.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+        .await?
+        .map_err(|e: tokio_postgres::error::Error| NtpError::IoError(io::Error::new(io::ErrorKind::Other, e.to_string())))
```
jgrund: The error should be preserved here instead of converted to a string.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,166 @@
...
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let subscriber = Subscriber::builder()
+        .with_env_filter(EnvFilter::from_default_env())
+        .finish();
+    tracing::subscriber::set_global_default(subscriber).unwrap();
+
+    while let Some((fqdn, time_status)) = consume_ntp_queue().try_next().await? {
+        tracing::info!("fqdn: {:?} time_status: {:?}", fqdn, time_status);
```
jgrund: `debug!`

iml-agent/Cargo.toml
johnsonw: The regex is used to get the time offset in the ntp daemon plugin.

iml-agent/src/daemon_plugins/ntp.rs
johnsonw: agreed, I will update all of my comments to use a debug level.

iml-services/iml-ntp/cargo.toml
johnsonw: Remove both env_logger and log

iml-services/iml-ntp/cargo.toml
johnsonw: removed

iml-services/iml-ntp/cargo.toml
johnsonw: Doesn't look like it. Removed.

iml-services/iml-ntp/src/main.rs
johnsonw: Removed

iml-services/iml-ntp/cargo.toml
johnsonw: I went through each of the items and reduced it to:

```
[package]
name = "iml-ntp"
version = "0.1.0"
authors = ["IML Team <iml@whamcloud.com>"]
edition = "2018"

[dependencies]
tokio = "0.2.0-alpha.4"
tokio-postgres = "0.4.0-rc.3"
iml-wire-types = { path = "../../iml-wire-types", version = "0.2" }
iml-rabbit = { path = "../../iml-rabbit", version = "0.1.0" }
iml-postgres = { path = "../../iml-postgres", version = "0.1.0" }
iml-service-queue = { path = "../iml-service-queue", version = "0.1.0" }
futures-preview = { version = "0.3.0-alpha.18", features = ["compat"] }
tracing = "0.1.9"
tracing-subscriber = "0.1.3"
```

iml-services/iml-ntp/src/main.rs
johnsonw: sure, sounds good.

johnsonw: Created a lib.rs and define the error mod as well as the iml_ntp mod. To pull the error in I just use `use crate::error`.

iml-services/iml-ntp/src/main.rs
johnsonw: Makes sense, though isn't a trait used for an instance? I think what I need is an associative function, right?
```
let table_name = ManagedHostRecord::table_name();
```
For now, i'll use the MANAGED_HOST_TABLE_NAME const.

johnsonw: Updated

iml-services/iml-ntp/src/main.rs
johnsonw: Looks like I can get rid of this function after a little more research.

iml-services/iml-ntp/src/db.rs
```diff
@@ -0,0 +1,88 @@
...
+/// containing the managed host id is returned.
+pub async fn get_managed_host_items(
+    fqdn: &str,
+    client: &mut Client,
+) -> Result<Option<(i32, i32, String)>, iml_postgres::Error> {
+    iml_postgres::select(
+        client,
+        format!(
+            "SELECT id, content_type_id, state FROM {} WHERE fqdn=$1 AND not_deleted=True;",
+            ManagedHostRecord::table_name().to_string()
```
jgrund: Why is `to_string` needed here?

iml-services/iml-ntp/src/db.rs
```diff
@@ -0,0 +1,88 @@
...
+    iml_postgres::select(
+        client,
+        format!(
+            r#"SELECT S.id FROM {} AS S LEFT JOIN {} AS MH 
+ON S.alert_item_id = MH.id WHERE S.record_type='NtpOutOfSyncAlert' AND 
+S.active = True AND MH.fqdn=$1 AND MH.not_deleted=True;"#,
+            AlertStateRecord::table_name(),
+            ManagedHostRecord::table_name(),
+        )
+        .as_str(),
```
jgrund: nit: could just do a borrow here instead.

iml-services/iml-ntp/src/db.rs
```diff
@@ -0,0 +1,88 @@
...
+    .await
+    .map(|row| row.map(|r| r.get::<_, i32>(0)))
+}
+
+pub async fn set_alert_inactive(id: i32, client: &mut Client) -> Result<u64, iml_postgres::Error> {
+    iml_postgres::update(
+        client,
+        &format!(
+            r#"UPDATE {} SET active=Null, "end"=Now() WHERE id=$1;"#,
+            AlertStateRecord::table_name().to_string()
```
jgrund: Why is to_string needed here?

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,68 @@
...
+
+use futures::stream::{StreamExt, TryStreamExt};
+use iml_postgres::connect;
+use iml_wire_types::{ntp::{TimeSync, TimeStatus}, Fqdn};
+use iml_service_queue::service_queue::consume_data;
+use tracing_subscriber::{fmt::Subscriber, EnvFilter};
+
+pub fn valid_state(mh_state: String) -> Option<String> {
+    ["monitored", "managed", "working"]
+        .iter()
```
jgrund: If you use into_iter here instead, I think you can avoid the to_string at the end.

iml-services/iml-ntp/src/db.rs
johnsonw: removed

iml-services/iml-ntp/src/db.rs
johnsonw: sure

iml-services/iml-ntp/src/db.rs
johnsonw: removed

iml-services/iml-ntp/src/main.rs
johnsonw: I moved the `is_some` inside the function so it can return a bool, which will be nicer to work with.

iml-agent/src/daemon_plugins/ntp.rs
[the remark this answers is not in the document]
utopiabound: I meant like a trace::warning() to log an unexpected value,

iml-ntp/cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+iml-rabbit = { path = "../iml-rabbit", version = "0.1.0" }
+iml-postgres = { path = "../iml-postgres", version = "0.1.0" }
+iml-manager-env = { path = "../iml-manager-env", version = "0.1.0" }
+tokio-runtime-shutdown = { path = "../tokio-runtime-shutdown", version = "0.1.0" }
+iml-services = { path = "../iml-services", version = "0.1.0" }
+parking_lot = "0.8"
+chrono = "0.4"
+futures-preview = { version = "0.3.0-alpha.18", features = ["compat"] }
+
+[dev-dependencies]
```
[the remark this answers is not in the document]
johnsonw: outdated

iml-ntp/main.rs
```diff
@@ -0,0 +1,147 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+#![feature(async_await)]
```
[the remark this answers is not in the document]
johnsonw: removed

chroma_core/models/ntp.py
```diff
@@ -114,3 +118,25 @@ def description(self):
 
     def get_steps(self):
         return [(UnconfigureNTPStep, {"ntp_configuration": self.ntp_configuration})]
+
+
+class NtpOutOfSyncAlert(AlertStateBase):
+    # When a server is out of sync from the IML node this can cause a lot of problems
+    # and is thus a high severity error.
+    def get_message(self, host):
+        "NTP out of sync on server '%s'" % host
```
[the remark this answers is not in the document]
johnsonw: upated

iml-agent/src/daemon_plugins/ntp.rs
johnsonw: These should be cleaned up now.

iml-agent/src/daemon_plugins/ntp.rs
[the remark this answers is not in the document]
johnsonw: Added a warning.

jgrund:
A few things I've noticed while testing:

1. IML reports that time is unsynced if either ntp or chrony are started but the service is not enabled. This is misleading as the time is actually synced correctly, the service is just not enabled. Instead we should be raising a separate (warning?) alert that the service is not enabled.
2. Any out of sync alert appears as `NtpOutOfSyncAlert`. This would probably be better as `TimeOutOfSyncAlert` as we may be dealing with different clients for handling time sync.
3. Using the agent cli to configure IML NTP initially, I used `iml-agent ntp configure`, but this actually unconfigures NTP. Instead the `-s` flag should be required and there should be a separate `unconfigure` command.
4. In the same configure command, ntp should be enabled and chronyd should be stopped / disabled.

jgrund:
Todo:

  - [ ] Figure out reusable model representation for orm items
  - [ ] abstract resend logic from ostpool into ntp daemon plugin
  - [ ] Make use of the rest of the state / alerts.

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,139 @@
...
+
+pub async fn get_host_by_fqdn(
+    x: &str,
+    pool: &iml_orm::DbPool,
+) -> Result<Option<ChromaCoreManagedhost>, tokio_diesel::AsyncError> {
+    ChromaCoreManagedhost::by_fqdn(x)
+        .first_async(pool)
+        .await
+        .optional()
+}
```
utopiabound, line 23: It feels like this function should be over in `iml-orm`

iml-services/iml-ntp/src/main.rs
```diff
@@ -0,0 +1,139 @@
...
+
+    tracing::subscriber::set_global_default(subscriber).unwrap();
+
+    let pool = iml_orm::pool()?;
+
+    let mut s = consume_data::<State>("rust_agent_ntp_rx");
+
+    while let Some((fqdn, state)) = s.try_next().await? {
+        tracing::debug!("fqdn: {:?} state: {:?}", fqdn, state);
+
```
utopiabound, line 39: It feels like simpler code if it were structured:
 ```rust
if let Some(host) = get_host_by_fqdn(&fqdn.0, &pool).await? {
    match state {
        State::Synced => {
             alerts::lower(
    ...
```

File: iml-services/iml-ntp/src/main.rs (hunk 37 above)
Comment (jgrund) [reply iml-services/iml-ntp/src/main.rs:23]: Maybe... this doesn't really do anything specific from a business-logic perspective; that part is already over in iml-orm.

All it does is asynchronously fetch the first record and allow for it to be empty.

File: iml-services/iml-ntp/src/main.rs (hunk 38 above)
Comment (jgrund) [reply iml-services/iml-ntp/src/main.rs:39]: Even simpler is to return early on `None` (removes a level of nesting).

**Comment by jgrund** [comment]
## Stopping time client

stop-time-client

File: .copr/Makefile [hunk 39]
```diff
@@ -34,10 +34,11 @@ substs: $(SUBSTS)
 	chmod +x $(SUBSTS_SHELL)
 
 srpm:
+	yum install -y postgresql96-devel
 	mkdir -p ${TMPDIR}/_topdir/{SOURCES,SPECS}
 	mkdir -p ${TMPDIR}/release/rust-iml
-	cargo build --release
-	cp target/release/iml-{action-runner,agent,agent-comms,agent-daemon,api,mailbox,ostpool,postoffice,stats,device,warp-drive} \
+	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo build --release
```
johnsonw, line 40: Is this related to the postgres patch?

iml-agent/src/cli.rs
```diff
@@ -508,6 +532,47 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+                match action_configure::update_and_write_new_config(server).await {
+                    Ok(_) => {
+                        println!("{}", msg);
+                        println!("Restarting ntpd daemon.");
+                        match systemd::restart_unit("ntpd".into()).await {
+                            Ok(_) => {
+                                println!("ntpd service restarted successfully.");
+                            }
+                            Err(e) => {
+                                println!("{:?}", e);
```
ip1981, line 555:
```suggestion
                                eprintln!("{:?}", e);
```

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,340 @@
...
+        let s = r#"b true"#;
+
+        assert_eq!(parse_ntp_synced(s), Some(time::Synced::Synced));
+    }
+
+    #[test]
+    fn test_is_ntp_not_synced() {
+        let s = r#"b false"#;
+
+        assert_eq!(parse_ntp_synced(s), Some(time::Synced::Unsynced));
```
ip1981, line 232: These tests are tautological. `parse_ntp_synced` is literally mapping strings to an enum.

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,340 @@
...
+        assert_eq!(parse_ntp_synced(s), None);
+    }
+
+    #[test]
+    fn test_get_ntp_time_offset() {
+        let s = r#"synchronised to NTP server (10.73.10.10) at stratum 12
+time correct to within 949 ms
+polling server every 64 s"#;
+
+        assert_eq!(parse_ntp_time_offset(s), Some("949 ms".to_string().into()));
```
ip1981, line 248: I think it would be better to have something more explicit, like `Some(949)`. But I could not find `time::Offset` :)

iml-agent/src/daemon_plugins/ntp.rs
```diff
@@ -0,0 +1,340 @@
...
+Authenticated   : No
+TX timestamping : Kernel
+RX timestamping : Kernel
+Total TX        : 1129
+Total RX        : 1129
+Total valid RX  : 1129"#;
+
+        assert_eq!(
+            parse_chrony_time_offset(s),
+            Some("0.000026767 seconds".to_string().into())
```
ip1981, line 284: Likewise here. Now it looks like you are using the same parser on both sides. And actually you are, you only use a regex to extract a string, and then parse the string the same way on both sides. Tautological :)

iml-agent/src/daemon_plugins/ntp.rs
jgrund: I didn't write these tests, but I guess the question is whether we gain anything by enforcing this invariant. 

Likewise does it harm anything to leave them here?

ip1981: No harm other than compile and execution time. And if we delete them as soon as they fail :)

iml-agent/src/daemon_plugins/ntp.rs
jgrund: We could probably represent it as a duration + a Sign (postitive, negative). However, we don't actually use this value anywhere on the manager side.

PR #1208, opened 2019-09-16
Add tests to iml-agent check_ha

utopiabound:
Issue #1190

### DIFF
--- a/iml-agent/src/action_plugins/check_ha.rs
+++ b/iml-agent/src/action_plugins/check_ha.rs
@@ -71,7 +71,7 @@ impl AgentInfo {
     }
 }
 
-fn _check_ha(output: &[u8]) -> Result<Vec<AgentInfo>, ImlAgentError> {
+fn do_check_ha(output: &[u8]) -> Result<Vec<AgentInfo>, ImlAgentError> {
     match Element::from_reader(output) {
         Err(err) => Err(ImlAgentError::XmlError(err)),
         Ok(elem) => Ok(elem
@@ -92,12 +92,12 @@ fn _check_ha(output: &[u8]) -> Result<Vec<AgentInfo>, ImlAgentError> {
 pub fn check_ha(_: ()) -> impl Future<Item = Vec<AgentInfo>, Error = ImlAgentError> {
     cmd_output_success("cibadmin", &["--query", "--xpath", "//resources"])
         .instrument(span!(Level::INFO, "Read cib"))
-        .and_then(|output| _check_ha(&output.stdout.as_slice()))
+        .and_then(|output| do_check_ha(&output.stdout.as_slice()))
 }
 
 #[cfg(test)]
 mod tests {
-    use super::{AgentInfo, ResourceAgentType, _check_ha};
+    use super::{AgentInfo, ResourceAgentType, do_check_ha};
     use crate::agent_error;
     use std::collections::HashMap;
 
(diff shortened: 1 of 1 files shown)


iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +71,146 @@ impl AgentInfo {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{AgentInfo, ResourceAgentType, _check_ha};
+    use crate::agent_error;
+    use std::collections::HashMap;
+
+    #[test]
+    fn test_ha_only_fence_chroma() -> agent_error::Result<()> {
```
johnsonw: Any reason you need to return an agent_error::Result? I think you could just do:

```
fn test_ha_only_fence_chroma() {
        let testxml = r#"<resources>
  <primitive class="stonith" id="st-fencing" type="fence_chroma"/>
</resources>
"#;
        assert_eq!(
            _check_ha(&testxml.as_bytes()).unwrap(),
            vec![AgentInfo {
                agent: ResourceAgentType::new(
                    "stonith".to_string(),
                    None,
                    "fence_chroma".to_string()
                ),
                group: None,
                id: "st-fencing".to_string(),
                args: HashMap::new()
            }]
        );
}
```

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +71,146 @@ impl AgentInfo {
...
+                group: None,
+                id: "st-fencing".to_string(),
+                args: HashMap::new()
+            }]
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_ha_mixed_mode() -> agent_error::Result<()> {
```
johnsonw: Same here, you should be able to return unit instead of agent_error::Result.

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +71,146 @@ impl AgentInfo {
     }
 }
 
+fn _check_ha(output: &[u8]) -> Result<Vec<AgentInfo>, ImlAgentError> {
```
jgrund: Any reason we need the leading `_`. It's sort of implied it's private by the lack of the `pub` keyword.

iml-agent/src/action_plugins/check_ha.rs
jgrund: Yeah, since this test it's not using `?` it doesn't need to return a result.

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +71,146 @@ impl AgentInfo {
...
+                id: "st-fencing".to_string(),
+                args: HashMap::new()
+            }]
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn test_ha_mixed_mode() -> agent_error::Result<()> {
+        let testxml = r#"<resources>
```
jgrund: Instead of maintaining this string, use https://docs.rs/insta/0.11.0/insta/ for snapshot testing.

utopiabound:
Should this be renamed to check_resources, or even get_resources, and check_ha would return something more like 
```
enum ComponentState {
    Unconfigured,
    Configured,
    Started,
    Setup, // Started + Configured
}

struct HaState {
    corosync: ComponentState,
    pacemaker: ComponentState,
    pcsd: ComponentState,
}
```

`ComponentState` may want some extra info in the states for if they are "configurable"

utopiabound:
Rolled into #1288

PR #1211, opened 2019-09-18
Fixup python cli

jgrund:
The python manager cli currently is expecting a sessionid
for a non-logged in user.

In more recent versions of Django, a sessionid is not provided
until a user has logged in.

In addition, we should update the csrftoken header after logging in to
match the csrftoken value passed back.

### DIFF
--- a/chroma_cli/api.py
+++ b/chroma_cli/api.py
@@ -78,8 +78,7 @@ class ChromaSessionClient(object):
             raise RuntimeError("No session (status: %s, text: %s)" % (r.status_code, r.content))
 
         self.session.headers["X-CSRFToken"] = r.cookies["csrftoken"]
-        self.session.cookies["csrftoken"] = r.cookies["csrftoken"]
-        self.session.cookies["sessionid"] = r.cookies["sessionid"]
+        self.session.cookies.set("csrftoken", r.cookies["csrftoken"])
 
     def login(self, **credentials):
         if not self.is_authenticated:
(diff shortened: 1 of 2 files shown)


chroma_cli/api.py
```diff
@@ -78,18 +78,20 @@ def start_session(self):
             raise RuntimeError("No session (status: %s, text: %s)" % (r.status_code, r.content))
 
         self.session.headers["X-CSRFToken"] = r.cookies["csrftoken"]
-        self.session.cookies["csrftoken"] = r.cookies["csrftoken"]
-        self.session.cookies["sessionid"] = r.cookies["sessionid"]
+        self.session.cookies.set("csrftoken", r.cookies["csrftoken"])
```
johnsonw: Should this be:
```
self.session.cookies.set("csrftoken", r.cookies.get("csrftoken")
```
?

chroma_cli/api.py
```diff
@@ -78,18 +78,20 @@ def start_session(self):
...
         if not self.is_authenticated:
             if "sessionid" not in self.session.cookies:
                 self.start_session()
 
             r = self.post(self.session_uri, data=json.dumps(credentials))
+
             if not 200 <= r.status_code < 300:
                 raise AuthenticationFailure()
             else:
+                self.session.headers["X-CSRFToken"] = r.cookies["csrftoken"]
```
johnsonw: Should this also use r.cookies.get(...)?

chroma_cli/api.py
jgrund: Doesn't matter.

chroma_cli/api.py
jgrund: Doesn't matter.

PR #1258, opened 2019-10-14
Wasm bump

jgrund:


johnsonw:
Does the version not change in the cargo.toml file?

jgrund:
Running `cargo update` without changing anything in the cargo.toml will only change unpinned deps.

PR #1265, opened 2019-10-17
Centrally define base packages

### LINKED ISSUE #1264: Updates_ready will always be false when upgrading a monitored filesystem from 4.0.10.x to 5.1
The chroma-agent will poll package information from yum on each of the server nodes and send a blob of data over a session to the manager. This message is picked up and processed in https://github.com/whamcloud/integrated-manager-for-lustre/blob/b5.1/chroma_core/services/lustre_audit/update_scan.py#L108. The problem is that it looks at the packages associated with a profile:
```
for package in self.host.server_profile.serverprofilepackage_set.all():
```

However, there are no packages associated with a monitored profile, which means it will never enter the for loop and thus the package data is never processed. This means that the server nodes will continue to run a previous version of the agent and the "Updates Available" button will remain disabled.

utopiabound:
Define "base" packages in a single location (ServerProfile)
base packages are packages that are ALWAYS installed on a host.

Fixes #1264 

### DIFF
--- a/chroma_core/services/lustre_audit/update_scan.py
+++ b/chroma_core/services/lustre_audit/update_scan.py
@@ -5,6 +5,7 @@
 
 
 import json
+from itertools import chain
 from chroma_core.services import log_register
 
 from django.db import transaction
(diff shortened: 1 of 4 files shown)


chroma_core/services/lustre_audit/update_scan.py
```diff
@@ -106,28 +107,28 @@ def _updates_available(installed_versions, available_versions):
         updates = False
 
         repos = package_report.keys()
-        for package in self.host.server_profile.serverprofilepackage_set.all():
+        for package_name in chain(self.host.server_profile.base_packages, self.host.server_profile.packages):
```
jgrund: Should this be coerced to a `set`?

jgrund:
Looks like SSI testing had an issue deploying the agent

chroma_core/services/lustre_audit/update_scan.py
johnsonw: Makes sense to me, there could be duplicates among the two lists.

Issue #1248, opened 2019-10-11
Add package-upgrade check


jgrund:
Add travis CI test that parses the Requires from the IML spec and ensures there are no newer versions available than what is pinned. Haven't done too much research but you should be able to use rpmspec to read the spec file.

jgrund:
Most of the code needed for this task (reading the repo) can be modeled after this:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/a70212cacaad7e3f42dce6b8f600aa2c11e680da/chroma-agent/chroma_agent/device_plugins/lustre.py#L52-L107

PR #1271, opened 2019-10-21
GH-1248 Add package-upgrade check

johnsonw:
Add travis CI test that parses the Requires from the IML spec and ensures there are no newer versions available than what is pinned. Haven't done too much research but you should be able to use rpmspec to read the spec file.


tests/integration/installation_and_upgrade/test_update_with_yum.py
```diff
@@ -13,6 +13,8 @@ class TestYumUpdate(TestInstallationAndUpgrade):
     def test_yum_update(self):
         """ Test for lustre kernel is set to boot after yum update"""
 
+        from remote_pdb import RemotePdb
```
jgrund: Is this here on purpose?

johnsonw: Yes, I wanted to try and hit this block on the builder.

PR #1282, opened 2019-10-23
Agent check_stonith

utopiabound:
Call to check_stonith() to see if stonith is enabled and configured.
Return true/false plus explanitory text.

Fixes Issue #1191

### DIFF
--- a/iml-agent/src/action_plugins/check_stonith.rs
+++ b/iml-agent/src/action_plugins/check_stonith.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use elementtree::Element;
 use futures::Future;
+use iml_wire_types::{ComponentState, ConfigState, ServiceState};
 use std::process::Command;
 use tracing::{span, Level};
 use tracing_futures::Instrument;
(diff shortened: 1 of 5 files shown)


iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,391 @@
...
+///   - static-list - pcmk_host_list defines comma deliniated list of hosts
+///     controlled by this fencing agent
+///   - dynamic-list (default) - host list is derived from querying fencing
+///     agent (assume same as "none")
+///   - none - any host can fence any host
+/// * pcmk_host_list - list of hosts controlled
+///
+/// Returns: Some("ERROR MESSAGE") on an issue, or None for no issues
+///
+fn inst_attr_problem(elem: &Option<&Element>, node: &String) -> Option<String> {
```
johnsonw: Any reason we can't return a result type instead of an option?

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -0,0 +1,391 @@
...
+    }
+    None
+}
+
+/// This processes <meta_attributes> section of a stonith <primitive>
+/// from the pacemaker cib
+///
+/// Returns: Some("ERROR MESSAGE") on an issue, or None for no issues
+///
+fn meta_attr_problem(elem: &Option<&Element>) -> Option<String> {
```
johnsonw: Any reason we can't return a result type?

Issue #1232, opened 2019-10-08
Modify check_ha to use ComponentState


utopiabound:
* [x] Rename current `check_ha` to `get_ha_resources`
* [x] Create new `check_ha` which returns a tuple of `ComponentState` for `corosync`, `pacemaker`, and `pcsd`
This completes the work on Issue #1224

PR #1288, opened 2019-10-25
check_ha use ComponentState

utopiabound:
check_ha use ComponentState
    
convert current check_ha to get_ha_resource_list()

Issue #1190 #1232

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1288)

### DIFF
--- a/iml-agent/src/action_plugins/check_ha.rs
+++ b/iml-agent/src/action_plugins/check_ha.rs
@@ -6,84 +6,37 @@ use crate::{agent_error::ImlAgentError, systemd};
 use elementtree::Element;
 use futures::try_join;
 use iml_cmd::cmd_output_success;
-use iml_wire_types::{ComponentState, ConfigState, ServiceState};
-use std::{collections::HashMap, fmt, str};
+use iml_wire_types::{
+    ComponentState, ConfigState, ResourceAgentInfo, ResourceAgentType, ServiceState,
+};
+use std::collections::HashMap;
 use tokio::fs::metadata;
 
-/// standard:provider:ocftype (e.g. ocf:heartbeat:ZFS, or stonith:fence_ipmilan)
-#[derive(serde::Deserialize, serde::Serialize, PartialEq, Clone, Debug)]
-pub struct ResourceAgentType {
-    pub standard: String,         // e.g. ocf, lsb, stonith, etc..
-    pub provider: Option<String>, // e.g. heartbeat, lustre, chroma
-    pub ocftype: String,          // e.g. Lustre, ZFS
-}
-
-impl ResourceAgentType {
-    pub fn new(standard: String, provider: Option<String>, ocftype: String) -> Self {
-        ResourceAgentType {
-            standard,
-            provider,
-            ocftype,
-        }
+fn create(elem: &Element, group: Option<String>) -> ResourceAgentInfo {
+    ResourceAgentInfo {
+        agent: ResourceAgentType::new(
+            elem.get_attr("class").unwrap_or("").to_string(),
+            elem.get_attr("provider").map(|s| s.to_string()),
+            elem.get_attr("type").unwrap_or("").to_string(),
+        ),
+        group,
+        id: elem.get_attr("id").unwrap_or("").to_string(),
+        args: match elem.find("instance_attributes") {
+            None => HashMap::new(),
+            Some(e) => e
+                .find_all("nvpair")
+                .map(|nv| {
+                    (
+                        nv.get_attr("name").unwrap_or("").to_string(),
+                        nv.get_attr("value").unwrap_or("").to_string(),
+                    )
+                })
+                .collect(),
+        },
     }
 }
 
-impl fmt::Display for ResourceAgentType {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match &self.provider {
-            Some(provider) => write!(f, "{}:{}:{}", self.standard, provider, self.ocftype),
-            None => write!(f, "{}:{}", self.standard, self.ocftype),
-        }
-    }
-}
-
-impl PartialEq<String> for ResourceAgentType {
-    fn eq(&self, other: &String) -> bool {
-        self.to_string() == *other
-    }
-}
-
-impl PartialEq<&str> for ResourceAgentType {
-    fn eq(&self, other: &&str) -> bool {
-        self.to_string() == *other
-    }
-}
-
-#[derive(serde::Deserialize, serde::Serialize, PartialEq, Clone, Debug)]
-pub struct AgentInfo {
-    pub agent: ResourceAgentType,
-    pub group: Option<String>,
-    pub id: String,
-    pub args: HashMap<String, String>,
-}
-
-impl AgentInfo {
-    pub fn create(elem: &Element, group: Option<String>) -> Self {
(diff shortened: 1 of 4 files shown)


iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +92,208 @@ impl AgentInfo {
...
+
+async fn check_corosync() -> Result<ComponentState<bool>, ImlAgentError> {
+    let mut corosync = ComponentState::<bool> {
+        ..Default::default()
+    };
+
+    if file_exists("/etc/corosync/corosync.conf").await {
+        corosync.config = ConfigState::Unknown;
+        corosync.service = systemd_unit_servicestate("corosync").await?;
+        // @@ check corosync setup
```
johnsonw: Is this a "Todo" and if so can it be done in a separate patch or should it be done in this one?

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +92,208 @@ impl AgentInfo {
...
+}
+
+/// property: org.freedesktop.systemd1
+/// format!("/org/freedesktop/systemd1/unit/{}_2eservice", name)
+/// UnitFileState   +   ActiveState  -> RunState
+/// disabled            inactive        Stopped
+/// enabled             inactive        Enabled
+/// disabled            active          Started
+/// enabled             active          Setup
+async fn systemd_unit_props(name: &str) -> Result<RunState, ImlAgentError> {
```
jgrund: Instead of unwrapping, use the ? operator to return early:

```rust
async fn systemd_unit_props(name: &str) -> Result<RunState, ImlAgentError> {
    let s = format!("/org/freedesktop/systemd1/unit/{}_2eservice", name);
    let output = cmd_output(
        "busctl",
        &[
            "--system",
            "--no-pager",
            "get-property",
            "org.freedesktop.systemd1",
            &s,
            "org.freedesktop.systemd1.Unit",
            "UnitFileState",
            "ActiveState",
        ],
    )
    .compat()
    .await?;

    let res: Vec<&str> = str::from_utf8(&output.stdout)?
        .lines()
        .map(|o| busctl_to_string(o))
        .collect::<Result<Vec<_>, ImlAgentError>>()?;

    match res[..] {
        ["disabled", "inactive"] => Ok(RunState::Stopped),
        ["enabled", "inactive"] => Ok(RunState::Enabled),
        ["disabled", "active"] => Ok(RunState::Started),
        ["enabled", "active"] => Ok(RunState::Setup),
        _ => Err(ImlAgentError::from(RequiredError(format!(
            "Unknown busctl ({}) output: {:?}",
            s, output.stdout
        )))),
    }
}
```

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +92,208 @@ impl AgentInfo {
...
+
+async fn systemd_unit_servicestate(name: &str) -> Result<ServiceState, ImlAgentError> {
+    if systemd_unit_enabled(name).await? {
+        Ok(ServiceState::Configured(systemd_unit_props(name).await?))
+    } else {
+        Ok(ServiceState::Unconfigured)
+    }
+}
+
+async fn file_exists(path: &str) -> bool {
```
jgrund: Why not use: https://docs.rs/tokio/0.1.22/tokio/fs/fn.metadata.html

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +92,208 @@ impl AgentInfo {
...
+    match cmd_output("test", &["-f", path]).compat().await {
+        Err(_) => false,
+        Ok(o) => o.status.success(),
+    }
+}
+
+/// systemctl is-enabled SERVICE :: call: org.freedesktop.systemd1
```
jgrund: Should this comment be here?

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -71,27 +92,208 @@ impl AgentInfo {
...
+        Ok(ServiceState::Unconfigured)
+    }
+}
+
+async fn file_exists(path: &str) -> bool {
+    match cmd_output("test", &["-f", path]).compat().await {
+        Err(_) => false,
+        Ok(o) => o.status.success(),
+    }
+}
```
jgrund: These systemd functions should be in systemd.rs.

iml-agent/src/action_plugins/check_ha.rs
utopiabound: I got errors, something about running in tokio threads. (I don't remember exactly)

jgrund: ah, right, needs to be inside a runtime. Let's wait till 0.2

iml-agent/src/action_plugins/check_ha.rs
jgrund: Looks like this is still a todo?

johnsonw: Just following up on this. I'm assuming there is still some checks that need to be added?

johnsonw:
Still wondering if there are some checks that need to be added based on the comment.

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -2,94 +2,234 @@
...
+
+fn process_resource_list(output: &[u8]) -> Result<Vec<ResourceAgentInfo>, ImlAgentError> {
+    let element = Element::from_reader(output)?;
+
+    Ok(element
+        .find_all("group")
+        .flat_map(|g| {
+            let name = g.get_attr("id").unwrap_or("").to_string();
```
jgrund: Is `id` a required param? Is there a valid case for it to be missing?

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -2,94 +2,234 @@
...
-    pub standard: String,         // e.g. ocf, lsb, stonith, etc..
-    pub provider: Option<String>, // e.g. heartbeat, lustre, chroma
-    pub ocftype: String,          // e.g. Lustre, ZFS
+use iml_wire_types::{
+    ComponentState, ConfigState, ResourceAgentInfo, ResourceAgentType, ServiceState,
+};
+use std::collections::HashMap;
+use tokio::fs::metadata;
+
+fn create(elem: &Element, group: Option<String>) -> ResourceAgentInfo {
```
jgrund: This could be:

`fn create<'a>(elem: &Element, group: impl Into<Option<&'a str>>) -> ResourceAgentInfo`

which would simplify callers.

iml-agent/src/action_plugins/check_ha.rs
utopiabound: `id` is a required parameter.

iml-agent/src/action_plugins/check_ha.rs
utopiabound: Or it's okay as it is...  The callers are not that complicated.

jgrund: Yeah, it works fine without it. I think it makes more sense in the big picture to push the complexity into trait bounds as the code either needs to repeat the same patterns each time or have them handled by the bound.

PR #1296, opened 2019-10-30
Create new Add server CLI

### LINKED ISSUE #1185: Create new Add server CLI
Create a command line interface for adding servers so that it has parity to adding them via the GUI.

jgrund:
Create an add server cli that can deploy n servers in parallel
using hostlist expressions and set their profiles.

This first revision is limited in scope:

  - Only existing keys are supported.
  - Any issue in deployment will abort the command.

Fixes #1185.


jgrund:
add-server

johnsonw:
This looks great!

iml-manager-cli/src/server.rs
```diff
@@ -3,29 +3,155 @@
...
+                    )))
+            })
+            .collect::<Result<Vec<bool>, ImlManagerCliError>>()?;
+
+        if profile_checks.into_iter().all(std::convert::identity) {
+            return Ok(());
+        }
+
+        delay(Instant::now() + Duration::from_millis(500)).await;
+    }
```
jgrund: This fn will poll continually if the profile is truly invalid. 

The easiest thing to do at the moment would be to set a timeout on agent starting, which will exit if there are no updates after n ticks.

Issue #1293, opened 2019-10-29
Add check_kernel agent plugin


utopiabound:
Add ability to pass a list of modules and get back the latest kernel that has all of those modules.

```rust
check_kernel(modules: Vec<String>) -> Result<String, ImlAgentError>
```

This addresses issue like [LU-12902](https://jira.whamcloud.com/browse/LU-12902) where a kernel update breaks part of the KMI that a lustre module needs.

PR #1298, opened 2019-10-30
Find latest kernel that supports listed modules

utopiabound:
Will iterate all installed kernels (found via rpm -q),
and check for listed modules, and will return kernel, that
supports ALL modules.

Fixes Issue #1293


johnsonw:
This looks good overall but it should probably be rebased on the async patch such that it no longer uses futures 0.1 and thus no longer needs the `compat` method.

PR #1305, opened 2019-11-04
Update seed crate from 0.4.0 to 0.4.1

ip1981:
Some functions in "seed", like add_class(), now accept and return mutable
reference.

P. S.
Seed's commit => [Make chainable methods on El and Node more convenient](https://github.com/David-OConnor/seed/commit/095422b4c08a88574c224d827c1801500f09de3c). It says "more convenient", maybe we should make deeper refactoring?

P. P. S. At least now we don't need to do `foo = foo.add_style(...)` for example.

Closes https://github.com/whamcloud/integrated-manager-for-lustre/issues/1302.

### DIFF
--- a/iml-wasm-components/Cargo.lock
+++ b/iml-wasm-components/Cargo.lock
@@ -51,7 +51,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 name = "bootstrap-components"
 version = "0.1.0"
 dependencies = [
- "seed 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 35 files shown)


johnsonw:
In a lot of these changes chaining has been removed. Will it no longer compile when chaining operations on elements?

ip1981:
> In a lot of these changes chaining has been removed. Will it no longer compile when chaining operations on elements?

I think I kept chaining. To my understanding, chaining now operates on mutable references, but still works, like this:

```
let mut foo = ...;
foo.add_class(...).add_attrs(...);
foo
```

Maybe `*(...)` would work, but in most place it was ugly, in the others did not compile and I did not try hard because it was ugly.

PR #1306, opened 2019-11-05
Add PolledMetric type to wire-types

johnsonw:
The PolledMetric type represents polled metric data that will be handled
in the wasm-components filesystem and filesystem detail pages.


jgrund:
Given this is basically all derived from the query (and not really following a set schema), does it make sense to be in wire-types?

PR #1311, opened 2019-11-06
Fixup SVG styles

jgrund:
There are a few SVG styles that no longer work with seed 0.4.1 and cause
a panic at runtime.

They should to be Attrs instead.

Screen Shot 2019-11-06 at 9 53 30 AM


jgrund:
> Should changes to Cargo.lock be part of this commit?

As long as the code compiles / passes through github actions, I think it's ok if we bundle `cargo update` changes in.

PR #1314, opened 2019-11-06
Revert to rust stable

johnsonw:
With the new release of rust we can now compile all rust code using
stable instead of beta. Update all azure templates to use stable.

### DIFF
--- a/.github/actions/copr-rust/action.yml
+++ b/.github/actions/copr-rust/action.yml
@@ -2,4 +2,4 @@ name: copr rust
 description: build in the copr-rust image
 runs:
   using: "docker"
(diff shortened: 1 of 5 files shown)


.github/actions/copr-rust/action.yml
```diff
@@ -2,4 +2,4 @@ name: copr rust
 description: build in the copr-rust image
 runs:
   using: "docker"
-  image: "docker://imlteam/copr-rust:beta"
+  image: "docker://imlteam/copr-rust"
```
jgrund:
```suggestion
  image: "docker://imlteam/copr-rust:stable"
```

PR #1327, opened 2019-11-13
Check stratagem

### LINKED ISSUE #1192: Create `check_stratagem` Action Plugin
Create `check_stratagem` Action Plugin that will check if stratagem is installed or not.

The returned data will indicate whether the required stratagem rpm is currently installed on a node.

mkpankov:
I tested it locally via CLI and manual installation of RPMs and it works, but I want to do 2 more things:
1. Research automated tests for this. It seems we need to mock `rpm` for that. Don't want to get lost in the woods, just consider what is the direction we could take.
2. Account for possible `rpm` errors where it returns non-zero not because package is not found, but RPM DB is locked/corrupted etc.

Closes #1192


iml-agent/src/action_plugins/check_stratagem.rs
```diff
@@ -0,0 +1,10 @@
+use crate::{agent_error::ImlAgentError, cmd::cmd_output};
+
+pub async fn check_stratagem(_: ()) -> Result<bool, ImlAgentError> {
+    let rpm = cmd_output("rpm", vec!["--query", "rust-iml-stratagem"]).await?;
```
jgrund:
```suggestion
    let rpm = cmd_output("rpm", vec!["--query", "lipe"]).await?;
```

fwiw, the actual package name we want to query here is called `lipe`.

jgrund:
Regarding the second point, it would be useful to have a general module / crate that could be used for RPM / yum / dnf based tasks. There are a few crates floating around but none of them looked feature complete last time I checked.

iml-agent/src/action_plugins/check_stratagem.rs
mkpankov: I was under impression it's one of our packages and it's ... not?

mkpankov:
"Upgrade tests" failure seems unrelated?

iml-agent/src/action_plugins/check_stratagem.rs
jgrund: Correct.

jgrund:
> "Upgrade tests" failure seems unrelated?

Yes, needs https://github.com/whamcloud/integrated-manager-for-lustre/issues/1325 at least before they are functional again.

mkpankov:
Okay so apparently there's nothing else we can do with return codes of `rpm`, as 

> > Is there a defined list of the exit codes that rpm will return based on the type of error it had?

> Yes. Unfortunately, it's a very short list: the exit code equals the number
of failed packages, capped at 255.

https://www.redhat.com/archives/rpm-list/2005-July/msg00071.html

So there's not much else we can do w/o parsing the output's text.

Should I start doing it in this PR?

iml-agent/src/action_plugins/check_stratagem.rs
mkpankov: Fixed.

jgrund:
> Should I start doing it in this PR?

Up to you, either this one or a new one that can be stacked.

FWIW some prior art in this area: https://github.com/whamcloud/iml-agent/blob/3b05716b6e7f187faea6e4f5fc51d3d8ca90f6ea/chroma_agent/lib/yum_utils.py#L9-L83

Does yum do any better with reporting error codes?

In addition, we also have a standalone daemon that really should be inside the IML plugin framework and could utilize this lib: 

https://github.com/whamcloud/iml-update-check/

mkpankov:
> Up to you, either this one or a new one that can be stacked.

I think I'm going to work on proper RPM management in a separate PR.

The closest crate there is [librpm](https://lib.rs/crates/librpm). It's far from complete and has versioning [issues](https://github.com/RustRPM/librpm.rs/issues/6) on top of that. I could dive into this problem and fork or write something similar.

> Does yum do any better with reporting error codes?

Yum is worse, it wants to go to repos over the network, and if I disable that, it still doen't tell me if the package is installed. `yum -C info lipe` return `0` both when it's installed and not installed.

mkpankov:
So on my WIP points:
1. Requires something more of a proper interface to RPM instead of shelling out. Postponed until we have that.
2. Also requires a proper interface.

This PR is ready for merge then.

mkpankov:
I've integrated librpm.rs but discovered an [issue](https://github.com/RustRPM/librpm.rs/issues/12) there. Current work is pushed on a [branch](https://github.com/whamcloud/integrated-manager-for-lustre/tree/check_stratagem_rpm).

IMO it's worth investigation but it's going to be pretty hard one, since I'll need to dive into RPM inner workings.

On top of that, there isn't any better error handling in current version of librpm.rs, and to add one is probably as good as reimplement it.

Are we sure we don't want to work on current version first? I think we can get it done - even with output/error parsing - and then advance to proper integration.

jgrund:
> Are we sure we don't want to work on current version first? I think we can get it done - even with output/error parsing - and then advance to proper integration.

That's fine. Given the issues in librpm, do we want to extract the cli calls / parsing to it's own module to get some code reuse?

mkpankov:
> Given the issues in librpm, do we want to extract the cli calls / parsing to it's own module to get some code reuse?

I've extracted a module and updated the PR.

iml-agent/src/rpm.rs
```diff
@@ -0,0 +1,25 @@
+use regex::Regex;
+
+use std::process::Output;
+
+use crate::{agent_error::ImlAgentError, cmd::cmd_output};
+
+fn parse(output: Output) -> Result<bool, ImlAgentError> {
```
johnsonw: Should this be parsing to something instead of returning a result of bool?

mkpankov:
This error doesn't make sense to me
https://github.com/whamcloud/integrated-manager-for-lustre/runs/313329400#step:3:170

I checked locally and our Cargo.lock contains 0.2.55 and iml-gui/crate/Cargo.lock 0.2.53.

RPMs successfully build locally too with `docker run -it -e SPEC="rust-iml.spec" -e LOCAL_ONLY="True" -e CARGO_HOME=/build/.cargo -v $(pwd):/build:rw imlteam/copr-wasm`.

iml-agent/src/rpm.rs
mkpankov: This is what we came up with @jgrund initially, but I can change it.

What do you have in mind? `Result<YumQueryResult, ImlAgentError>` where
```rust
enum YumQueryResult {
  Installed,
  NotInstalled,
}
```
?

mkpankov:
> This error doesn't make sense to me
https://github.com/whamcloud/integrated-manager-for-lustre/runs/313329400#step:3:170
> I checked locally and our Cargo.lock contains 0.2.55 and iml-gui/crate/Cargo.lock 0.2.53.
> RPMs successfully build locally too with docker run -it -e SPEC="rust-iml.spec" -e LOCAL_ONLY="True" -e CARGO_HOME=/build/.cargo -v $(pwd):/build:rw imlteam/copr-wasm.

It passed this time for whatever reason :man_shrugging:

jgrund:
> This error doesn't make sense to me
> https://github.com/whamcloud/integrated-manager-for-lustre/runs/313329400#step:3:170
> 
> I checked locally and our Cargo.lock contains 0.2.55 and iml-gui/crate/Cargo.lock 0.2.53.
> 
> RPMs successfully build locally too with `docker run -it -e SPEC="rust-iml.spec" -e LOCAL_ONLY="True" -e CARGO_HOME=/build/.cargo -v $(pwd):/build:rw imlteam/copr-wasm`.

Update your branch to tip of master, this was fixed yesterday in #1346

iml-agent/src/rpm.rs
jgrund: I don't think we gain anything by using an enum over a bool here.

iml-agent/src/rpm.rs
```diff
@@ -0,0 +1,25 @@
...
+        let s = String::from_utf8(stdout)?;
+        if re.is_match(&s) {
+            Ok(false)
+        } else {
+            Err(ImlAgentError::CmdOutputError(output))
+        }
+    }
+}
+
+pub(crate) async fn query(package_name: &str) -> Result<bool, ImlAgentError> {
```
jgrund: Perhaps change the name of this fn so it's return type is more clear:

```suggestion
pub(crate) async fn installed(package_name: &str) -> Result<bool, ImlAgentError> {
```

iml-agent/src/action_plugins/check_stratagem.rs
```diff
@@ -0,0 +1,5 @@
+use crate::{agent_error::ImlAgentError, rpm};
+
+pub async fn check_stratagem(_: ()) -> Result<bool, ImlAgentError> {
```
jgrund: Will need to be in a differently named file (`rpm_installed.rs`?), but this could be something like;

```rust
pub async fn check_stratagem(package: String) -> Result<bool, ImlAgentError> {
    rpm::query(&package).await
}
```

mkpankov: Done

iml-agent/src/rpm.rs
mkpankov: Done

iml-agent/Cargo.toml
```diff
@@ -40,6 +40,7 @@ parking_lot = "0.9"
 iml-wire-types = { path = "../iml-wire-types", version = "0.2" }
 liblustreapi = { path = "../liblustreapi", version = "0.1" }
 iml-fs = { path = "../iml-fs", version = "0.1.0" }
+regex = "1.3.1"
```
jgrund: We can shrink the binary size a bit by not including all features:

```
[dependencies.regex]
version = "1.3"
default-features = false
features = ["std"]
```

Looking at the usage, I think this should be ok.

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -89,6 +89,11 @@ pub fn create_registry() -> HashMap<ActionName, Callback> {
         mk_callback(check_stonith::check_stonith),
     );
 
+    map.insert(
+        "action_check_stratagem".into(),
```
jgrund:
```suggestion
        "package_installed".into(),
```

mkpankov: Done

iml-agent/Cargo.toml
mkpankov: Done

PR #1328, opened 2019-11-14
Add new smart input UI component

ip1981:
It parses input and stores the result of successful parsing or displays an error in a tooltip.

Currently parsing is performed instantly on any change of input. Probably this behavior can be improved, but we'll need more actual use cases.

Also showing the exact position of an error is tricky. Some options are: "contenteditable" elements with richtext capabilities, shown an excerpt in the tooltip itself. 

Example:
sample

This is what I used for testing the example:

```diff
diff --git a/iml-wasm-components/iml-fs/src/fs_page.rs b/iml-wasm-components/iml-fs/src/fs_page.rs
index 0ac8e9140..4aefe9442 100644
--- a/iml-wasm-components/iml-fs/src/fs_page.rs
+++ b/iml-wasm-components/iml-fs/src/fs_page.rs
@@ -11,11 +11,13 @@ use iml_environment::ui_root;
 use iml_lock_indicator::{lock_indicator, LockIndicatorState};
 use iml_utils::{Locks, WatchState};
 use iml_wire_types::{Alert, Filesystem, Target, TargetConfParam, ToCompositeId};
-use seed::{a, attrs, class, div, h1, h4, i, prelude::*, table, tbody, td, th, thead, tr};
+use seed::{a, attrs, class, div, h1, h4, i, li, ol, prelude::*, table, tbody, td, th, thead, tr};
 use std::collections::{HashMap, HashSet};
 use wasm_bindgen::JsValue;
 use web_sys::Element;
 
+use std::collections::BTreeSet;
+
 cfg_if! {
     if #[cfg(feature = "console_log")] {
         fn init_log() {
@@ -46,6 +48,7 @@ struct Model {
     pub rows: Rows,
     pub targets: HashMap<u32, Target<TargetConfParam>>,
     pub locks: Locks,
+    pub input: iml_input::Model<BTreeSet<String>>,
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
@@ -80,10 +83,14 @@ enum Msg {
     Alerts(Vec<Alert>),
     SetLocks(Locks),
     ActionDropdown(dad::IdMsg<Filesystem>),
+    Input(iml_input::Msg),
 }
 
 fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
     match msg {
+        Msg::Input(i) => {
+            iml_input::update::<BTreeSet<String>>(i, &mut model.input, hostlist_parser::parse);
+        }
         Msg::Destroy => {
             model.destroyed = true;
 
@@ -276,6 +283,16 @@ fn view(model: &Model) -> Node<Msg> {
         if model.has_fs() {
             div![
                 h4![class!["section-header"], "File Systems"],
+                iml_input::view(&model.input).map_message(Msg::Input),
+                if let Ok(l) = &model.input.value {
+                    let mut ol = ol![];
+                    for e in l {
+                        ol.add_child(li![e]);
+                    }
+                    ol
+                } else {
+                    seed::empty![]
+                },
                 table![
                     class!["table"],
                     thead![tr![
```

Closes https://github.com/whamcloud/integrated-manager-for-lustre/issues/1309.


iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,123 @@
+#[macro_use]
```
johnsonw: Shouldn't need this: https://doc.rust-lang.org/edition-guide/rust-2018/macros/macro-changes.html

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+                    format!("{}", t)
+                }
+            }
+            Info::Static(s) => String::from(*s),
+            Info::Owned(s) => String::from(s),
+            Info::Range(r) => String::from(*r),
+        }
+    }
+
+    for x in &err.errors {
```
johnsonw: Looping like this works fine. As an alternative, you can eliminate the mutable variables `un` and `ex` above by folding:
```
let (un, ex) = &err
        .errors
        .iter()
        .fold((vec![], vec![]), |(mut un, mut ex), x| match x {
            Error::Unexpected(i) => {
                un.push(format_info(&i));
                (un, ex)
            }
            Error::Expected(i) => {
                ex.push(format_info(&i));
                (un, ex)
            }
            _ => (un, ex),
        });
```

ip1981: Yeah, but it looks less readable :)

johnsonw: To be clear, i'm ok with the for loop. I shared the fold as an alternative because it reduces mutable state within the scope. Personally I find the fold a bit easier to reason about because I don't have to think about the possibility of `un` and `ex` changing within the scope of the function since they are immutable.

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,123 @@
+#[macro_use]
+extern crate seed;
+
+extern crate combine;
+
+use combine::stream::easy::Error;
+use combine::stream::easy::Errors;
+use combine::stream::easy::Info;
+use seed::prelude::*;
```
johnsonw:
```suggestion
use seed::{div, input, prelude::*};
```

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,121 @@
+extern crate seed;
+
+extern crate combine;
+
+use combine::stream::easy::Error;
+use combine::stream::easy::Errors;
+use combine::stream::easy::Info;
```
ip1981: Maybe squash them too?

ip1981:
Wow, I can simply write `iml_input::update(i, &mut model.input, hostlist_parser::parse);` and the compiler will understand.

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,118 @@
+extern crate combine;
+extern crate seed;
```
jgrund: extern crate is not needed.

```suggestion
```

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,118 @@
+extern crate combine;
```
jgrund: extern crate is not needed.

```suggestion
```

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+pub struct Model<T> {
+    pub value: Result<T, String>,
+    pub tooltip: iml_tooltip::Model,
+}
+
+impl<T> Default for Model<T> {
+    fn default() -> Self {
+        Self {
+            value: Err(String::new()),
```
jgrund: Do we want the default state to be an empty error?

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+                }
+            }
+            Info::Static(s) => String::from(*s),
+            Info::Owned(s) => String::from(s),
+            Info::Range(r) => String::from(*r),
+        }
+    }
+
+    for x in &err.errors {
+        match x {
```
jgrund: This is not totally covering the error types. Do we want to do something with those?

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+pub fn update<T>(msg: Msg, m: &mut Model<T>, parse: ParserFn<T>) {
+    match msg {
+        Msg::FlipTooltip => {
+            m.tooltip.placement = match m.tooltip.placement {
+                TooltipPlacement::Bottom => TooltipPlacement::Top,
+                _ => TooltipPlacement::Bottom,
+            }
+        }
+        Msg::InputChanged(txt) => match parse(&txt) {
```
jgrund: nit, could also be:

`m.value = parse(&txt).map_err(|e| format_err(&e));`

Since the Ok branch is a noop.

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+                m.value = Err(format_err(&e));
+            }
+        },
+    }
+}
+
+pub fn view<T>(m: &Model<T>) -> Node<Msg> {
+    let mut input = input![];
+    input.add_listener(input_ev(Ev::Input, Msg::InputChanged));
+    input.add_class("form-control");
```
jgrund: nit, we should prefer the macro style for building the view when we can use it.

This can be:

```rust
let input = input![
        class!["form-control"],
        input_ev(Ev::Input, Msg::InputChanged)
    ];
```

iml-wasm-components/iml-input/src/lib.rs
```diff
@@ -0,0 +1,118 @@
...
+
+    let mut group = div![input];
+    group.add_class("input-group");
```
jgrund: Same as above, this can be:

```rust
div![
input,
class!["input-group"]
];
```

iml-wasm-components/iml-input/src/lib.rs
ip1981: The other two is a "generic" message and "other, but Combine [never generates them](https://docs.rs/combine/4.0.0-beta.1/combine/stream/easy/struct.Errors.html), and their use is not clear to decide how to show them.

iml-wasm-components/iml-input/src/lib.rs
ip1981: I think it could be the default of underlying type.

jgrund: I suppose this means the component starts in the Error state?

ip1981: Anyway, `Ok` does not seem right, because it is not ok - we don't know yet the result. So `Err` looks the best. Then, I don't want to show  a user an error tooltip while they hasn't typed anything yet. So empty string indicates no user error, but still invalid state of input. I guess I can user `Option` instead.

PR #1336, opened 2019-11-18
Add tokio_utils

jgrund:
Add a new module, tokio_utils to the iml-util crate.

Within it, add three new fns to deal with incoming socket streams of
different types.

We should use these fns when we don't know which type of stream we will
be using at runtime.

### DIFF
--- a/iml-util/util.rs
+++ b/iml-util/util.rs
@@ -56,6 +56,76 @@ impl<A, B, C, D, E> Pick<A, B> for (A, B, C, D, E) {
     }
 }
 
+pub mod tokio_utils {
+    use futures::{future::Either, Stream, TryStreamExt};
+    use std::{
+        convert::TryFrom,
+        env, io,
+        os::unix::{io::FromRawFd, net::UnixListener as NetUnixListener},
+        pin::Pin,
+    };
+    use tokio::{
+        io::{AsyncRead, AsyncWrite},
+        net::{TcpListener, UnixListener},
+    };
+
+    pub trait Socket: AsyncRead + AsyncWrite + Send + 'static + Unpin {}
+    impl<T: AsyncRead + AsyncWrite + Send + 'static + Unpin> Socket for T {}
+
+    /// Given an environment variable that resolves to a port,
+    /// Return a stream containing `TcpStream`s that have been erased to
+    /// `AsyncRead` + `AsyncWrite` traits. This is useful when you won't know which stream
+    /// to choose at runtime
+    pub async fn get_tcp_stream(
+        port: String,
+    ) -> Result<impl Stream<Item = Result<Pin<Box<dyn Socket>>, io::Error>>, io::Error> {
+        let addr = format!("127.0.0.1:{}", port);
+
+        tracing::debug!("Listening over tcp port {}", port);
+
+        let s = TcpListener::bind(&addr)
+            .await?
+            .incoming()
+            .map_ok(|x| -> Pin<Box<dyn Socket>> { Box::pin(x) });
+
+        Ok(s)
+    }
+
+    /// Returns a stream containing `UnixStream`s that have been erased to
+    /// `AsyncRead` + `AsyncWrite` traits. This is useful when you won't know which stream
+    /// to choose at runtime
+    pub fn get_unix_stream(
+    ) -> Result<impl Stream<Item = Result<Pin<Box<dyn Socket>>, io::Error>>, io::Error> {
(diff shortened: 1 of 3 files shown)


johnsonw:
I was using `env::var("LISTEN_PID")` in my patch. Should I now do:
```
get_tcp_or_unix_listener("LISTEN_PID")
```
or should I use a different environment variable?

jgrund:
It should now be: `let listener = get_tcp_or_unix_listener("ACTION_RUNNER_PORT").await?;`

iml-util/util.rs
```diff
@@ -56,6 +56,76 @@ impl<A, B, C, D, E> Pick<A, B> for (A, B, C, D, E) {
...
+
+        Ok(s)
+    }
+
+    /// Returns a stream containing `UnixStream`s that have been erased to
+    /// `AsyncRead` + `AsyncWrite` traits. This is useful when you won't know which stream
+    /// to choose at runtime
+    pub fn get_unix_stream(
+    ) -> Result<impl Stream<Item = Result<Pin<Box<dyn Socket>>, io::Error>>, io::Error> {
+        let addr = unsafe { NetUnixListener::from_raw_fd(3) };
```
ip1981: 3 is a kind of magick number? I know that 0 is stdin, 1 is stdout, 2 is stderr.

iml-util/util.rs
```diff
@@ -56,6 +56,76 @@ impl<A, B, C, D, E> Pick<A, B> for (A, B, C, D, E) {
...
+
+        tracing::debug!("Listening over unix domain socket");
```
ip1981: Just a question: what is opening this unix socket? It must be a file somewhere?

jgrund: This is opened when we use as a systemd service with socket-activation. ex: 

https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-action-runner.socket

and

https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-action-runner.service

iml-util/util.rs
jgrund: yeah, pretty much. the service manager passes in the descriptors starting at 3 and going from there. Ref: https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html#

ip1981: Maybe systemd passes some environment variables, where we can find explicit descriptor?

jgrund: Here's the environment variables for the .service:

```
[root@adm ~]# cat /proc/14470/environ
LANG=en_US.UTF-8PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/binLISTEN_PID=14470LISTEN_FDS=1RUST_LOG=infoREALTIME_PORT=8888VIEW_SERVER_PORT=8889WARP_DRIVE_PORT=8890MAILBOX_PORT=8891DEVICE_AGGREGATOR_PORT=8008HTTP_AGENT2_PORT=8003HTTP_AGENT2_PROXY_PASS=http://127.0.0.1:8003ALLOW_ANONYMOUS_READ=trueBUILD=IS_RELEASE=trueLOG_PATH=/var/log/chromaSERVER_HTTP_URL=https://adm.local:443SITE_ROOT=/usr/share/chroma-managerVERSION=5.1.0-1API_USER=apiAPI_KEY=9b2b11b0a2ded142322340e20df611f762545a93SOURCE_MAP_PATH=/usr/share/iml-manager/iml-gui/main.5b0746a778967e0af2e7.css.mapMAILBOX_PATH=/var/spool/iml/mailboxPROXY_HOST=127.0.0.1DB_HOST=DB_NAME=chromaDB_USER=chromaDB_PASSWORD=AMQP_BROKER_USER=chromaAMQP_BROKER_PASSWORD=chroma123AMQP_BROKER_VHOST=chromavhostAMQP_BROKER_HOST=localhostAMQP_BROKER_PORT=5672AMQP_BROKER_URL=amqp://chroma:chroma123@localhost:5672/chromavhost
```

Unfortunately I don't see anything explictly referencing fd 3.

jgrund: I think the closest thing to what you're looking for is this:

https://docs.rs/libsystemd-sys/0.2.2/libsystemd_sys/daemon/fn.sd_is_socket_unix.html

However, that will involve pulling in a dep and making an unsafe call each time the service is started.

Just to note, I haven't seen any issues directly relying on fd 3 for our socket activated services.

PR #1337, opened 2019-11-19
Add breadcrumbs

ip1981:
https://github.com/whamcloud/integrated-manager-for-lustre/issues/1331.

**Normal**
normal

**Overflow**
overflow

### DIFF
--- a/iml-gui/crate/src/components/breadcrumbs.rs
+++ b/iml-gui/crate/src/components/breadcrumbs.rs
@@ -1,0 +1,106 @@
+use crate::{generated::css_classes::C, Msg, Page};
+use seed::{prelude::*, *};
+use std::cmp::PartialEq;
+use std::collections::LinkedList;
+
+pub struct BreadCrumbs<Crumb> {
+    crumbs: LinkedList<Crumb>,
+}
+
+impl<Crumb> Default for BreadCrumbs<Crumb> {
+    fn default() -> Self {
+        Self {
+            crumbs: LinkedList::new(),
+        }
+    }
+}
+
+impl<Crumb: PartialEq> BreadCrumbs<Crumb> {
+    pub fn iter(self: &Self) -> impl DoubleEndedIterator<Item = &Crumb> {
+        self.crumbs.iter()
+    }
+
+    pub fn clear(self: &mut Self) -> &mut Self {
+        self.crumbs.clear();
+        self
+    }
+
+    pub fn push(self: &mut Self, n: Crumb) -> &mut Self {
+        let mut new_crumbs = LinkedList::new();
+
+        while let Some(c) = self.crumbs.pop_front() {
+            if c != n {
+                new_crumbs.push_back(c);
+            } else {
+                break;
+            }
+        }
+
+        new_crumbs.push_back(n);
+        self.crumbs = new_crumbs;
+        self
+    }
+}
+
+pub fn view(bc: &BreadCrumbs<Page>) -> impl View<Msg> {
+    let mut ol = ol![class![C.justify_center, C.truncate, C.mx_4, C.rtl]];
+
+    // XXX the list has "direction: rtl" to put ellipsis to the left on overflow,
+    // XXX need to reverse the crumbs to show them in the correct left-to-right order:
+    for (n, p) in bc.iter().rev().enumerate() {
+        let mut cr = li![
+            class![C.inline],
+            a![
+                class![C.hover__underline],
+                attrs! {At::Href => p.to_href()},
+                p.to_string(),
+            ]
+        ];
+        if n == 0 {
+            cr.add_class(C.font_bold);
+        } else {
+            ol.add_child(i!["/", class![C.px_2]]);
+        }
+        ol.add_child(cr);
+    }
+    ol
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn one_crumb() {
+        let mut m = BreadCrumbs::default();
+        m.push("foo");
+        assert_eq!(m.iter().count(), 1);
+    }
+
+    #[test]
+    fn two_different_crumbs() {
+        let mut m = BreadCrumbs::default();
+        m.push("foo");
+        m.push("bar");
+        assert_eq!(m.iter().count(), 2);
+    }
+
+    #[test]
+    fn two_identical_crumbs() {
+        let mut m = BreadCrumbs::default();
+        m.push("xxx");
+        m.push("xxx");
+        assert_eq!(m.iter().count(), 1);
+    }
+
+    #[test]
+    fn cycle() {
+        let mut m = BreadCrumbs::default();
+        m.push("aaa");
+        m.push("bbb");
+        m.push("ccc");
+        m.push("ddd");
+        m.push("ccc");
+        assert_eq!(m.iter().count(), 3);
+    }
+}
--- a/iml-gui/crate/src/components/mod.rs
+++ b/iml-gui/crate/src/components/mod.rs
@@ -1,5 +1,6 @@
 pub(crate) mod activity_indicator;
 pub(crate) mod arrow;
+pub(crate) mod breadcrumbs;
 pub(crate) mod font_awesome;
 pub(crate) mod popover;
 pub(crate) mod tooltip;
--- a/iml-gui/crate/src/page/partial/header.rs
+++ b/iml-gui/crate/src/page/partial/header.rs
@@ -1,5 +1,5 @@
 use crate::{
-    components::{activity_indicator, font_awesome},
+    components::{activity_indicator, breadcrumbs, font_awesome},
     generated::css_classes::C,
     Model, Msg, Route,
     Visibility::*,
@@ -312,7 +312,7 @@ pub fn view(model: &Model) -> impl View<Msg> {
                 C.text_center,
                 C.py_2
             ],
-            model.route.to_string()
+            breadcrumbs::view(&model.breadcrumbs).els()
         ],
     ]
 }
--- a/iml-gui/css/custom.css
+++ b/iml-gui/css/custom.css
@@ -49,3 +49,16 @@
     transform: rotate(360deg);
   }
 }
+
+.rtl {
+  direction: rtl;
+}
+
+/*
+ * This is standard conformant and default in Firefox, but not in Chrome.
+ * Set it explicitly to mitigate possible difference.
+ */
+li {
+  unicode-bidi: isolate;
+}
+
--- a/iml-gui/crate/src/generated/css_classes.rs
+++ b/iml-gui/crate/src/generated/css_classes.rs
@@ -12170,6 +12170,11 @@ pub struct CssClasses<'a> {
     pub rotate: &'a str,
 
     /**
+        direction: rtl;
+    */
+    pub rtl: &'a str,
+
+    /**
         position: absolute;    @media (min-width: 569px)
         width: 1px;    @media (min-width: 569px)
         height: 1px;    @media (min-width: 569px)
@@ -66580,6 +66585,11 @@ pub static C: CssClasses = CssClasses {
         animation: rotate-center 1s linear infinite both;
     */
     rotate: "rotate",
+
+    /**
+        direction: rtl;
+    */
+    rtl: "rtl",
 
     /**
         position: absolute;    @media (min-width: 569px)
--- a/iml-gui/crate/src/lib.rs
+++ b/iml-gui/crate/src/lib.rs
@@ -10,7 +10,9 @@ mod generated;
 mod page;
 mod route;
 
-use components::{update_activity_health, ActivityHealth};
+use components::{
+    breadcrumbs::BreadCrumbs, update_activity_health, ActivityHealth,
+};
 use generated::css_classes::C;
 use iml_wire_types::warp_drive;
 use js_sys::Function;
@@ -100,6 +102,7 @@ pub struct Model {
     pub records: warp_drive::Cache,
     pub locks: warp_drive::Locks,
     pub activity_health: ActivityHealth,
+    pub breadcrumbs: BreadCrumbs<Page>,
 }
 
 pub fn register_eventsource_handle<T, F>(
@@ -169,6 +172,7 @@ fn after_mount(url: Url, orders: &mut impl Orders<Msg>) -> AfterMount<Model> {
         records: warp_drive::Cache::default(),
         locks: HashMap::new(),
         activity_health: ActivityHealth::new(),
+        breadcrumbs: BreadCrumbs::default(),
     })
 }
 
@@ -219,6 +223,10 @@ pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
         Msg::RouteChanged(url) => {
             model.route = url.into();
             orders.send_msg(Msg::UpdatePageTitle);
+            if model.page == Page::Home {
+                model.breadcrumbs.clear();
+            }
+            model.breadcrumbs.push(model.page);
         }
         Msg::UpdatePageTitle => {
             let title =


iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,103 @@
...
+    dom_types::{At, El, Node, UpdateEl, View},
+    li, ol,
+    prelude::IndexMap,
+};
+
+use std::cmp::PartialEq;
+use std::collections::LinkedList;
+
+pub struct Model<Crumb> {
+    crumbs: LinkedList<Crumb>,
```
jgrund: nice

iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+impl<Crumb> Default for Model<Crumb> {
```
jgrund: Could this not be derived?

iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,103 @@
...
+    a, attrs, div,
+    dom_types::{At, El, Node, UpdateEl, View},
+    li, ol,
+    prelude::IndexMap,
+};
+
+use std::cmp::PartialEq;
+use std::collections::LinkedList;
+
+pub struct Model<Crumb> {
```
jgrund: I'm not convinced that we need a separate model for this and it couldn't live in the top-level model in lib. The messages are already on the top-level as well.

jgrund:
Breadcrumbs should be rendered here:

Screen Shot 2019-11-19 at 3 24 23 PM

iml-gui/crate/src/components/breadcrumbs.rs
ip1981: I think it can. Probably it's left from previous ideas.

iml-gui/crate/src/components/breadcrumbs.rs
ip1981: I think we would need  a wrapping type (struct) anyway. We could call it something different than Model.
The Msg is only needed for the view, but  AFAIK it needs to be defined next to the "model", as well as other methods (goto, goback). Need to think how not to put everything in one top-level module and at the same time not make it overly complicated :)

ip1981: Maybe use a trait instead of a struct.

jgrund: What we ultimately have in the model is a single top-level state atom. I suppose it's really subjective whether we use a separate discrete struct as a model or not, since in the end it serves the same purpose. As long as we keep the messages in the top level, it should work ok.

iml-gui/crate/src/components/breadcrumbs.rs
jgrund: Does it make sense for the Model to be generic at this point? We only have a single contained type, correct? Should we wait for some variant usage before making it generic?

ip1981: I was trying to isolate the logic of breadcrumbs. Not to try to make it generic, but to write minimum code require for it. And I think I have succeeded with lats update. Being generic is a side effect, no more :)

ip1981: Also at the beginning I wasn't sure what a crumb would be. I started with a string, but had in mind some struct. Then I found that the Page enum was also a good choice. It may change in the future, but it does no matter with generic approach.

jgrund: Ok, fair enough :)

iml-gui/crate/src/components/mod.rs
```diff
@@ -1,3 +1,4 @@
+pub(crate) mod breadcrumbs;
```
ip1981: Since it is not using anything from Seed, I am not sure now it here is a good place for it.

iml-gui/css/custom.css
```diff
@@ -49,3 +49,19 @@
     transform: rotate(360deg);
   }
 }
+
+.breadcrumbs {
```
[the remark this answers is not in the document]
ip1981: I thought Tailwind was encouraging adding custom classes. But anyway, here I used only Tailwind bits, and this look-and-feel in much harder (if possible) to do in Rust/HTML.

ip1981:
> A few comments.
> 
> Thinking about this a bit more, It seems we need a stack that contains a stack in each slot.
> 
> The behavior could be something like:
> 
>     1. We start with an empty parent stack. When we get a new route we push a new child stack into the parent.
> 
>     2. Each navigation that doesn't reset the breadcrumbs pushes onto the child stack.
> 
>     3. A navigation that resets the breadcrumbs pushes a new child stack on the parent.
> 
>     4. Hitting the back button pops from the child stack. Iff we are at the bottom of the stack, hitting the back button moves to the previous stack in the parent stack.
> 
> 
> This assumes that each link the UI has a way to indicate that clicking it resets the breadcrumbs (creates a new child stack). I think the most straightforward way to do this is to append a param (`#reset`?) to each link that should reset the breadcrumbs when clicked.

I understand it technically, but not from the user's perspective.
Do you have in mind any existing site which implements it? It is somehow similar to [location breadcrumbs](https://en.wikipedia.org/wiki/Breadcrumb_navigation#Types)? Child stacks are meant for submenus? Which pages should reset breadcrumbs?

And the problem with the linear browser history remains: a user can go back to a page belonging to a substack, while breadcrumbs will switch to the previous substack. I can't tell more without a more detailed picture.

Also spoiling URL with some parameters just for the sake of  a secondary feature suggests me, that it might not be the right thing to do :)

Currently, resetting breadcrumbs is not even necessary: a user starts at some page anyway, and if they get back to this page, the breadcumbs are effectively reset to that page.

ip1981:
BTW, a cycle in breadcrumbs technically resets it, but not completely.

iml-gui/crate/src/page/partial/header.rs
```diff
@@ -235,7 +235,13 @@ pub fn view(model: &Model) -> impl View<Msg> {
                 C.text_center,
                 C.py_2
             ],
-            model.page.to_string()
+            ol![
```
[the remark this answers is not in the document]
ip1981: I wasn't sure. This is so simple, de facto a singleton, does not use its own model. That's why I questioned the location of `breadcrumbs.rs` itself.

jgrund:
> I understand it technically, but not from the user's perspective.
Do you have in mind any existing site which implements it? It is somehow similar to location breadcrumbs? Child stacks are meant for submenus? Which pages should reset breadcrumbs?

It's probably closer to path breadcrumbs. The current IML GUI *mostly* implements it, but breaks when clicking the back button due to it's use of a single stack.

Here's an example:

breadcrumbs

1. The user starts on the dashboard page.
1. The user clicks the servers link in the sidebar. This has no relevance to the current page, so clicking it resets the breadcrumb stack.
1. The user clicks mds1.local within the server page. This *does* have relevance to the current page, so clicking it pushes on the breadcrumb stack.
1. The user clicks the jobstats link in the menu bar. This has no relevance to the current page, so clicking it resets the breadcrumb stack.
1. The user clicks the back button. This is where the current GUI breadcrumb implementation is broken as it uses a single stack. It now shows "Job Stats / Server Detail" which is not correct. It should instead show the previous stack of "Server / Server Detail" as that was the navigation path at that point.

By using child stacks, we can restore previous paths correctly, when we are at the bottom of an existing stack and hit the back button.

jgrund:
> Also spoiling URL with some parameters just for the sake of a secondary feature suggests me, that it might not be the right thing to do :)

Yes. The problem is we need to be able to replay navigation in the forwards direction after page refresh as well. To do this, we need to store the state of resetting the page *somewhere* that persists between refreshes. I'm open for suggestions, but using a query param seems like the simplest way to do this.

jgrund:
> BTW, a cycle in breadcrumbs technically resets it, but not completely.
> 
> Should it not? I. e.:
> 
> `Page1 / YOUAREHERE / AlsoWereHere`
> 
> And if the user goes somewhere else than `AlsoWereHere`, we reset the crumbs right after `YOUAREHERE`?

Depends really, I think the key to this is that the links themselves are in control of whether clicking them resets the stack or not. It's really not possible to tell what should happen otherwise. For example, clicking an item in the sidebar or in the menu is really a new breadcrumb trail and not the continuation of the current one.

ip1981:
So it seems like instead of a single "global" breadcrumbs, you talking about a few "local"? There are a couple of ides that come to my mind, e. g. group by URL prefix (`/server/...`, `/configuration/...`), or reset breadcrumbs when a menu item clicked (instead of a link on the page), or explicit grouping pages and having separate breadcrumbs for each group. The latter is probably the best, but I'd lie to see more pages :) For example, currently we have a flat enum for pages, what it will look like when we add pages for viewing servers and FSs?

Anyway, It is really important for users that pages were relevant to each other? When implementing this breadcrumbs I was thinking about the unified/flat/global breadcrumbs which would help users navigate: `Dashboard / Servers/ mds1.local / About / Logs / mds2.local`. This is extremely simple and understandable.

jgrund:
> Anyway, It is really important for users that pages were relevant to each other? When implementing this breadcrumbs I was thinking about the unified/flat/global breadcrumbs which would help users navigate: Dashboard / Servers/ mds1.local / About / Logs / mds2.local. This is extremely simple and understandable.

Yeah, I think we'd be ok moving to this approach. 

A few things to consider here: 

1. to match the design mocks, the breadcrumbs should be centered.
1. With a static breadcrumb list, the list has the potential to be much longer. How should we handle the case of longer breadcrumb trails?

ip1981:
Centering is not a problem :)

Longer lists, I think, could be put in a container with some overflow property (CSS).

And one new thing: do you think is it a good idea to keep visited pages as well, until new path diverse significantly? I. e. keeping `AlsoWereHere` in `Page1 / YOUAREHERE / AlsoWereHere` until the user goes from `YOUAREHERE` to a page other than `AlsoWereHere`?

ip1981:
Also what do you think about putting `/` before the first crumb too? :)

jgrund:
If we are going with a static breadcrumb listing, then the only thing that should truncate the list is clicking on one of the breadcrumb items themselves (which should reset the trail to that point).

jgrund:
Leave off the leading slash so it matches intended design

ip1981:
> If we are going with a static breadcrumb listing, then the only thing that should truncate the list is clicking on one of the breadcrumb items themselves (which should reset the trail to that point).

Yes, that is what happens.
Removing all crumbs still makes sense if open home page.
And there is no problem at all with the Back button: cumbs work not matter how user gets to a page.

In the future, if you want multiple breadcrumbs, we can have them by grouping pages into categories and switching breadcrumbs for a category before updating. This would be a you have few open tabs with separate crumbs in each.

ip1981:
Something like this. There are some CSS tricks, I think I can do it otherwise:
bc

The scrollbar is worrying me.

ip1981:
Ready! I had to add one custom class anyway (with `direction: rtl`). The screen getting wider and a horizontal scrollbar is because of body's `width: 100vw`. Maybe we should set it 100%.

ip1981:
Maybe the NotFound page will need special treatment :)

iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    #[test]
+    fn cycle() {
+        let mut m = BreadCrumbs::default();
+        m.push("aaa");
+        m.push("bbb");
+        m.push("ccc");
+        m.push("ddd");
+        m.push("ccc");
+        assert_eq!(m.iter().count(), 3);
```
nlinker: Nice

iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,108 @@
...
+        self.crumbs = new_crumbs;
+        self
+    }
+}
+
+pub fn view(bc: &BreadCrumbs<Page>) -> impl View<Msg> {
+    let mut ol = ol![class![C.justify_center, C.truncate, C.mx_4, C.rtl]];
+    let mut it = bc.iter().rev().enumerate(); // XXX rev() because of "direction: rtl;"
+
+    while let Some((n, p)) = it.next() {
```
jgrund: nit: this is a bit more succinct by just consuming the enumerated iterator directly:

`for (n, p) in bc.iter().enumerate() {`

iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        new_crumbs.push_back(n);
+        self.crumbs = new_crumbs;
+        self
+    }
+}
+
+pub fn view(bc: &BreadCrumbs<Page>) -> impl View<Msg> {
+    let mut ol = ol![class![C.justify_center, C.truncate, C.mx_4, C.rtl]];
+    let mut it = bc.iter().rev().enumerate(); // XXX rev() because of "direction: rtl;"
```
jgrund: The direction of crumbs should be ltr.

ip1981: But then we cannot put ellipsis on the left. The visible order of crumbs is correct, btw.

ip1981: I use rlt to pus ellipsis on the left, and then reversed crumbs to show them in correct ltr order. This appears to be the only way.

iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,106 @@
...
+        self
+    }
+}
+
+pub fn view(bc: &BreadCrumbs<Page>) -> impl View<Msg> {
+    let mut ol = ol![class![C.justify_center, C.truncate, C.mx_4, C.rtl]];
+
+    // XXX the list has "direction: rtl" to put ellipsis to the left on overflow,
+    // XXX need to reverse the crumbs to show them in the correct left-to-right order:
+    for (n, p) in bc.iter().rev().enumerate() {
```
nlinker: The comment prefix `XXX` means something, correct? For me it looks unnecessary.

iml-gui/crate/src/components/breadcrumbs.rs
```diff
@@ -0,0 +1,106 @@
...
+        self.crumbs = new_crumbs;
+        self
+    }
+}
+
+pub fn view(bc: &BreadCrumbs<Page>) -> impl View<Msg> {
+    let mut ol = ol![class![C.justify_center, C.truncate, C.mx_4, C.rtl]];
+
+    // XXX the list has "direction: rtl" to put ellipsis to the left on overflow,
+    // XXX need to reverse the crumbs to show them in the correct left-to-right order:
```
nlinker: The comment prefix `XXX` means something, correct? For me it looks unnecessary.

ip1981: It is a placeholder for "WTF" :)

> Use XXX in a comment to flag something that is bogus but works. Use FIXME to flag something that is bogus and broken.

[[1]](https://www.oracle.com/technetwork/java/javase/documentation/codeconventions-137265.html#395)

ip1981:
I just added 

```css
li {
  unicode-bidi: isolate;
}
```

which is right standards wise, eliminates difference (existing or future) between engines, and helps keep code simple.

ip1981:
https://bugs.chromium.org/p/chromium/issues/detail?id=296863

jgrund:
The breadcrumb trail doesn't seem to restore the same way when the back button is clicked:

breadcrumb-back

Given a starting path of 'Activity / Logs / Jobstats' , then clicking 'Activity', then clicking the back button,  I would expect the previous path to be restored, but instead it puts 'Activity / Jobstats'.

ip1981:
> The breadcrumb trail doesn't seem to restore the same way when the back button is clicked:
> 
> breadcrumb-back
> 
> Given a starting path of 'Activity / Logs / Jobstats' , then clicking 'Activity', then clicking the back button, I would expect the previous path to be restored, but instead it puts 'Activity / Jobstats'.

This is kind of by design. Breadcrumbs just show the path user went to the current page, and does not distinguish _how_. Clicking the Back button acts as if user clicked on page link. In your example, you get to  page Jobstats from  page Activity by clicking the Back button, same as if you clicked on Jobstats from the menu. In analogy, once you get back to the initial page (or a page where you have already been), you start a new path with new breadcrumb.

But I think this can be improved... :)

PR #1344, opened 2019-11-20
Add font-awesome componet

jgrund:
Add a new fn that can render a given font-awesome icon.

This fn will use the svg sprite sheet for rendering rather then CSS.

In addition, any attributes (classes, styles, etc...) can be passed in
as a param.


jgrund:
a call like:

```rust
                font_awesome(
                    class![C.text_green_500, C.h_4, C.w_4, C.inline],
                    "check-circle",
                ),
```

will render an icon like:

Screen Shot 2019-11-20 at 6 20 21 PM

PR #1338, opened 2019-11-19
Implement initial async retry functions

nlinker:
This is the implementation of `retry_future` and `retry_future_gen`
https://github.com/whamcloud/integrated-manager-for-lustre/issues/1301

Issue #1301, opened 2019-11-01
Create method to retry futures on error


jgrund:
We have a number of futures where it would be useful to retry if a failure occurs. Ideally we could retry a  set number of times before failing.

Implement a method that can be called on a `TryFuture` and will retry on Error.

This needs to be implemented against [Futures 0.3](https://docs.rs/futures-preview/0.3.0-alpha.19/futures/future/index.html)

nlinker:
I use this is little Flask server app to test the client against
```
import time
import flask
from flask.json import jsonify

app = flask.Flask(__name__)

# single threaded server only
counter = {}

@app.route('/echo')
def echo():
    return flask.request.data

@app.route('/timeout')
def timeout():
    time.sleep(flask.request.args.get('sleep', 5))  # sleep 5 seconds before returning
    return 'ok'

@app.route('/fail')
def fail():
    raise RuntimeError('failed')

@app.route('/fail2')
def fail2():
    # method, that fails, when counter2 is odd
    if not 'fail2' in counter:
        counter['fail2'] = 0
    counter['fail2'] = counter['fail2'] + 1
    if counter['fail2'] % 2 == 1:
        raise RuntimeError('failed')
    else:
        return jsonify(counter['fail2'])

if __name__ == '__main__':
    app.run(port=8080)
```

PR #1348, opened 2019-11-25
Implement initial async retry functions

nlinker:
This is the implementation of retry_future and retry_future_gen
#1301
(This is the copy of #1338 but done from the branch, not fork (according to IML workflow))


iml-mailbox/Cargo.toml
```diff
@@ -10,7 +10,7 @@ futures-preview = "0.3.0-alpha.19"
 iml-manager-env = { path = "../iml-manager-env", version = "0.1.0" }
 iml-fs = { path = "../iml-fs", version = "0.1" }
 tokio = "0.2.0-alpha.6"
-warp = { git = "https://github.com/seanmonstar/warp.git" }
+warp = { version = "0.2.0-alpha.0", git = "https://github.com/seanmonstar/warp.git", rev = "9bb58102"}
```
jgrund: This shouldn't need to be pinned.

iml-request-retry/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+tracing = "0.1.10"
+
+[dev-dependencies]
+tokio-test = "0.2.0-alpha.6"
+http = "0.1.19"
+reqwest = { version = "0.10.0-alpha.1", features = ["json"] }
+url = "2.1.0"
+rand_xorshift = "0.2.0"
+warp = { version = "0.2.0-alpha.0", git = "https://github.com/seanmonstar/warp.git", rev = "9bb58102"}
+serde = "1.0.102"
```
jgrund: This can be:

`serde = { version = "1", features = ['derive'] }`

and line 23 can be dropped.

iml-mailbox/Cargo.toml
nlinker: Okay, reverting this.

iml-request-retry/Cargo.toml
nlinker: This fix in `Cargo.toml` caused the failure on the CI, that I cannot reproduce locally. Just weird 🔨. However, I have found (in the reddit /r/rust), that the issue is fixed by
```
- use serde_derive::{Serialize, Deserialize};
+ use serde::{Serialize, Deserialize};
```
This might be reproduced, if one builds it in a docker container. Ugh :-(

PR #1356, opened 2019-11-26
Extract Action plugin concept

jgrund:
Extract the action plugin concept into iml_utils so it can be shared
between crates that run on the agent and the manager.

### DIFF
--- a/iml-util/util.rs
+++ b/iml-util/util.rs
@@ -126,6 +126,68 @@ pub mod tokio_utils {
     }
 }
 
+pub mod action_plugins {
+    use futures::{Future, FutureExt};
+    use iml_wire_types::{ActionName, ToJsonValue};
+    use std::{collections::HashMap, fmt::Display, pin::Pin};
+
+    type BoxedFuture =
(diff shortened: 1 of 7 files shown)


iml-util/util.rs
```diff
@@ -126,6 +126,68 @@ pub mod tokio_utils {
...
 
+pub mod action_plugins {
+    use futures::{Future, FutureExt};
+    use iml_wire_types::{ActionName, ToJsonValue};
+    use std::{collections::HashMap, fmt::Display, pin::Pin};
+
+    type BoxedFuture =
+        Pin<Box<dyn Future<Output = Result<serde_json::value::Value, String>> + Send>>;
+
+    type Callback = Box<dyn Fn(serde_json::value::Value) -> BoxedFuture + Send + Sync>;
```
nlinker: Could you suggest, what the json contains, for example?
What data may appear here?

jgrund: Sure, I'll add a comment, but it's really anything that can be serialized.

This is basically just for RPC, so this is the incoming data from a remote node.

nlinker: Got it, thanks

PR #1357, opened 2019-11-26
Add local action plugins

jgrund:
This patch adds a way to run any sort of Rust code as part of a step
within the manager.

From the Python side, two new methods are exposed that can be called
within a step:

`invoke_rust_local_action` and `invoke_rust_local_action_expect_result`.

They work basically the same as their remote counterparts:

`invoke_rust_agent` and `invoke_rust_agent_expect_result`

Except they run the plugins on the manager and don't require a fqdn to
route the call to.

The first two plugins deal with getting the current session and waiting
for a new session, but future plugins can basically be anything.

COPR Module Manager: managerforlustre/rust-local-plugins

### DIFF
--- a/chroma_core/lib/job.py
+++ b/chroma_core/lib/job.py
@@ -6,7 +6,7 @@ import json
 import django.db.models
 
 from chroma_core.services import log_register
-from chroma_core.lib.util import invoke_rust_agent, RustAgentCancellation
+from chroma_core.lib.util import invoke_rust_agent, invoke_rust_local_action, RustAgentCancellation
 from iml_common.lib.agent_rpc import agent_result
 
 job_log = log_register("job")
@@ -200,6 +200,34 @@ class Step(object):
         except AgentException as e:
             self._log_subprocesses(e.subprocesses)
             raise
+
+    def invoke_rust_local_action(self, command, args={}):
+        """
+        Talks to the iml-action-runner service
+        """
+
+        return invoke_rust_local_action(command, args, self._cancel_event)
+
+    def invoke_rust_local_action_expect_result(self, command, args={}):
+        from chroma_core.services.job_scheduler.agent_rpc import LocalActionException
+
+        try:
+            result = self.invoke_rust_local_action(command, args)
+        except RustAgentCancellation as e:
+            raise LocalActionException(command, args, "Cancelled: {}".format(e))
+        except Exception as e:
+            raise LocalActionException(command, args, "Unexpected error: {}".format(e))
+
+        try:
+            result = json.loads(result)
+        except ValueError as e:
+            raise LocalActionException(command, args, "Error parsing json: {}".format(e))
+
+        if "Err" in result:
+            self.log(json.dumps(result["Err"], indent=2))
+            raise LocalActionException(command, args, result["Err"])
+
+        return result["Ok"]
 
     def invoke_rust_agent(self, host, command, args={}):
         """
--- a/chroma_core/lib/util.py
+++ b/chroma_core/lib/util.py
@@ -225,16 +225,17 @@ def normalize_nid(string):
     return string
 
 
+IML_ACTION_RUNNER_SOCKET_PATH = "http+unix://%2Fvar%2Frun%2Fiml-action-runner.sock/"
+
+
 class RustAgentCancellation(Exception):
     pass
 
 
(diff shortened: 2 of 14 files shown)


chroma_core/lib/job.py
```diff
@@ -201,6 +201,34 @@ def invoke_agent(self, host, command, args={}):
...
 
+    def invoke_rust_local_action(self, command, args={}):
+        """
+        Talks to the iml-action-runner service
+        """
+
+        return invoke_rust_local_action(command, args, self._cancel_event)
+
+    def invoke_rust_local_action_expect_result(self, command, args={}):
+        from chroma_core.services.job_scheduler.agent_rpc import LocalActionException
```
mkpankov: Why is this being imported locally?

chroma_core/lib/job.py
```diff
@@ -201,6 +201,34 @@ def invoke_agent(self, host, command, args={}):
...
+        try:
+            result = self.invoke_rust_local_action(command, args)
+        except RustAgentCancellation as e:
+            raise LocalActionException(command, args, "Cancelled: {}".format(e))
+        except Exception as e:
+            raise LocalActionException(command, args, "Unexpected error: {}".format(e))
+
+        try:
+            result = json.loads(result)
+        except ValueError as e:
```
mkpankov: Is it only ever throwing `ValueError`? Shouldn't we handle other possible exceptions?

chroma_core/lib/job.py
johnsonw: If it's only needed in one particular spot sometimes it's easier to import locally. This is especially the case where circular dependencies happen when importing at the top.

chroma_core/lib/job.py
johnsonw: If there is a circular reference it could throw an `OverflowError`. It could also throw a `TypeError` but I think `ValueError` is really the most common. https://docs.python.org/2.7/library/json.html

chroma_core/lib/util.py
```diff
@@ -225,16 +225,64 @@ def normalize_nid(string):
...
+
+    t = Thread(target=start_action, args=(ActionResult, trigger))
+
+    t.start()
+
+    # Wait for action completion, waking up every second to
+    # check cancel_event
+    while True:
+        if cancel_event.is_set():
+            requests_unixsocket.post(
```
johnsonw: This will need to be updated to work with docker.

chroma_core/lib/util.py
```diff
@@ -247,8 +295,10 @@ class ActionResult:
     def start_action(ActionResult, trigger):
         try:
             ActionResult.ok = requests_unixsocket.post(
```
johnsonw: This will need to be updated for docker.

chroma_core/lib/util.py
```diff
@@ -263,7 +313,9 @@ def start_action(ActionResult, trigger):
     # check cancel_event
     while True:
         if cancel_event.is_set():
-            requests_unixsocket.post(SOCKET_PATH, json=(host, {"type": "ACTION_CANCEL", "id": str(request_id)})).content
+            requests_unixsocket.post(
```
johnsonw: This will need to be updated for docker.

chroma_core/lib/job.py
jgrund: > If there is a circular reference it could throw an OverflowError

Given this is *deserializing* a JSON string, I'm not sure how an `OverflowError` is possible.

Also if we reach this point, we are basically guaranteed to be dealing with `bytes`, so I don't think TypeError applies either.

Long story short, `ValueError` is the only exception we should worry about here.

chroma_core/lib/util.py
jgrund: Yes, but your patch hasn't landed yet.

Whichever patch lands first, the other will need to rebase and update..

chroma_core/lib/util.py
jgrund: Yes, but your patch hasn't landed yet.

Whichever patch lands first, the other will need to rebase and update..

chroma_core/lib/job.py
jgrund: I suspect it's either to avoid side-effects from importing too early or to avoid a circular dependency.

The other functions that use these exceptions do the same thing, so I'm basically carrying it forward.

PR #1358, opened 2019-11-27
Revert "Fix RemovedInDjango20Warning for user.is_authenticated (#1334)"

jgrund:
This reverts commit 21973d6b33250b21c24d75a92264d8d342372626.

### DIFF
--- a/chroma_api/alert.py
+++ b/chroma_api/alert.py
@@ -53,7 +53,7 @@ class AlertSubscriptionAuthorization(PatchedDjangoAuthorization):
         if request.method is None:
             # Internal request, it's all good.
             return object_list
-        elif not request.user.is_authenticated:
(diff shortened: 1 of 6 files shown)


johnsonw:
What issue was this causing?

jgrund:
This unfortunately breaks authentication. Tastypie checks auth by strictly checking against the value `True`, and calling the magic property version of this returns a `Callablebool(True)`.

Using the magic property version of this call returns something different than the method call:

Screen Shot 2019-11-26 at 7 18 34 PM

And tastypie will reject anything that's not exactly `True`, causing all auth checks to fail.

Screen Shot 2019-11-26 at 7 21 03 PM


We should be able to apply this patch once tastypie updates to check both kinds of true 😄

johnsonw:
Ah I see!

PR #1360, opened 2019-12-02
Wire up more routes

jgrund:
This patch wires up more routes and adds their links to the menu.

It also cleans up the styling a bit so it's closer to spec.

Still need to fixup smaller than lg layout a bit, but that can happen in
another patch.

menu-bar


iml-gui/crate/src/page/partial/header.rs
```diff
@@ -1,45 +1,83 @@
...
+            li![
+                a![&cls, "Filesystems"],
+                attrs! {
+                    At::Href => Page::Filesystem.to_href(),
+                },
+            ],
+            li![a![&cls, "HSM"]],
+            li![a![&cls, "Storage"]],
+            li![
+                a![&cls, "Users"],
```
ip1981: I noticed that menu items and page titles differ from each other, maybe `p.to_string()` could be used for menu items too?

jgrund: Yeah, that makes sense.

PR #1361, opened 2019-12-02
Add activity indicator

jgrund:
Add the activity indicator that will update in realtime as activity
changes.

yellow-to-red

red-to-green


iml-gui/crate/src/components/activity_indicator.rs
```diff
@@ -0,0 +1,49 @@
...
+        health = AlertSeverity::ERROR
+    } else if s.contains(&AlertSeverity::WARNING) {
+        health = AlertSeverity::WARNING;
+    }
+
+    ActivityHealth {
+        health,
+        count,
+    }
+}
```
ip1981: I just performed some Rust exercise:

If  we had this definition of  `AlertSeverity`:
```rust
#[derive(
    serde::Serialize, serde::Deserialize, Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord,
)]
pub enum AlertSeverity {
    DEBUG,
    INFO,
    WARNING,
    ERROR,
    CRITICAL,
}
```

then we could have it much shorter (and probably faster, at least with just a single pass):
```rust
pub fn update_activity_health(
    active_alert: &HashMap<u32, Alert>,
) -> ActivityHealth {
    let (health, count) = active_alert
        .iter()
        .filter(|(_, x)| x.severity > AlertSeverity::INFO)
        .fold((AlertSeverity::INFO, 0), |(h, c), (_, x)| {
            (max(h, x.severity), c + 1)
        });

    ActivityHealth {
        health,
        count,
    }
}
```

jgrund: Sure, I'll update. We may as well just fold over `ActivityHealth` directly.

iml-wire-types/wire-types.rs
```diff
@@ -890,7 +890,9 @@ pub enum AlertType {
     StratagemUnconfiguredAlert,
 }
 
-#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
+#[derive(
+    serde::Serialize, serde::Deserialize, Copy, Clone, Debug, Hash, PartialOrd, Ord, PartialEq, Eq,
+)]
 pub enum AlertSeverity {
     INFO,
     DEBUG,
```
ip1981: I think order is important for Ord to make sense.

iml-wire-types/wire-types.rs
```diff
@@ -890,7 +890,9 @@ pub enum AlertType {
     StratagemUnconfiguredAlert,
 }
 
-#[derive(serde::Serialize, serde::Deserialize, Clone, Debug, Hash, PartialEq, Eq)]
+#[derive(
+    serde::Serialize, serde::Deserialize, Copy, Clone, Debug, Hash, PartialOrd, Ord, PartialEq, Eq,
```
ip1981: Nit: Hash wasn't used, probably it is not needed.

iml-wire-types/wire-types.rs
jgrund: Good catch.

iml-gui/crate/src/components/activity_indicator.rs
```diff
@@ -0,0 +1,44 @@
...
+            count: 0,
+        }
+    }
+}
+
+pub fn activity_indicator<T>(activity_health: &ActivityHealth) -> Node<T> {
+    span![
+        class![C.mr_3, C.text_green_500 => activity_health.health == AlertSeverity::INFO,
+                            C.text_yellow_500 => activity_health.health == AlertSeverity::WARNING,
+                            C.text_red_500 => activity_health.health == AlertSeverity::ERROR],
```
ip1981: While not all level are used, maybe there could be green for <= info, yellow for == warning, red for >= error. Later it would be trivial to add more colors :)

jgrund: It could, but I don't see these levels ever changing at the API level beyond those three.

PR #1370, opened 2019-12-04
Enable detail routing

jgrund:
This patch enables us to use a route that contains data (meaning a route
enum variant can now carry some data).

This enables detail pages where we need to include an id in the route.

In addition, this patch renames `Page` enum to `Route`, which seems like
a better name, and extracts it to it's own file.

detail-routing

### DIFF
--- a/iml-gui/crate/src/route.rs
+++ b/iml-gui/crate/src/route.rs
@@ -1,0 +1,127 @@
+use seed::prelude::*;
+use std::{borrow::Cow, ops::Deref};
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct RouteId<'a>(Cow<'a, str>);
+
+impl<'a> From<u32> for RouteId<'a> {
+    fn from(n: u32) -> Self {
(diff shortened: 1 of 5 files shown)


iml-gui/crate/src/route.rs
```diff
@@ -0,0 +1,127 @@
...
+            Self::FilesystemDetail => "Filesystem Detail".into(),
+            Self::Home => "Home".into(),
+            Self::Jobstats => "Jobstats".into(),
+            Self::Login => "Login".into(),
+            Self::Logs => "Logs".into(),
+            Self::Mgt => "MGTs".into(),
+            Self::NotFound => "404".into(),
+            Self::PowerControl => "Power Control".into(),
+            Self::Server => "Servers".into(),
+            Self::ServerDetail(_) => "Server Detail".into(),
```
ip1981: Maybe it should use server id somehow.

jgrund: Yeah. It should probably resolve the id into the corresponding db record. But I figure that can happen in a separate patch.

PR #1371, opened 2019-12-05
Implement checking for loaded kernel modules

mkpankov:
The main pain point here is `MyKmodError`. I introduced it because wrapping `kmod::Error` directly in a variant of `ImlAgentError` broke code in unrelated parts of IML because `ImlAgentError` stopped being `Sync` apparently. So there's this middleware conversion.

Anyone knows how to deal with that?

#1187

### DIFF
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -1,5 +1,19 @@
-use crate::agent_error::ImlAgentError;
+use crate::agent_error::{ImlAgentError, MyKmodError};
+
+use kmod;
 
 pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
-    Ok(false)
+    let ctx = kmod::Context::new().map_err(|e| MyKmodError::from(e))?;
+
+    let mut result = false;
+    for m in ctx.modules_loaded().map_err(|e| MyKmodError::from(e))? {
+        let name = m.name();
+
+        if name == module {
+            result = true;
+        } else {
+            continue;
+        }
+    }
+    Ok(result)
 }
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -5,15 +5,5 @@ use kmod;
 pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
     let ctx = kmod::Context::new()?;
 
-    let mut result = false;
-    for m in ctx.modules_loaded()? {
-        let name = m.name();
-
-        if name == module {
-            result = true;
-        } else {
-            continue;
-        }
-    }
-    Ok(result)
+    Ok(ctx.modules_loaded()?.find(|m| m.name() == module).is_some())
 }
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -1,0 +1,5 @@
+use crate::agent_error::ImlAgentError;
+
+pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
+    Ok(false)
+}
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -1,12 +1,12 @@
-use crate::agent_error::{ImlAgentError, MyKmodError};
+use crate::agent_error::ImlAgentError;
 
 use kmod;
 
 pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
-    let ctx = kmod::Context::new().map_err(|e| MyKmodError::from(e))?;
+    let ctx = kmod::Context::new()?;
 
     let mut result = false;
-    for m in ctx.modules_loaded().map_err(|e| MyKmodError::from(e))? {
+    for m in ctx.modules_loaded()? {
         let name = m.name();
 
         if name == module {
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -10,7 +10,10 @@ pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
     run(move || {
         let ctx = kmod::Context::new()?;
 
-        Ok(ctx.modules_loaded()?.find(|m| m.name() == module).is_some())
+        Ok(ctx
+            .modules_loaded()?
+            .find(|m| m.name() == module && m.refcount() > 0)
+            .is_some())
     })
     .await
 }
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -10,10 +10,7 @@ pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
     run(move || {
         let ctx = kmod::Context::new()?;
 
-        Ok(ctx
-            .modules_loaded()?
-            .find(|m| m.name() == module && m.refcount() > 0)
-            .is_some())
+        Ok(ctx.modules_loaded()?.find(|m| m.name() == module).is_some())
     })
     .await
 }
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -1,3 +1,7 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
 use crate::agent_error::ImlAgentError;
 use kmod;
 use tokio_executor::blocking::run;
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -1,9 +1,12 @@
 use crate::agent_error::ImlAgentError;
-
 use kmod;
+use tokio_executor::blocking::run;
 
 pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
-    let ctx = kmod::Context::new()?;
+    run(move || {
+        let ctx = kmod::Context::new()?;
 
-    Ok(ctx.modules_loaded()?.find(|m| m.name() == module).is_some())
+        Ok(ctx.modules_loaded()?.find(|m| m.name() == module).is_some())
+    })
+    .await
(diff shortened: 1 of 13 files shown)


iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,19 @@
...
+    let ctx = kmod::Context::new().map_err(|e| MyKmodError::from(e))?;
+
+    let mut result = false;
+    for m in ctx.modules_loaded().map_err(|e| MyKmodError::from(e))? {
+        let name = m.name();
+
+        if name == module {
+            result = true;
+        } else {
+            continue;
```
ip1981: Is it needed?

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,19 @@
...
+
+pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
+    let ctx = kmod::Context::new().map_err(|e| MyKmodError::from(e))?;
+
+    let mut result = false;
+    for m in ctx.modules_loaded().map_err(|e| MyKmodError::from(e))? {
+        let name = m.name();
+
+        if name == module {
+            result = true;
```
ip1981: Can return from the function immediately?

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,19 @@
...
+use kmod;
+
+pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
+    let ctx = kmod::Context::new().map_err(|e| MyKmodError::from(e))?;
+
+    let mut result = false;
+    for m in ctx.modules_loaded().map_err(|e| MyKmodError::from(e))? {
+        let name = m.name();
+
+        if name == module {
```
ip1981: Maybe use something like `find(...).is_some()` with `ModuleIterator`?

mkpankov:
It also fails to build because there's now a dependency on `kmod-devel` RPM. Where do I add that?

mkpankov:
I reworked the `kmod-rs` and dropped `error-chain` from there. So it's all clean on our side now.

iml-agent/src/action_plugins/kernel_module.rs
mkpankov: Sure, dropped the draft code

mkpankov:
It needs https://github.com/whamcloud/Dockerfiles/pull/40 to build, but otherwise ready for review

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,9 @@
+use crate::agent_error::ImlAgentError;
+
+use kmod;
+
+pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
+    let ctx = kmod::Context::new()?;
```
jgrund: I haven't looked very deeply at what happens in kmod_sys for this, but if it does any sort of work it should be inside a `run` call so it's in a separate blocking thread.

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,9 @@
...
+
+    Ok(ctx.modules_loaded()?.find(|m| m.name() == module).is_some())
```
jgrund: Since this is not async by itself, this call should be inside a blocking thread.

ref: 

https://github.com/whamcloud/integrated-manager-for-lustre/blob/203a2d67cdb8bdb0ece520f01e1997ac451e128e/iml-agent/src/action_plugins/stratagem/action_purge.rs#L24-L26

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,9 @@
+use crate::agent_error::ImlAgentError;
+
```
jgrund:
```suggestion
```

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,9 @@
+use crate::agent_error::ImlAgentError;
```
jgrund:
```suggestion
// Copyright (c) 2019 DDN. All rights reserved.
// Use of this source code is governed by a MIT-style
// license that can be found in the LICENSE file.

use crate::agent_error::ImlAgentError;
```

iml-agent/src/action_plugins/kernel_module.rs
mkpankov: Right, I've been thinking about this too

mkpankov:
Ok so it's harder than I thought.

There are failures on Travis and GitHub Actions. This comment is about GitHub Actions.

GitHub Actions CI runs on Ubuntu (defined in `.github/workflows/rust-ci.yml`):
```
  test:
    name: Test Suite
    runs-on: ubuntu-latest
...
```

Other options [are](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on) Windows and MacOS.

I'm now going to install development libraries of `kmod` for Ubuntu in the job itself.

It's probably OK for this particular PR, but we could get to situation where there's a version mismatch between libraries for Ubuntu and our target CentOS 7, or other dependencies on CentOS environment. This will become more apparent as we have more bindings to native libraries.

jgrund:
Do you think it's worth introducing these native deps over just shelling out for the calls?

utopiabound:
> Do you think it's worth introducing these native deps over just shelling out for the calls?

Just shelling out is fine.  Getting clang involved seem heavyweight.

mkpankov:
@jgrund I think yes, and next dependencies can be added much easier, if the need arises.

Also, I already have implementation of getting kernel module versions in separate branch, based on this.

So this and one new smaller PR will cover points 2 and 3 of https://github.com/whamcloud/integrated-manager-for-lustre/issues/1187#issuecomment-559830336

Overall, I think it's worth discussing our strategy on interaction with outside world. Overall

* Shelling out
  * Pros
    * Easier to setup, no native dependencies
    * Safer, as crash in separate process won't affect the process
  * Cons
    * Underspecified, ad-hoc parsing may be prone to errors
    * Handling errors isn't straightforward, some CLI tools don't properly exit with non-zero code when errors happen, print errors to different streams
    * Less performant than direct binding to the library
    * More complex interfaces are harder to represent and parse output of, might involve a lot of boilerplate
* Binding
  * Pros
    * Just as specified as API of particular software can be
    * Best error handling capabilities of all possible alternatives
    * More performant
    * API is represented 1:1
  * Cons
    * More complex setup, native dependencies
    * Less safe, crash in native library will bring down entire process

I don't have all the context, so I'm not sure how much weight there is on each point for `iml-agent` and IML as a whole.

jgrund:
> I don't have all the context, so I'm not sure how much weight there is on each point for `iml-agent` and IML as a whole.

I think there is a case for native dependencies, but I think it's very important to not tie them into the main dependency tree as it then forces *everything* to support those deps.

IML already uses a few native libraries (namely `libzfs` and `liblustre`), but neither is tied directly into the tree, allowing us to build IML without those libs and their build-time deps in place.

Using native deps without hardcoding their deps into the tree can be done in a few ways.

- In the case of `libzfs` we have bindings and a rust wrapper [here](https://github.com/whamcloud/rust-libzfs), and usage is scoped to the [device-scanner](https://github.com/whamcloud/device-scanner) repo (which specializes in detecting / updating storage in realtime). The iml-agent consumes data from device-scanner over UDS, isolating the rest of the iml project from owning `libzfs` and `clang`.

- In the case of [`liblustre` usage](https://github.com/whamcloud/integrated-manager-for-lustre/tree/42895261d929268061a6871e619402def681beb8/liblustreapi), a different approach was taken. https://docs.rs/libloading/0.5.2/libloading/ Is used to get around needing `liblustre` at compile time, and building is [skipped](https://github.com/whamcloud/integrated-manager-for-lustre/blob/42895261d929268061a6871e619402def681beb8/liblustreapi-types/build.rs) if the bindings file already exists, which avoids the need for clang.

In general, I think using native deps allows for a nicer (and possibly more stable) API than reading stdout and parsing with regex / parser-combinator etc..., but it needs to be balanced against the cost of owning a native wrapper lib and whether CLI output is thought of as a stable interface for a given project.

mkpankov:
>I think there is a case for native dependencies, but I think it's very important to not tie them into the main dependency tree as it then forces everything to support those deps.

It seems you're talking about [this](https://github.com/rust-lang/cargo/issues/7020) problem?

Other than that, dependencies are per workspace package, so in our case only `iml-agent` and `iml-agent-daemon` do get `libkmod` - not entire project.

I'm fixing it via `libloading` as you suggested.

This does look like it has one downside - when library incompatibly updates on target system, we won't notice it until runtime, as we're not even linking to it during build. So that means we'll have to look into a bug or a system test failure instead of failing to build a specific package.

mkpankov:
I've looked at `libloading` and it means I'll effectively need to rewrite all of bindings, because the loaded native library becomes the owned value itself, and everything is tied to its lifetime, instead of being implicit global state. 

Rewriting itself is effort, but I also think it makes the API worse. I believe `libloading` usecase is different and best when you want fine-grained control over search of the library, or only need specific symbols, or hot reloading.

Also, I think it's possible to restrict the runtime dependency on `libkmod` to only `iml-agent` binary by splitting `iml-agent` crate into `iml-agent` CLI and `iml-agent-daemon` - basically separate the two current binaries.

But that doesn't help with `libkmod` and `clang` being present in build containers - they're still needed. Currently, I don't quite understand if that's what you wanted us to avoid. Why is it a problem?

I only see possible problem in that our build infrastructure is currently split and different CI passes use different services and containers. If we were to reconcile this to only our target system, adding further bindings available at build time would be easy enough.

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,19 @@
...
+use kmod;
+use tokio_executor::blocking::run;
+
+pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
+    run(move || {
+        let ctx = kmod::Context::new()?;
+
+        Ok(ctx
+            .modules_loaded()?
+            .find(|m| m.name() == module && m.refcount() > 0)
```
ip1981:
```suggestion
            .find(|m| m.name() == module)
```
?

mkpankov:
Closing in favor of #1402

Issue #1363, opened 2019-12-03
Utilize notifications API for incoming alerts


jgrund:
The [notifications api](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API) can be used to send system level messages from the browser.

We can use this to display any incoming active alert changes. This will replace the current way of displaying alert changes, which does this:

![alert-slider.gif](https://images.zenhubusercontent.com/5d25f5c19c24773a1026b5eb/f35373f5-47d3-4759-a9e9-1ff7dd0ce22a)

Some implementation notes:

1. We should display the logo as part of the notification.
1. We should avoid spamming notifications by using the tag property when sending notifications.

PR #1373, opened 2019-12-05
Use browser notification API to show alerts

ip1981:
#1363.

napi


ip1981:
I think that's it. It is mostly what it looks like at the picture, except now there is only one notification widget. This widget is forcibly closed after 5 seconds to work around possible system and browser differences.

ip1981:
Closing in favor of https://github.com/whamcloud/integrated-manager-for-lustre/pull/1432

Issue #1378, opened 2019-12-06
create `lctl` action plugin


jgrund:
Create an action plugin `lctl` (in rust) that does the following:

1. Run `lctl vec<String>` that returns a `String`.

PR #1383, opened 2019-12-06
agent: Add lctl action plugin

utopiabound:
lctl(Vec<String>) -> String

Fixes Issue #1378 

### DIFF
--- a/iml-agent/src/action_plugins/lctl.rs
+++ b/iml-agent/src/action_plugins/lctl.rs
@@ -4,7 +4,7 @@
 
 use crate::{
     agent_error::ImlAgentError,
-    cmd::{self},
+    cmd,
 };
 
 /// Runs lctl with given arguments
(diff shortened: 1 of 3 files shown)


iml-agent/src/action_plugins/lctl.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::{
+    agent_error::ImlAgentError,
+    cmd::{self},
```
jgrund:
```suggestion
    cmd,
```

utopiabound: Don't I need cmd::{self} if I reference cmd?

jgrund: It should work without it.

jgrund: `self` is just a helper if you are importing other things under that namespace and still want to use `cmd` as well.

PR #1393, opened 2019-12-10
Implement context-sensitive help

### LINKED ISSUE #1352: Implement context sensitive help
The help menu icon should open the embedded help to the relevant page (this can be determined by what page we are currently on).

Prior art, though the current implementation can be *much* simpler.

https://github.com/whamcloud/GUI/blob/f57465618fd4a91327540259cb9bbde99f77a745/source/iml/help-mapper/help-mapper-directive.js#L8-L33

Probably easiest to associate a new `HelpPage` trait with the existing `Page` concept.

nlinker:
Fixes #1352
context-sensitive-help


iml-gui/crate/src/components/activity_indicator.rs
```diff
@@ -19,6 +19,7 @@ pub fn update_activity_health(
     )
 }
 
+#[derive(Debug)]
```
nlinker: This is to debug model contents

iml-gui/crate/src/lib.rs
```diff
@@ -5,10 +5,14 @@
 #![allow(clippy::non_ascii_literal)]
 #![allow(clippy::enum_glob_use)]
 
+#[macro_use]
+extern crate seed;
```
nlinker: This is to enable `log!` macro from `seed` framework

jgrund: With the 2018 edition you can bring macros into scope as needed instead of importing them globally.

In order to use this, you can just do `seed::log!(...)` wherever you want to log.

iml-gui/crate/src/page/partial/header.rs
```diff
@@ -193,6 +186,23 @@ fn main_menu_items(model: &Model) -> Node<Msg> {
     ]
 }
 
+fn context_sensitive_help_link(model: &Model, menu_class: &Attrs) -> Node<Msg> {
+    let attrs = if let Some(help_link) = model.route.help_link() {
+        // open the link in a new tab
+        attrs! { At::Target => "_blank", At::Href => help_link }
+    } else {
+        attrs! { At::Href => "" }
```
jgrund: If we happen to hit a page that doesn't have a corresponding help page, we should default to just opening `CTX_HELP` I think. The `help_link` could just return `Option<String>` where `String` is `"#<SPECIFIC_SECTION_HERE"`. Then this code could look something like:

 ```rust
    let attrs = attrs! {
        At::Target => "_blank",
        At::Href => format!(
            "{}{}",
            &CTX_HELP,
            model.route.help_link().unwrap_or_else(|| "".into())
        )
    };
```

File: iml-gui/crate/src/lib.rs [hunk 4]
```diff
@@ -29,7 +33,18 @@ const STATIC_PATH: &str = "static";
...
+/// This depends on where and how https://github.com/whamcloud/Online-Help is deployed.
+/// With `nginx` when config is like
+/// ```
+/// location /help {
+///     alias /usr/lib/iml-manager/iml-online-help;
+///     index index.html;
+/// }
+/// ```
+/// help url becomes `https://localhost:8443/help/docs/Graphical_User_Interface_9_0.html`
+const CTX_HELP: &str = "help/docs/Graphical_User_Interface_9_0.html";
```
ip1981: Should the link be absolute?

iml-gui/crate/src/lib.rs
nlinker: Yeah, indeed.

iml-gui/crate/src/lib.rs
nlinker: I think, it is too strong dependency, because the host and port number are pretty easy to change.
The downside is that we cannot have `Online-Help` on a different host, which can be fixed with an option in the (hypothetical) application config.

ip1981: I meant starting with a slash: `/help...` :)

nlinker: Hah, lol. I tried to do this, and this breaks the logic and refuses to open the new tab, instead it goes to 404 :-P
image

ip1981: But removing `/` does not sound like a fix ;-) Something else is not right. What's going on? On your picture I don't see the full URL in the status bar before you click on Help.

nlinker: image

nlinker: (More visual info)
image

ip1981: Ok, it works because of `<base href="/">`

nlinker: You think `base href=""` is more preferrable in general case?

jgrund: I think it's best to leave this decision undefined for now. The existing GUI uses `/ui/` for it's base href (and is served from `/ui/` by nginx) which sidesteps the issue of `/help` being a different base.

We may / may not want to continue doing that, but I think we need to push this behind nginx in a production scenario before we decide what we should do here.

nlinker:
> Debug is not needed?

Personally, I used it quite extensively while I tried to get understanding app logic.
Moreover, since the debugging of wasms is limited currently, I think logging becomes the only viable choice to see the internals.

ip1981:
It is ok to add  and then remove Debug as needed ;-)

jgrund:
> I've thought _target is a way to skip client-side routing, but why doesn't it work with absolute hrefs?

It is surprising that _target doesn't work that way.  Probably worth opening an issue on the seed repo asking about it.

PR #1364, opened 2019-12-03
Ostpool: Add agent daemon and queue processing service

utopiabound:
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1364)

PR #1394, opened 2019-12-10
Add ostpool notify support

jgrund:
Add LISTEN / NOTIFY support to the chroma_core_ost table.

This patch also extracts the logic to add / remove LISTEN / NOTIFY to
the __init__.py of migrations so it can be reused.


iml-wire-types/wire-types.rs
```diff
@@ -1212,6 +1212,84 @@ pub mod db {
         }
     }
 
+    /// Record from the `chroma_core_ostpool` table
+    #[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
+    pub struct OstPoolRecord {
+        pub id: u32,
+        pub name: String,
+        pub filesystem_id: u32,
+        pub not_deleted: Option<bool>,
```
nlinker: Probably `is_deleted` is preferable, since it has nice default (`false`, i.e. present) and doesn't require to perform double negation in `false` case.

However, if the field name clones the name from the other part of the system, then it is better to leave it as is.

jgrund: I agree 100%, but this naming convention predates me and is persistent throughout the API / db layer (which this type is simply representing.

iml-wire-types/wire-types.rs
```diff
@@ -1212,6 +1212,84 @@ pub mod db {
         }
     }
 
+    /// Record from the `chroma_core_ostpool` table
+    #[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
+    pub struct OstPoolRecord {
```
utopiabound: This duplicates #1364

jgrund: I wouldn't mind using your changes, but if so can you separate them out from that patch? Otherwise, let's just land this, as it's needed for this patch to function.

PR #1395, opened 2019-12-10
Implement create_ltuer_conf

### LINKED ISSUE #1382: create_ltuer_conf plugin
create a `create_ltuer_conf` plugin that does the following:

1. (Client) Writes the following conf out to `/etc/iml/ltuer.conf` that contains:

    ```
     mailbox=<MAILBOX_PATH>
     fs_name=<FS_NAME>
     cold_pool=<COLD_POOL_NAME>
    ```

mkpankov:
Not sure about the test - it's automated, but brittle and will break in case someone adds another test. Need your opinions.

```
USAGE:
    iml-agent create_ltuer_conf <MAILBOX_PATH> <FS_NAME> <COLD_POOL>
```

Fixes #1382


iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,46 @@
...
+// license that can be found in the LICENSE file.
+
+use crate::agent_error::ImlAgentError;
+use std::path::PathBuf;
+use tokio_fs as fs;
+
+#[cfg(not(test))]
+static CONFIGURATION_DIR: &str = "/etc/iml";
+#[cfg(test)]
+static CONFIGURATION_DIR: &str = "/tmp/etc/iml";
```
ip1981: It is possible-to avoid working with files at all in tests? 
If not, maybe use something like `mktemp`.?
What about cleaning after tests? Running multiple tests in parallel?

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,46 @@
...
+    );
+
+    let mut path = PathBuf::from(CONFIGURATION_DIR);
+    path.push("ltuer.conf");
+
+    fs::write(path, contents).await.map_err(|e| e.into())
+}
+
+// Warning: only one test at a time can be executing,
+// because they write to the same one file
```
ip1981: While I doubt a test is needed at all, I think it can be improved by separating formatting from I/O.

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,46 @@
...
+use tokio_fs as fs;
+
+#[cfg(not(test))]
+static CONFIGURATION_DIR: &str = "/etc/iml";
+#[cfg(test)]
+static CONFIGURATION_DIR: &str = "/tmp/etc/iml";
+
+pub async fn create_ltuer_conf(
+    (mailbox_path, fs_name, cold_pool): (String, String, String),
+) -> Result<(), ImlAgentError> {
```
nlinker: It would be nice to describe the reason why the signature this peculiar? (Probably the reason is that some JSON should be convertible to the arguments, right?)

iml-agent/src/action_plugins/ltuer.rs
nlinker: Agree

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,46 @@
...
+    let contents = format!(
+        "mailbox={}\nfs_name={}\ncold_pool={}\n",
+        mailbox_path, fs_name, cold_pool
+    );
+
+    let mut path = PathBuf::from(CONFIGURATION_DIR);
+    path.push("ltuer.conf");
+
+    fs::write(path, contents).await.map_err(|e| e.into())
+}
```
nlinker: When the file is closed?

iml-agent/src/action_plugins/ltuer.rs
ip1981: In addition, if the file from previous run is left, and the plugin is broken, the test will pass :)

jgrund: If the dir needed to be configurable, you could make a higher-order function that took the dir and returned the plugin. Then you could use something like https://docs.rs/tempfile/3.1.0/tempfile/fn.tempdir.html to ensure there was a unique dir for each run. 

That said, I'm not sure how a closure returned from a HOF would mesh with the current plugin interface (might be ok).

iml-agent/src/action_plugins/ltuer.rs
jgrund: Basically it's adhering to the plugin interface that takes a unary fn where the param is deserializable by serde and the result is serializable by serde.

ref: 

https://github.com/whamcloud/integrated-manager-for-lustre/blob/a3c1ed56df8d321709f36ac401e7115bb6a5bfa1/iml-util/util.rs#L137-L143

and:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/a3c1ed56df8d321709f36ac401e7115bb6a5bfa1/iml-agent/src/action_plugins/action_plugin.rs#L20-L53

mkpankov: Resolving as it doesn't have to do with this plugin itself

iml-agent/src/action_plugins/ltuer.rs
mkpankov: I reworked the test.

I didn't use HOF but rather couple of wrappers.

iml-agent/src/action_plugins/ltuer.rs
mkpankov: Why do you think a test isn't needed?

ip1981: No, the question would be "why a test is needed?" :) The function being tested is trivial (string formatting and IO tested elsewhere), the test has some knowledge about implementation (e. g. order of options in config file), this feel like a tautological test. Also attempt to write a test results in much more complicated code. That's why I doubt it.

mkpankov: The tests are always tautological and of course they would know order of options in config file, because that's the result of the function called.

I agree this makes the code more complex, but I think we need to explore writing smaller tests and this is first step.

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,76 @@
...
+        .write(contents.as_bytes())
+        .await
+        .map(|_| ())
+        .map_err(|e| e.into())
+}
+
+async fn wrap_file_creation<P>(
+    (mailbox_path, fs_name, cold_pool): (String, String, String),
+    dir_path: P,
+) -> Result<(), ImlAgentError>
```
nlinker: For me the name is suboptimal. Essentially it is `create_ltuer_conf_internal`, and the function above, which is writing stuff is something like `write_ltuer_conf`.

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,76 @@
...
+// license that can be found in the LICENSE file.
+
+use crate::agent_error::ImlAgentError;
+use std::path::{Path, PathBuf};
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio_fs as fs;
+
+static CONFIGURATION_DIR: &str = "/etc/iml";
+
+async fn create_ltuer_conf_internal<W>(
```
nlinker: I'd suggest a name like `write_ltuer_conf` or so.

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -41,7 +41,8 @@ pub fn create_registry() -> action_plugins::Actions {
         .add_plugin(
             "configure_ntp",
             action_configure::update_and_write_new_config,
-        );
+        )
+        .add_plugin("create_ltuer_conf", ltuer::create_ltuer_conf);
```
ip1981: While you are here, could you sort them all?

iml-agent/src/action_plugins/ltuer.rs
mkpankov: Good suggestion, renamed

iml-agent/src/action_plugins/ltuer.rs
mkpankov: Done

iml-agent/src/action_plugins/action_plugin.rs
mkpankov: Maybe in separate PR

iml-agent/Cargo.toml
```diff
@@ -41,6 +42,7 @@ iml-wire-types = { path = "../iml-wire-types", version = "0.2" }
 liblustreapi = { path = "../liblustreapi", version = "0.1" }
 iml-fs = { path = "../iml-fs", version = "0.1.0" }
 iml-util = { path = "../iml-util", version = "0.1.0" }
+tempfile = "3.1.0"
```
ip1981: Since it is used in tests only, it should go to [dev-dependencies].

mkpankov: Done

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::agent_error::ImlAgentError;
+use std::path::{Path, PathBuf};
+use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tokio_fs as fs;
```
ip1981: Is there any reason to prefer to`kio::fs` over `std::fs`? I can see that you use asynchronous functions always with `.await`.

ip1981: I much as I understand, only `write_ltuer_conf` is required to be async, Everything else inside it wuold be simpler if synchronous.

jgrund: Everything in iml-agent is required to be async.

There are many tasks that need to make progress over a finite number of cores. If any one call is blocking then it doesn't move into an idle state when possible and allow another task to run instead. This reduces the amount of concurrency possible.

In contrast the async version (tokio_fs in this case) will move into an idle state when it's blocked on an external resource, freeing up to allow another task to be moved forward.

This is a very important invariant as we scale the number of things happening at once through the use of action plugins and daemon plugins.

ip1981: Ok, I think I have confused `.await` with `block_on` :)

ip1981: At least it could be 
```suggestion
use tokio::fs;
```

mkpankov: We use separate crate versions of all `tokio` components at the moment (`iml-agent/Cargo.toml`):

```
tokio = "0.2.0-alpha.6"
tokio-fs = "0.2.0-alpha.6"
tokio-net = { version = "0.2.0-alpha.6", features = ["process"] }
tokio-timer = "0.3.0-alpha.6"
tokio-executor = "0.2.0-alpha.6"
tokio-tls = "0.3.0-alpha.6"
```

ip1981: Hm, tokio::fs works as is, and there are both variant for others (e. g. tokio::timer vs tokio_timer) :)

ip1981: It looks like tokio is moving towards providing everything under the root crate tokio... Right since version 0.2.0 (after the alphas).

mkpankov: I don't think upgrading/merging to one crate should happen in this PR.

Using separate creates doesn't bear any downsides currently. With one crate, all of components are behind features anyway.

ip1981: I agree, but there is not need in upgrade. I am trying to say, that even 0.2.0.alpha.6 enables and reexports `fs` by default, so `tokio::fs` works right now, and will work after upgrade.

mkpankov: Current solution is consistent with rest of codebase. I don't see value in breaking that consistency just for one source file and crate.

All of cleanup should happen simultaneously as work on separate issue.

ip1981: I don't think there is much of consistency :) Only tokio::fs variant has been used so far, but others, as I figured, e. g. tokio::timer, represented in both variants. So I'd say tokio_fs breaks consistency :D

mkpankov: No.

Off all imported crates

```
tokio = "0.2.0-alpha.6"
tokio-fs = "0.2.0-alpha.6"
tokio-net = { version = "0.2.0-alpha.6", features = ["process"] }
tokio-timer = "0.3.0-alpha.6"
tokio-executor = "0.2.0-alpha.6"
tokio-tls = "0.3.0-alpha.6"
```

only `tokio-tls` is not used in imports.

ip1981: You have  added tokio-fs as an explicit dependency in this very PR, but there were already `tokio::fs` in the code base. I admit that _in some places_ there was a reason to use tokio-timer 0.3.0 over whatever the root crate brings in (if anything), but for `fs` it seems unnecessary :)

mkpankov: Ok

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,78 @@
...
+// license that can be found in the LICENSE file.
+
+use crate::agent_error::ImlAgentError;
+use std::path::{Path, PathBuf};
+use tokio::{
+    fs,
+    io::{AsyncWrite, AsyncWriteExt},
+};
+
+static CONFIGURATION_DIR: &str = "/etc/iml";
```
jgrund: We should probably add this as an environment variable to https://github.com/whamcloud/integrated-manager-for-lustre/blob/402a75118afbd232c861273a47b9b7bf1b8396e1/iml-agent/systemd-units/rust-iml-agent.service#L7-L17


That way it's configurable outside the code, and can be used by all agent modules.

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,78 @@
...
+    create_ltuer_conf_internal((mailbox_path, fs_name, cold_pool), CONFIGURATION_DIR).await
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use tempfile::tempdir;
+
+    #[tokio::test]
+    async fn test_create_ltuer_conf() {
```
jgrund: nit: If you have this fn return  `Result<(), Box<dyn std::error::Error>>` then you can use the `?` operator instead of unwrapping

iml-agent/src/action_plugins/ltuer.rs
```diff
@@ -0,0 +1,78 @@
...
+) -> Result<(), ImlAgentError>
+where
+    W: AsyncWrite + Unpin,
+{
+    let contents = format!(
+        "mailbox={}\nfs_name={}\ncold_pool={}\n",
+        mailbox_path, fs_name, cold_pool
+    );
+
+    writer
```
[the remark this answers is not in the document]
mkpankov: Done

iml-agent/src/action_plugins/ltuer.rs
mkpankov: Done

iml-agent/src/action_plugins/ltuer.rs
mkpankov: I'm doing this on purpose, so that any unexpected error is fail-fast in test.

jgrund: It would still fail the same way, the question mark just returns the `Error` early.

mkpankov:
@jgrund can be merged now

PR #1399, opened 2019-12-12
Fix or silence warnings in iml-agent

mkpankov:
The intent is to have clean compilation to easily spot newly introduced warnings.

With `poller`, the behavior stays the same, since it's unclear how to fix the warnings w/o silencing them.

### DIFF
--- a/iml-agent/src/poller.rs
+++ b/iml-agent/src/poller.rs
@@ -83,6 +83,7 @@ pub async fn create_poller(
                         sessions.reset_empty(&name)
                     });
 
+                    #[allow(unused_must_use)]
                     tokio::spawn(async {
                         fut.await;
                     });
@@ -100,6 +101,7 @@ pub async fn create_poller(
                 now,
             );
 
+            #[allow(unused_must_use)]
             tokio::spawn(async move {
                 fut.await.map_err(|e| {
                     error!("{}", e);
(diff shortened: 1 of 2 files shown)


iml-agent/src/poller.rs
```diff
@@ -100,6 +101,7 @@ pub async fn create_poller(
                 now,
             );
 
+            #[allow(unused_must_use)]
             tokio::spawn(async move {
                 fut.await.map_err(|e| {
```
jgrund:
```rust
if let Err(e) = fut.await {
    error!("{}", e);
}
```

This will allow for the error to be "consumed" (even though it already is since the `map_err` returns unit, and we can avoid adding an annotation.

iml-agent/src/poller.rs
```diff
@@ -83,6 +83,7 @@ pub async fn create_poller(
                         sessions.reset_empty(&name)
                     });
 
+                    #[allow(unused_must_use)]
                     tokio::spawn(async {
                         fut.await;
```
jgrund: `fut.map(drop).await;`

The future is already carrying an `Err(())` at this point. See line 83 above.

iml-agent/src/poller.rs
mkpankov: Done

iml-agent/src/poller.rs
mkpankov: Done

PR #1402, opened 2019-12-13
Implement checking of loaded kernel modules

mkpankov:
This is another shot at point 3 of https://github.com/whamcloud/integrated-manager-for-lustre/issues/1187#issuecomment-559830336

Contrary to https://github.com/whamcloud/integrated-manager-for-lustre/pull/1371 this doesn't bind to native libraries and parses output of `lsmod`.

### DIFF
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -1,0 +1,5 @@
+use crate::agent_error::ImlAgentError;
+
+pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
+    Ok(false)
+}
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -10,21 +10,16 @@ pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
     if output.status.success() {
         let stdout = String::from_utf8_lossy(&output.stdout);
         let mut modules = stdout.lines().into_iter().filter_map(|m| {
-            println!("{:?}", m);
             let mut fields = m.split(' ').filter(|s| *s != "");
             let (name, _, refcount) = (fields.next(), fields.next(), fields.next());
-            println!("{:?} {:?}", name, refcount);
-            if let (Some(name), Some(refcount)) = (name, refcount) {
-                if let Ok(refcount) = refcount.parse::<u32>() {
-                    dbg!(Some((name, refcount)))
-                } else {
-                    dbg!(None)
-                }
+            let refcount = refcount.map(|x| x.parse::<u32>());
+            if let (Some(name), Some(Ok(refcount))) = (name, refcount) {
+                Some((name, refcount))
             } else {
-                dbg!(None)
+                None
             }
         });
-        println!("{:#?}", modules);
+
         let module = modules.find(|(m, refcount)| *m == module && *refcount > 0);
 
         if module.is_some() {
--- a/iml-agent/src/action_plugins/kernel_module.rs
+++ b/iml-agent/src/action_plugins/kernel_module.rs
@@ -9,17 +9,15 @@ pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
 
     if output.status.success() {
         let stdout = String::from_utf8_lossy(&output.stdout);
-        let mut modules = stdout.lines().into_iter().filter_map(|m| {
+        let module = stdout.lines().into_iter().find(|m| {
             let mut fields = m.split(' ').filter(|s| *s != "");
             let name = fields.next();
             if let Some(name) = name {
-                Some(name)
+                name == module
             } else {
-                None
+                false
             }
         });
-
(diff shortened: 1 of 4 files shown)


johnsonw:
Is this still in draft? It looks like you requested review but it's still marked as draft.

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,33 @@
...
+            let (name, _, refcount) = (fields.next(), fields.next(), fields.next());
+            let refcount = refcount.map(|x| x.parse::<u32>());
+            if let (Some(name), Some(Ok(refcount))) = (name, refcount) {
+                Some((name, refcount))
+            } else {
+                None
+            }
+        });
+
+        let module = modules.find(|(m, refcount)| *m == module && *refcount > 0);
```
ip1981: I think refcount is the number of other modules requiring this one, and thus can be 0, but the module is loaded.

ip1981:
Btw, there is /proc/modules which I believe lsmod is reading :)

iml-agent/src/action_plugins/kernel_module.rs
mkpankov: I got info from [RedHat](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/Reference_Guide/s2-proc-modules.html)

> The third column lists how many instances of the module are currently loaded. A value of zero represents an unloaded module.

But now I see [this](https://stackoverflow.com/questions/47645859/meaning-of-modules-instances-in-proc-modules).

So I guess even RedHat misdocuments :man_shrugging: 

Fixing now.

mkpankov:
@johnsonw moved out of draft

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,32 @@
...
+            let mut fields = m.split(' ').filter(|s| *s != "");
+            let name = fields.next();
+            if let Some(name) = name {
+                Some(name)
+            } else {
+                None
+            }
+        });
+
+        let module = modules.find(|m| *m == module);
```
ip1981: I think here you don't need to walk over the list/iterator again. ~~Checking for non-empty is enough, because above you have already filtered out non-matching lines.~~ You can compare with `Some(module)` in `filter_map` in one pass, or even use `find` to make it less than one pass :)

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,32 @@
...
+
+        if module.is_some() {
+            Ok(true)
+        } else {
+            Ok(false)
```
ip1981:
```suggestion
            Ok(module.is_some())
```

iml-agent/src/action_plugins/kernel_module.rs
mkpankov: Yep, done

iml-agent/src/action_plugins/kernel_module.rs
mkpankov: Done

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,26 @@
...
+    if output.status.success() {
+        let stdout = String::from_utf8_lossy(&output.stdout);
+        let module = stdout.lines().into_iter().find(|m| {
+            let mut fields = m.split(' ').filter(|s| *s != "");
+            let name = fields.next();
+            if let Some(name) = name {
+                name == module
+            } else {
+                false
+            }
```
ip1981: I think you could simply compare `name == Some(module)` without `if`.

mkpankov: Ok

ip1981:
BTW, reading  `/proc/modules` might be faster and simpler (no process spawning, single parsing). `/proc/modules` is a public interface, so we could trust it. Also would make possible testing with a sample file (`cat /proc/modules > test.txt`).

There is also one, I guess the fastest way, to check is module is loaded - [/sys/module/MODULENAME](https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-module).

mkpankov:
@ip1981 yeah we can go either of these ways if the need to improve arises

iml-agent/src/action_plugins/kernel_module.rs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::{agent_error::ImlAgentError, cmd::cmd_output};
+
+pub async fn loaded(module: String) -> Result<bool, ImlAgentError> {
+    let output = cmd_output("lsmod", vec![]).await?;
```
utopiabound: This could be `cmd_output_success()`

mkpankov: Done

PR #1407, opened 2019-12-17
Add edition to rustfmt.toml

ip1981:
This is to make rustfmt understand async/.await (and other modern things) when running in Vim with rust.vim. Ref.: https://github.com/rust-lang/rust.vim/issues/368


ip1981:
Hold on, it is probably already solved.

PR #1411, opened 2019-12-18
Integrate Mailbox into docker

johnsonw:
A dedicated docker service needs to be created to represent the mailbox that is used to store large amounts of data that can be streamed.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1411)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -286,6 +286,19 @@ services:
       - "manager-config:/var/lib/chroma"
     environment:
       - RUST_LOG=info
+  iml-mailbox:
+    image: "imlteam/iml-mailbox:5.1"
+    hostname: "iml-mailbox"
+    build:
+      context: ../
+      dockerfile: ./docker/iml-mailbox.dockerfile
+    deploy: *default-deploy
+    volumes:
+      - "mailbox:/var/spool/iml/mailbox"
+      - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
+      - "tmp_vol:/tmp"
+      - "run_vol:/run"
+      - "manager-config:/var/lib/chroma"
   iml-postoffice:
     image: "imlteam/iml-postoffice:5.1"
     hostname: "iml-postoffice"
@@ -315,6 +328,15 @@ volumes:
   grafana-storage:
   static-config1:
   static-config2:
+  mailbox:
+  tmp_vol:
+    driver_opts:
+      type: tmpfs
+      device: tmpfs
+  run_vol:
+    driver_opts:
+      type: tmpfs
+      device: tmpfs
 secrets:
   iml_pw:
     external: true
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -296,8 +296,7 @@ services:
     volumes:
       - "mailbox:/var/spool/iml/mailbox"
       - "/sys/fs/cgroup:/sys/fs/cgroup:ro"
-      - "tmp_vol:/tmp"
-      - "run_vol:/run"
+      - "mailbox_run:/run"
       - "manager-config:/var/lib/chroma"
   iml-postoffice:
     image: "imlteam/iml-postoffice:5.1"
@@ -329,11 +328,7 @@ volumes:
   static-config1:
   static-config2:
   mailbox:
-  tmp_vol:
-    driver_opts:
-      type: tmpfs
-      device: tmpfs
-  run_vol:
+  mailbox_run:
     driver_opts:
       type: tmpfs
       device: tmpfs
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -86,6 +86,7 @@ services:
     deploy: *default-deploy
     volumes:
       - "manager-config:/var/lib/chroma"
+      - "mailbox:/var/spool/iml/mailbox"
       - "static-config1:/usr/lib/iml-manager"
       - "static-config2:/usr/lib/node_modules/@iml"
     ports:
@@ -176,6 +177,7 @@ services:
     deploy: *default-deploy
     volumes:
       - "manager-config:/var/lib/chroma"
+      - "mailbox:/var/spool/iml/mailbox"
     environment:
       - PROXY_HOST=iml-action-runner
       - INFLUXDB_SERVER_FQDN=influxdb
--- a/docker/iml-mailbox.dockerfile
+++ b/docker/iml-mailbox.dockerfile
@@ -1,0 +1,17 @@
+FROM rust-iml-base as builder
+
+FROM centos:8
+COPY --from=builder /build/target/release/iml-mailbox /bin/
+COPY docker/iml-mailbox/iml-mailbox.service /etc/systemd/system/
+COPY docker/iml-mailbox/iml-mailbox.conf /etc/systemd/system/
+COPY iml-mailbox.conf /usr/lib/tmpfiles.d/
+COPY docker/wait-for-dependencies.sh /usr/local/bin/
+
+ENV container docker
+STOPSIGNAL SIGRTMIN+3
+
+RUN systemctl enable iml-mailbox
+
+ENTRYPOINT ["wait-for-dependencies.sh"]
+
+CMD ["/sbin/init"]
--- a/docker/iml-mailbox.dockerfile
+++ b/docker/iml-mailbox.dockerfile
@@ -1,17 +1,14 @@
 FROM rust-iml-base as builder
 
-FROM centos:8
+FROM imlteam/systemd-base
 COPY --from=builder /build/target/release/iml-mailbox /bin/
 COPY docker/iml-mailbox/iml-mailbox.service /etc/systemd/system/
 COPY docker/iml-mailbox/iml-mailbox.conf /etc/systemd/system/
 COPY iml-mailbox.conf /usr/lib/tmpfiles.d/
 COPY docker/wait-for-dependencies.sh /usr/local/bin/
 
-ENV container docker
-STOPSIGNAL SIGRTMIN+3
-
 RUN systemctl enable iml-mailbox
 
 ENTRYPOINT ["wait-for-dependencies.sh"]
 
-CMD ["/sbin/init"]
+CMD ["/usr/sbin/init"]
--- a/docker/iml-mailbox/iml-mailbox.service
+++ b/docker/iml-mailbox/iml-mailbox.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=IML Mailbox Service
+After=syslog.target network.target
+
+
+[Service]
+Type=simple
+Environment=RUST_LOG=info
+EnvironmentFile=/var/lib/chroma/iml-settings.conf
+EnvironmentFile=/etc/systemd/system/iml-mailbox.conf
+ExecStart=/bin/iml-mailbox
+Restart=always
+StandardOutput=journal
+StandardError=journal
+
+[Install]
+WantedBy=multi-user.target
--- a/docker/iml-mailbox/iml-mailbox.conf
+++ b/docker/iml-mailbox/iml-mailbox.conf
@@ -1,0 +1,1 @@
+PROXY_HOST=iml-mailbox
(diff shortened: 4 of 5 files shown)


docker/docker-compose.yml
```diff
@@ -277,6 +277,19 @@ services:
       - "manager-config:/var/lib/chroma"
     environment:
       - RUST_LOG=info
+  iml-mailbox:
+    image: "imlteam/iml-mailbox:dev"
```
jgrund: Remove dev tag.

docker/docker-compose.yml
```diff
@@ -277,6 +277,19 @@ services:
...
+    hostname: "iml-mailbox"
+    build:
+      context: ../
+      dockerfile: ./docker/iml-mailbox.dockerfile
+    deploy: *default-deploy
+    volumes:
+      - "mailbox:/var/spool/iml/mailbox"
```
jgrund: Will these files ever age out? In RPM based installs I think they are removed after 30 days.

johnsonw: Looking into this.

johnsonw: I don't see anything specific on aging files out on volumes. How is this done on the rpm version?

jgrund: https://github.com/whamcloud/integrated-manager-for-lustre/blob/ba3fd1aa6328431486367bdfc8451a1348b57656/iml-mailbox.conf

johnsonw:
I added a systemd override file to replace certain values in the service files but unfortunately, some of the items are dependencies and cannot be reset in an overrides config file. Scroll to the bottom of the following systemd man page to see an explanation: https://www.freedesktop.org/software/systemd/man/systemd.unit.html

docker/docker-compose.yml
```diff
@@ -330,6 +330,14 @@ volumes:
   static-config1:
   static-config2:
   mailbox:
+  tmp_vol:
```
jgrund: Why are these needed?

johnsonw:
I ran a test to verify that the temporary file I added to /var/spool/iml/mailbox would be removed when the systemd-tmpfiles-clean.timer executes. I updated the configuration to age out at 0. Here is some sample output:

```
[root@iml-mailbox mailbox]# ls
test.txt
[root@iml-mailbox mailbox]# systemctl status systemd-tmpfiles-clean.timer
● systemd-tmpfiles-clean.timer - Daily Cleanup of Temporary Directories
   Loaded: loaded (/usr/lib/systemd/system/systemd-tmpfiles-clean.timer; static; vendor preset: disabled)
   Active: active (waiting) since Tue 2020-01-28 12:55:58 UTC; 10min ago
  Trigger: Tue 2020-01-28 13:10:57 UTC; 4min 1s left
     Docs: man:tmpfiles.d(5)
           man:systemd-tmpfiles(8)

Jan 28 12:55:58 iml-mailbox systemd[1]: Started Daily Cleanup of Temporary Directories.

# ... after 5 minutes

[root@iml-mailbox mailbox]# ls
[root@iml-mailbox mailbox]# pwd
/var/spool/iml/mailbox
```

docker/docker-compose.yml
johnsonw: Now that this is using systemd I ran a test to verify that the file would age out and confirmed that it was removed. Note that I set the Age out value to be 0 so that as soon as the systemd-tmpfiles-clean.timer executed the file would be removed.

docker/docker-compose.yml
johnsonw: I was under the impression that systemd may require /run and /tmp to be a tmpfs filesystem. I'll try removing these and verify that it still works.

johnsonw:
To show that the config is loaded:
```
systemd-tmpfiles --cat-config

# /usr/lib/tmpfiles.d/iml-mailbox.conf
# See tmpfiles.d(5) for details

d /var/spool/iml/mailbox - - - 0
```

docker/docker-compose.yml
johnsonw: To follow up, I took out the tmp_vol and tmp_run volumes, cleared all docker volumes, networks, etc and deployed the stack. I connected to the iml-mailbox container and checked the status of the service and received the following:
```
[root@iml-mailbox /]# systemctl status iml-mailbox
System has not been booted with systemd as init system (PID 1). Can't operate.
Failed to connect to bus: Host is down
```
I then removed the stack and cleared out the volumes, added the tmp volumes back to the docker-compose file and re-deployed the stack. Checking the status of the mailbox service revealed:

```
[root@iml-mailbox /]# systemctl status iml-mailbox
● iml-mailbox.service - IML Mailbox Service
   Loaded: loaded (/etc/systemd/system/iml-mailbox.service; enabled; vendor preset: disabled)
   Active: active (running) since Tue 2020-01-28 14:08:07 UTC; 16s ago
 Main PID: 140 (iml-mailbox)
    Tasks: 1 (limit: 78642)
   Memory: 1.4M
   CGroup: /docker/c0edb9328280642d18a47a6d7618fa5cbe4baf63812038944f7fcf34942e8128/system.slice/iml-mailbox.service
           └─140 /bin/iml-mailbox

Jan 28 14:08:07 iml-mailbox systemd[1]: Started IML Mailbox Service.
Jan 28 14:08:07 iml-mailbox iml-mailbox[140]: Jan 28 14:08:07.748  INFO iml_mailbox: Starting on V4(10.0.46.21:8891)
```

docker/docker-compose.yml
jgrund: How often does `systemd-tmpfiles-clean.timer` fire? Do we need to persist any metadata about it outside the container?

docker/iml-mailbox.dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM rust-iml-base as builder
+
+FROM imlteam/systemd-base
+COPY --from=builder /build/target/release/iml-mailbox /bin/
+COPY docker/iml-mailbox/iml-mailbox.service /etc/systemd/system/
+COPY docker/iml-mailbox/iml-mailbox.conf /etc/systemd/system/
+COPY iml-mailbox.conf /usr/lib/tmpfiles.d/
+COPY docker/wait-for-dependencies.sh /usr/local/bin/
+
+RUN systemctl enable iml-mailbox
```
ip1981: It does not require systemd daemon running etc., does it?

docker/iml-mailbox/iml-mailbox.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=IML Mailbox Service
+After=syslog.target network.target
+
+
+[Service]
+Type=simple
+Environment=RUST_LOG=info
+EnvironmentFile=/var/lib/chroma/iml-settings.conf
+EnvironmentFile=/etc/systemd/system/iml-mailbox.conf
```
jgrund: Hrrm, I don't think you can have multiple `EnvironmentFile` directives.

jgrund: nm, you can.

ip1981: Apparently [it is ok](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=):

> This option may be specified more than once in which case all specified files are read. If the empty string is assigned to this option, the list of file to read is reset, all prior assignments have no effect.

jgrund:
Just a question if we need to persist any data outside the container to help `systemd-tmpfiles-clean.timer`, or if it will work the same regardless

docker/iml-mailbox.dockerfile
johnsonw: It's available to enable the service at this point. It's actually really neat that we can do this inside a container. Here's the link for more info in case you're interested: https://developers.redhat.com/blog/2016/09/13/running-systemd-in-a-non-privileged-container/

docker/iml-mailbox/iml-mailbox.service
johnsonw: Yup, this threw me off at first. So the problem here is that we load the iml-settings file to load the environment variables. But we need to override the PROXY_HOST variable. My first thought was to just set an environment variable here in the service file thinking that it would override the value that was loaded from iml-settings.conf. But this doesn't work. After doing some research, I found an open issue for systemd in which others were trying to do the same thing and couldn't override the environment variables loaded from the EnvironmentFile directive. It turns out, systemd only lets you override a variable that's loaded from an environment file by loading an additional environment file with a variable containing the same name. Seems like a poor design in my opinion, but that appears to be the way to do it. It's easy to verify that this works because when the service loads it will indicate the IP and port in which the service is running. It will display the IP for the iml-timer domain but if we removed the EnvironmentFile for iml-mailbox.conf and replaced it with a standard environment variable you would see 127.0.0.1.

johnsonw:
If we do need to persist anything it would be the `/usr/lib/tmpfiles.d/` directory. I'm not convinced we need to do this though as this is baked into the container. I'm going to remove and then redeploy the stack and make sure that the timer still knows about it. I'll post my results shortly.

johnsonw:
Looks like we shouldn't need to add a volume for the tmpfile. I removed the stack and then re-deployed. Checking systemd-tmpfiles I see the following:

```
systemd-tmpfiles --cat-config
...
# /usr/lib/tmpfiles.d/iml-mailbox.conf
# See tmpfiles.d(5) for details

d /var/spool/iml/mailbox - - - 15d
...
```

jgrund:
My question was more about the actual timer trigger than than the config. I.E:

```
[root@iml-mailbox mailbox]# systemctl status systemd-tmpfiles-clean.timer
● systemd-tmpfiles-clean.timer - Daily Cleanup of Temporary Directories
   Loaded: loaded (/usr/lib/systemd/system/systemd-tmpfiles-clean.timer; static; vendor preset: disabled)
   Active: active (waiting) since Tue 2020-01-28 12:55:58 UTC; 10min ago
  Trigger: Tue 2020-01-28 13:10:57 UTC; 4min 1s left
     Docs: man:tmpfiles.d(5)
           man:systemd-tmpfiles(8)
```

Does this bit `Trigger: Tue 2020-01-28 13:10:57 UTC; 4min 1s left` need to persist between restarts since it looks like a daily cleanup.

johnsonw:
Sorry, I misunderstood your question. Let me make sure I understand. I need to check that, if for instance, the trigger is set to go off in 4 minutes and then suddenly the stack is removed. We then bring the stack back up and now I need to verify that the trigger value is still set and didn't reset to say, 24 hours. Am I understanding correctly?

johnsonw:
Just to follow up the systemd-tmpfiles-clean.timer unit looks like this:
```
[Unit]
Description=Daily Cleanup of Temporary Directories
Documentation=man:tmpfiles.d(5) man:systemd-tmpfiles(8)

[Timer]
OnBootSec=15min
OnUnitActiveSec=1d
```

**OnBootSec** defines a timer relative to when the machine was booted up.
**OnUnitActiveSec** defines a timer relative to when the unit the timer is activating was last activated.

If I understand this correctly, this would mean that it would trigger after the first 15 minutes from boot. It would then trigger again 1 day after the 15 minutes. This would mean that if we removed the stack and redeployed it would trigger 15 minutes after that. It would then trigger every day after that moment. If we wanted to, we could override the OnUnitActiveSec property by creating a new file in /etc/systemd/etc, which we already have a volume for. Does this answer your question?

jgrund:
> I need to verify that the trigger value is still set and didn't reset to say, 24 hours. Am I understanding correctly?

Yes.

jgrund:
> Just to follow up the systemd-tmpfiles-clean.timer unit looks like this:
> 
> ```
> [Unit]
> Description=Daily Cleanup of Temporary Directories
> Documentation=man:tmpfiles.d(5) man:systemd-tmpfiles(8)
> 
> [Timer]
> OnBootSec=15min
> OnUnitActiveSec=1d
> ```
> 
> **OnBootSec** defines a timer relative to when the machine was booted up.
> **OnUnitActiveSec** defines a timer relative to when the unit the timer is activating was last activated.
> 
> If I understand this correctly, this would mean that it would trigger after the first 15 minutes from boot. It would then trigger again 1 day after the 15 minutes. This would mean that if we removed the stack and redeployed it would trigger 15 minutes after that. It would then trigger every day after that moment. If we wanted to, we could override the OnUnitActiveSec property by creating a new file in /etc/systemd/etc, which we already have a volume for. Does this answer your question?

Yes, `OnBootSec=15min` means we don't need to persist an active timer

johnsonw:
Note that I did a quick reset of the stack and the trigger appears to be in 23 hours. I'll dig into the OnBootSec property real quick in the man pages.

PR #1415, opened 2019-12-20
Fix dependencies of subcrates

ip1981:
So that individual crates can be built.

I suppose that when building the entire workspace cargo chooses the most
wide set of features and use it to build all members. And thus some crates
are buildable by accident, but they should list their dependencies explicitly.


ip1981:
I used this to test:

```
for f in iml-*/Cargo.toml; do (cd $(dirname $f) && pwd  && cargo build && cargo test ) || break; done
```

Maybe it's worth a CI job.

jgrund:
> Maybe it's worth a CI job.

Probably. You can add another test to the github actions

PR #1417, opened 2019-12-20
Trim down docker image size

jgrund:

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,15 @@
-.github
+**/target
+**/node_modules
+**/.copr
+**/.github
+**/.vscode
+**/.git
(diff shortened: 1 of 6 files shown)


ip1981:
I wonder if it could be possible to use white listing instead of black listing (.dockerignore).

jgrund:
It looks like `*` followed by a number of `!...` lines can be used to simulate whitelisting.

jgrund:
I think for now we can stick to a blacklist as this is basically working.

jgrund:
> What are the numbers before and after?

I'll post after it publishes to docker-hub, but I think it was ~ 80MB.

PR #1427, opened 2019-12-23
Implement tree-view component

### LINKED ISSUE #1349: Implement tree-view component in iml-gui
Implement the tree-view component in iml-gui.

Open question is how to use items populated from the warp-drive cache. Should we pass the cache into the update fn for tree-view and patch up the tree-view model?

Should we forward individual cache changes to the relevant update fns so they can then patch up their internal representations?

jgrund:
Fixes #1349.

This patch implements the tree-view in the iml-gui crate.

After evaluating holding refs to records through `Arc` + `RwLock`, the
approach being taken is to hold pointers to records through ids that
implement `Copy`.

treeview-v2


jgrund:
This patch is functional sans-updating records. I'm thinking a good approach is to hold indexes that point from `RecordId` -> `TreeId` and from `TreeId` to `RecordId`. 

This will mean we do a full tree-population on load, but also means we don't figure out tree children on each render.

jgrund:
This now shows any alerts associated with relevant items:

Screen Shot 2019-12-30 at 10 24 50 AM

PR #1429, opened 2019-12-25
Fix Django warnings

mkpankov:
```
/root/iml/chroma_api/session.py:134: RemovedInDjango20Warning: Using user.is_authenticated() and user.is_anonymous() as a method is deprecated. Remove the parentheses to use it as an attribute
```


ip1981:
There was a PR for that, and it didn't work out.

PR #1430, opened 2019-12-25
DDN test task Starostin  (Issue 1421: Implement --prompt / --force flags for adding new servers)

Basil22:
There are my code for this issue/task.

I didn't invent "y/n" dialog and used existing crate "clt". However, it doesn't compile under Windows. But this crate status is green so I expect it will compile successfully at your Unix machines. Please tell me if not and I will implement the dialog manually.


iml-manager-cli/src/server.rs
```diff
@@ -31,6 +31,12 @@ pub struct AddHosts {
     /// The profile to deploy to each host
     #[structopt(short = "p", long = "profile")]
     profile: String,
+    /// Prompt to continue if command fails
+    #[structopt(long)]
+    prompt: bool,
+    /// Always continue if command fails
+    #[structopt(long)]
+    force: bool,
```
ip1981: It looks line the new options (`--force/--prompt`) are not mutually exclusive.

iml-manager-cli/src/server.rs
```diff
@@ -95,6 +101,15 @@ fn filter_known_hosts<'a, 'b>(
         .collect()
 }
 
+fn can_continue(config: &AddHosts, err_msg: &str) -> bool {
+    display_error(&err_msg);
+    config.force || (config.prompt && get_confirm())
+}
+
+fn get_confirm() -> bool {
+    clt::confirm("Continue deployment?", true, "> ", true)
```
ip1981: Maybe it's not worth a separate function.

Basil22: This function expected to be long but much thanks to clt crate :) I like separate variant because it makes can_continue() to have pure logic only.

iml-manager-cli/src/server.rs
Basil22: You are right. Structopt doesn't have mutex options, at least in official documentation. I could add a manual check, but it splits parameter checking and reporting in two places.

iml-manager-cli/Cargo.toml
```diff
@@ -24,6 +24,7 @@ tokio = { version = "0.2", features = ["macros"] }
 tracing = "0.1"
 tracing-subscriber = "0.1"
 number-formatter = { path = "../number-formatter", version = "0.1" }
+clt = "0.0.6"
```
mkpankov: I don't like adding a dependency that's wasn't updated for more than a year and isn't maintained at this point.

Given that there's just one function needed, I propose implementing it in our code.

iml-manager-cli/src/server.rs
mkpankov: I think manual check should be added, specifying `--force` and `--prompt` makes no sense. It could happen by accident, need to protect against that.

ip1981: In general, the last seen option should override any previous.

nlinker: I'd make an error first, if some conflicting options provided. The lesser implicit behavior - the earlier errors can be detected, hence the easier to maintain the things.

ip1981: I think the ideal could be something like `--on-failure=(ask|force|stop)` :)

iml-manager-cli/Cargo.toml
jgrund: We'll probably need more of these in the future. 

I'd suggest using https://docs.rs/dialoguer/0.5.0/dialoguer/ instead, which similar to `console` does not have a dependency on termion (it actually depends on `console`) which means we can still compile on windows.

iml-manager-cli/src/server.rs
jgrund: Good point; since the options are mutually exclusive, let's go with an enum approach which makes multiple selections impossible. 

Like @ip1981 said, if we go with something like:

`--on-failure=(prompt|force|)` we will catch either case. I don't think we need stop as not specifying the flag at all can mean stop.

iml-manager-cli/src/server.rs
```diff
@@ -349,7 +365,8 @@ pub async fn server_cli(command: ServerCommand) -> Result<(), ImlManagerCliError
             let objects: Vec<_> = objects
                 .into_iter()
                 .filter_map(|x| x.host_profiles)
-                .filter(|x| is_profile_valid(&x, &profile.name))
+                .filter(|x| is_profile_valid(&x, &profile.name) ||
```
jgrund: I think we'll need to split this into two phases.

The first phase is simply checking that the profile selected by the user exists in the objects being iterated over. If the profile does not exist in these objects, then we want to return early; there is no point in overriding / prompting if we can not deploy the profile we want.

The second phase will check if all the profiles checks have passed. If they have not passed, this is where we want to use the prompt / override logic.

For the first phase, if we use `find`, we can pick out the single profile we're interested in. 

Then we can look at that profile and check it's items to see if they all pass or not. If we continue forward we can then convert it into a `HostProfileConfig` and continue onwards.

jgrund:
Also be sure to commit changes to `Cargo.lock`

iml-manager-cli/src/server.rs
Basil22: Good idea, I'll try to implement om weekend.

iml-manager-cli/Cargo.toml
Basil22: Ok, I'll switch to dialoguer.

iml-manager-cli/src/server.rs
ip1981: I was thinking about "stop" (which would be the default) for completeness, so that there could be a way to override anything (even back to the default). 
Maybe implementation can be simpler too.

iml-manager-cli/src/server.rs
Basil22: 1. If no profile then always filter out object, right?
2. Do you mean use ```iter().find()``` instead of ```get()```? Maybe ```contains_key()```?

iml-manager-cli/src/server.rs
Basil22: I will have to implement Stop, at least in the enum, because I need a state for current behaviour. Implementation as cli parameter will depend on structopt capabilities.

Basil22: Well, I have found very simple solution. Structopt can directly use clap methods, and clap surely has mutex options. So I just use clap from structopt directive.

Basil22: The result is "error: The argument '--force' cannot be used with '--prompt'" which is exactly what we want.

Basil22:
I have pushed the update, it should fix all the issues: mutex options, y/n dialog, phased filtering, ftm, lock.

The concerns are:
1. Please make sure that I implemented ```is_profile_exist``` correctly, as well as new filtering combination.
2. I used '''unwrap_or``` in ```get_confirm``` to keep logic easy. May be it is worth to use ```Result``` (I think it isn't).

iml-manager-cli/src/server.rs
```diff
@@ -31,6 +32,12 @@ pub struct AddHosts {
     /// The profile to deploy to each host
     #[structopt(short = "p", long = "profile")]
     profile: String,
+    /// Prompt to continue if command fails
+    #[structopt(long, conflicts_with = "force")]
```
jgrund: neat

iml-manager-cli/src/server.rs
```diff
@@ -349,7 +375,11 @@ pub async fn server_cli(command: ServerCommand) -> Result<(), ImlManagerCliError
             let objects: Vec<_> = objects
                 .into_iter()
                 .filter_map(|x| x.host_profiles)
-                .filter(|x| is_profile_valid(&x, &profile.name))
+                .filter(|x| is_profile_exist(&x, &profile.name))
```
jgrund: This should be a hard error. If the profile doesn't exist, we should print an error and return. Something like:

"Profile {} not found".

This means using `find` at this point and handling the `Some` / `None`  variants.

Basil22: So if one of the objects doesn't have a profile, we stop the whole processing, right?

jgrund: Right If we can't find the profile we are looking for in this iterator, we return early with an error.

Basil22: So each element in objects must have a profile to start the next step?

jgrund: There should only be one match, and yes this is different from current logic where we filtered non-matches previously

Basil22: Ok. New Year is a holiday for my family so I'll make a patch by the 2nd.

Basil22: I have put an updated version. 
Please notice that the logic has been changed so it needs to be checked again.

iml-manager-cli/src/server.rs
```diff
@@ -95,6 +102,24 @@ fn filter_known_hosts<'a, 'b>(
...
+fn get_confirm() -> bool {
+    Confirmation::new()
+        .with_text("Continue deployment?")
+        .default(true)
+        .show_default(true)
+        .interact()
+        .unwrap_or(false)
+}
+
+fn is_profile_exist(x: &HostProfile, profile_name: &str) -> bool {
```
jgrund:
```suggestion
fn does_profile_exist(x: &HostProfile, profile_name: &str) -> bool {
```

jgrund:
Screen Shot 2020-01-03 at 5 56 25 PM

PR #1431, opened 2019-12-27
Reenable job_scheduler/host tests

mkpankov:

### DIFF
--- a/tests/unit/chroma_core/helpers/mock_agent_rpc.py
+++ b/tests/unit/chroma_core/helpers/mock_agent_rpc.py
@@ -106,7 +106,7 @@ class MockAgentRpc(object):
     @classmethod
     def _call(cls, host, cmd, args):
         cls.calls.append((cmd, args))
-        cls.host_calls[host].append((cmd, args))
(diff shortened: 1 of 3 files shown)


mkpankov:
So here's our failure I found locally:
https://travis-ci.com/whamcloud/integrated-manager-for-lustre/jobs/270667248#L2962

Any pointers to where to fix that?

ip1981:
> Empty file has been added only?
> How does it fix the issue?

It's Python :)

ip1981:
I think the problem is somewhere around here => https://github.com/whamcloud/integrated-manager-for-lustre/blob/25b89153a863cbee7527ce01e742ca14c91cdbda/chroma_core/models/host.py#L72

The `hosts` argument is passed to `__init__`, it does not accept it.

mkpankov:
@nlinker `__init__.py` marks a "regular package", somewhat akin to `mod.rs`. [Details](https://stackoverflow.com/a/448279)

jgrund:
There were actually a number of errors in these tests. Are all fixed now.

PR #1433, opened 2019-12-31
Create `lamigo` service unit

nlinker:
Fixes https://github.com/whamcloud/integrated-manager-for-lustre/issues/1379

There are two additional parameters I have added to the original task, since Nate @utopiabound asked to check mount (and do mount if not mounted):
- LNet device on the client side, like `192.168.0.100@tcp0:/spfs` to be able to execute `mount` command like `$ mount -t lustre 192.168.0.100@tcp0:/spfs /mnt/lustre`
- And `force` flag to ignore mounts and checks and just create the unit.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433)


iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,180 @@
...
+        .arg(c.lustre_device.clone())
+        .arg(c.mount_point.clone())
+        .spawn()?;
+    let output: Output = process.wait_with_output().await?;
+    match output.status.success() {
+        true => Ok(()),
+        false => Err(io::Error::new(
+            io::ErrorKind::Other,
+            format!(
+                "Cannot execute command \'mount -t lustre {} {}\'",
```
ip1981: `Command` has the `Debug` trait, maybe use it?

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,180 @@
...
+    if c.force {
+        create_lamigo_service_unit_internal(SYSTEMD_DIR, &c)
+            .err_into()
+            .await
+    } else {
+        let is_mounted = is_filesystem_mounted(&c.mount_point).await?;
+        if !is_mounted {
+            mount_filesystem(&c).await?;
+        }
+        create_lamigo_service_unit_internal(SYSTEMD_DIR, &c).await?;
```
ip1981: This line looks duplicating `create_lamigo_service_unit_internal(SYSTEMD_DIR, &c)` above. Could it be something like this:

```rust
if !c.force {
  // mount stuff here
}
create_lamigo_service_unit_internal(SYSTEMD_DIR, &c).await?;
```

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,180 @@
...
+
+use crate::agent_error::ImlAgentError;
+use futures::future::TryFutureExt;
+use std::path::Path;
+use std::process::Output;
+use std::{fmt, io};
+use tokio::fs;
+use tokio::process::{Child, Command};
+
+static SYSTEMD_DIR: &str = "/Users/nick/tmp/system";
```
ip1981: oops :)

nlinker: Good eye!

iml-agent/src/action_plugins/lamigo.rs
nlinker: Yes, I tried to explicitly state that "here the things we do when force is true, here is otherwise". I'm ok with your proposal.

nlinker:
> I have some questions:
> 
> 1. I don't understand why mounting ("a run-time dependency") is considered when creating a systemd unit ("build time"). What is expected when the unit is created and FS is unmounted after then?
> 2. If there is a run-time dependency on a mount point, does it make sense to use systemd mechanism for such dependency?
> 3. What is lamigo's user? Does it make sense to start `lamigo` service as `user` via systemd (and not as root)?

1. @utopiabound has the following requirement for this:
> A start action should be to mount the filesystem if it is not mounted already.

So this requirement has really complicated stuff and pulls here the "run-time dependency". 
There is an alternative solution to separate this module into the 3 ones: CheckMount, DoMount and CreateLamigoServiceUnit, however I really don't like this, since it creates implicit dependencies between them (i.e. they are not independent anymore).

I'm open to other proposals how to implement this requirement.

2. There is systemd mount dependency, I've made a research on it is good or not to convert this "run-time" into "build-time", however there are gotchas with systemd on this part: https://github.com/systemd/systemd/issues/4468
So I think doing this in Rust is easier at least.

3. It is the special user, a so called "changelog user", that has an access to MDT records and is able to manipulate them (from the lamigo source code, _user on which behalf lamigo consumes changelogs from the MDT device_). It is the different one from the user, under which the `lamigo` process runs, as I understand.

iml-agent/src/action_plugins/lamigo.rs
nlinker: Ah, this is easy, thanks!

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,175 @@
...
+            hot_pool = self.hot_pool,
+            min_age = self.min_age,
+            user = self.user,
+            mount_point = self.mount_point,
+        )
+    }
+}
+
+pub async fn create_lamigo_service_unit(c: Config) -> Result<(), ImlAgentError> {
+    if !c.force {
```
jgrund: I don't think we should correlate different actions together. 

By keeping actions single-purpose, we make them reusable in different contexts.

It's trivial to perform a mount as a separate action from either the manager side / a CLI call / an agent invocation / or a resource agent.

It's also not clear if the work is being duplicated by the eventual resource agent being developed separately.

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,175 @@
...
+    // we need to execute mount command
+    //   mount -t lustre \
+    //     [-o <options> ] \
+    //     <MGS NID>[:<MGS NID>]:/<fsname> \
+    //     /lustre/<fsname>
+    let mut command = Command::new("mount");
+    let process: Child = command
+        .arg("-t")
+        .arg("lustre")
+        .arg(c.lustre_device.clone())
```
mkpankov: `.clone` isn't necessary:

```rust
        .arg(&c.lustre_device)
        .arg(&c.mount_point)
```

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,175 @@
...
+    async fn test_works() {
+        // cargo run --package iml-agent --bin iml-agent -- lamigo
+        // --cold_pool COLD_POOL --hot_pool HOT_POOL
+        // --fs LU_TEST --lustre_device 192.168.0.100@tcp0:/spfs
+        // --mdt MDT000 --min_age 10000 --user nick --mount_point /mnt/lustre
+        let config = Config {
+            fs: "lu_test".into(),
+            mdt: "lustre-MDT0000".into(),
+            user: "cl1".into(),
+            min_age: 35353,
```
mkpankov: `min_age` in the comment above says `10000`, why the difference?

mkpankov:
It seems my comments weren't answered or addressed

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 105 at r4](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r4-105:-LyVDHWn7X1ogbDlcYqk:bpzq5k5) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/9dd26350e080d85d4cd97fc30d6f65611ac70f78/iml-agent/src/action_plugins/lamigo.rs#L105)):*

`.clone` isn't necessary:

```rust
        .arg(&c.lustre_device)
        .arg(&c.mount_point)
```
</blockquote></details>

Thanks, Michael

iml-agent/src/action_plugins/lamigo.rs
nlinker: Agree, the comments and the code must correspond.

ip1981: > Agree, the comments and the code must correspond.

Or comments should not exist ;-)

nlinker:
___
*[iml-agent/systemd-units/rust-iml-agent.service, line 17 at r6](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-LyZiIA431SNu4YcEifC:-Ly_OAcc2crtak7nzLF6:b-cusy01) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/1cc9b964065ebea92b715497cabc8a980724ff01/iml-agent/systemd-units/rust-iml-agent.service#L17)):*

Hardcoding local path doesn't seem right
</blockquote></details>

Yes, I noticed it already, but thank you anyways!

iml-agent/Cargo.toml
```diff
@@ -34,6 +34,7 @@ url = "2.1"
 v_hist = { git = "https://github.com/jgrund/hist", branch = "use-console-instead-of-termion" }
 uuid = { version = "0.7", features = ["v4"] }
 parking_lot = "0.9"
+strfmt = "0.1.6"
```
nlinker: My note:
This is to make `format!` with dynamic format string. It is pretty popular, has all the error constructors public and has no dependencies. Dynamic format is needed to expand `/etc/systemd/system/lamigo-{fs}-{mdt}.service` string.

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -36,13 +36,16 @@ pub fn create_registry() -> action_plugins::Actions {
         .add_plugin("action_check_ha", check_ha::check_ha)
         .add_plugin("action_check_stonith", check_stonith::check_stonith)
         .add_plugin("get_kernel", check_kernel::get_kernel)
-        .add_plugin("lctl", lctl::lctl)
+        .add_plugin("is_mounted", lustre::is_mounted)
+        .add_plugin("try_mount", lustre::try_mount)
+        .add_plugin("lctl", lustre::lctl)
```
nlinker: `lctl.rs` was renamed to `lustre.rs`

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,177 @@
...
+            ))
+        );
+    }
+
+    #[tokio::test]
+    async fn test_works() {
+        // cargo run --package iml-agent --bin iml-agent -- lamigo
+        // --cold_pool SLOW_POOL --hot_pool FAST_POOL
+        // --fs LU_TEST2 --lustre_device 192.168.0.100@tcp0:/spfs
+        // --mdt MDT000 --min_age 35353 --user nick --mount_point /mnt/lustre
```
nlinker: This comment block is to help me (and maybe some others) to run cli and compare with the local test results. It could be removed of course, however it will require _significant time_ to properly rebuild it with all the options instead of just copy-n-paste

iml-agent/systemd-units/rust-iml-agent.service
```diff
@@ -13,6 +13,8 @@ Environment=CRT_PATH=/etc/iml/self.crt
 Environment=PFX_PATH=/etc/iml/identity.pfx
 Environment=AUTHORITY_CRT_PATH=/etc/iml/authority.crt
 Environment=LTUER_CONF_PATH=/etc/iml/ltuer.conf
+# Here {name} corresponds to the field names of iml_agent::action_plugins::lamigo::Config
+Environment=LAMIGO_UNIT_PATH=/etc/systemd/system/lamigo-{fs}-{mdt}.service
```
ip1981: It will be easy to break by accident. And since this path is for systemd and not for IML, I don't think it should be here. @mpankov

iml-agent/src/agent_error.rs
```diff
@@ -131,6 +132,7 @@ impl std::fmt::Display for ImlAgentError {
             ImlAgentError::InvalidHeaderValue(ref err) => write!(f, "{}", err),
             ImlAgentError::NativeTls(ref err) => write!(f, "{}", err),
             ImlAgentError::XmlError(ref err) => write!(f, "{}", err),
+            ImlAgentError::FmtError(ref err) => write!(f, "{}", err),
```
nlinker: This is caused by `strfmt` library, the new dependency

iml-agent/systemd-units/rust-iml-agent.service
nlinker: From the Michael's proposal to make the path to be the part of the environment.

I've decided not to do just `LAMIGO_DIR=/etc/systemd/system`, since it is one of the standard systemd locations and to have it configurable looks redundant.
Additionally to this, we might want to configure the name of the unit itself. Therefore the full path is configurable and the comment above.

nlinker: Yep, let's discuss it, if it feels not right.

iml-agent/src/cli.rs
```diff
@@ -208,6 +209,7 @@ pub enum StratagemClientCommand {
 
 #[derive(StructOpt, Debug)]
 #[structopt(name = "iml-agent")]
+#[structopt(raw(setting = "structopt::clap::AppSettings::ColoredHelp"))]
```
nlinker: Because it looks prettier :-)

iml-agent/src/cli.rs
```diff
@@ -489,7 +515,27 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         }
         App::GetKernel { modules } => match check_kernel::get_kernel(modules).await {
             Ok(s) => println!("{}", s),
-            Err(e) => println!("{:?}", e),
+            Err(e) => {
+                eprintln!("{:?}", e);
+                exit(exitcode::SOFTWARE);
```
nlinker: Added failure for `GetKernel`

iml-agent/src/cli.rs
```diff
@@ -489,7 +515,27 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            }
+        },
+        App::TryMount {
+            lustre_device,
+            mount_point,
+        } => match lustre::try_mount((lustre_device, mount_point)).await {
+            Ok(()) => println!("Done"),
```
nlinker: Since `try_mount` returns unit `()`, so I'm printing "Done" just to print something...

jgrund:
Looks like the latest push has merge conflicts.

nlinker:
___
*[iml-agent/Cargo.toml, line 37 at r8](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-LybNoL5E61ONHYrjVvQ:-LybNoL5E61ONHYrjVvR:blqqt0h) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/7963de5706d115d94d4ef62dbd163e979a002842/iml-agent/Cargo.toml#L37)):*
> ```TOML
> serde_json = "1"
> spinners = "1.2"
> strfmt = "0.1.6"
> ```

This is to make format! with dynamic format string. It is pretty popular, has all the error constructors public and has no dependencies. Dynamic format is needed to expand `/etc/systemd/system/lamigo-{fs}-{mdt}.service string.`

nlinker:
___
*[iml-agent/systemd-units/settings.conf, line 10 at r8](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-LybOj1E98Gq9It4VDHg:-LybOj1FDRkwnF-gJpoO:b-z9q3ia) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/7963de5706d115d94d4ef62dbd163e979a002842/iml-agent/systemd-units/settings.conf#L10)):*
> ```
> LTUER_CONF_PATH=/etc/iml/ltuer.conf
> # {name} correspond to the fields of iml_agent::action_plugins::lamigo::Config.name
> LAMIGO_UNIT_PATH=/etc/systemd/system/lamigo-{fs}-{mdt}.service
> ```

From the Michael's proposal to make the path to be the part of the environment.

I've decided _not_ to do just `LAMIGO_DIR=/etc/systemd/system`, since it is one of the standard systemd locations and to have it configurable looks redundant.

Additionally to this, we might want to configure the name of the unit itself. Therefore the full path is configurable and I put the comment above.

nlinker:
___
*[iml-agent/systemd-units/rust-iml-agent.service, line 17 at r7](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-LyZSPqv9npp5zipxa1o-r7-17:-LybPH1vDwnSOT73R4nh:b-siuqv0) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/f348af380e4fbe98bb8906ff1e2e3beb57eb8d9d/iml-agent/systemd-units/rust-iml-agent.service#L17)):*

Yep, let's discuss it, if it feels not right.
</blockquote></details>

This configuration is moved to `settings.conf`, let's continue discussing there.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 13 at r4](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r4-13:-LybPj837TNWgjbzFzrZ:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/9dd26350e080d85d4cd97fc30d6f65611ac70f78/iml-agent/src/action_plugins/lamigo.rs#L13)):*

I believe it's better to move this to systemd agent configuration and read from an environment variable:
https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-agent/systemd-units/rust-iml-agent.service
</blockquote></details>

Done.

nlinker:
> Looks like the latest push has merge conflicts.

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,177 @@
...
+            "[Unit]\n\
+             Description=Run lamigo service\n\
+             [Service]\n\
+             ExecStart=/usr/bin/lamigo \
+             -m {fs}-{mdt} \
+             -u {user} \
+             -s {hot_pool} \
+             -t {cold_pool} \
+             -a {min_age} \
+             {mount_point}\n\
```
utopiabound: We want lamigo to dump FIDs to socket (via postman) (c.f. EX-731)  The option is `--iml-socket`.  Do we need the mountpoint at all?

iml-agent/src/action_plugins/lustre.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Runs lctl with given arguments
+pub async fn lctl(args: Vec<String>) -> Result<String, ImlAgentError> {
+    lustre::lctl(args.iter().map(|s| s.as_ref()).collect())
+        .await
+        .map(|o| String::from_utf8_lossy(&o.stdout).to_string())
+}
+
+/// Executes `mountpoint -q /mnt/something`
+/// It provides no output, but signals the caller with the exit code
+pub async fn is_mounted(mount_point: String) -> Result<bool, ImlAgentError> {
```
ip1981: You can't be sure that the mounted fs is lustre, can you?

nlinker: So, I investigated the Lustre documentation, and sorry, I haven't found the tool to check that it was Lustre mounted to the mount point. The function will return `true` for anything, even flash drive or RAM disks.
One of the solution is `mount + grep`, looks too overcomlicated for me so far.
@utopiabound What do you suggest for checking the mount point?

mkpankov: I believe this is properly resolved by https://github.com/whamcloud/integrated-manager-for-lustre/pull/1176

@jgrund correct?

jgrund: @mkpankov yeah, between the data we get from `findmnt` and being able to expect zfs dataset props, we should know about all mounted devices and their backing type.

iml-agent/src/action_plugins/lamigo.rs
nlinker: Done, I'm using the option `--iml-socket` instead

iml-agent/src/action_plugins/lustre.rs
nlinker: Given the new information into account, I removed this action at all.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 13 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r1-13:-LzDGvgd8GtpBYzNwAsO:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/53032dbf1556b4fa082b4bae2d31600427a11339/iml-agent/src/action_plugins/lamigo.rs#L13)):*

Good eye!
</blockquote></details>

Done.

nlinker:
___
*[iml-agent/src/action_plugins/lustre.rs, line 17 at r8](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly_gTWeE9V50f5XdB4t-r8-17:-LzGE86n5KJLmyh6PfTL:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/7963de5706d115d94d4ef62dbd163e979a002842/iml-agent/src/action_plugins/lustre.rs#L17)):*

Given the new information into account, I removed this action at all.
</blockquote></details>

Done.

nlinker:
___
*[iml-agent/systemd-units/rust-iml-agent.service, line 17 at r7](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-LyZSPqv9npp5zipxa1o-r7-17:-LzGEFsU0B6Ht9c5EoNa:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/f348af380e4fbe98bb8906ff1e2e3beb57eb8d9d/iml-agent/systemd-units/rust-iml-agent.service#L17)):*

This configuration is moved to `settings.conf`, let's continue discussing there.
</blockquote></details>

Done.

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,176 @@
...
+        assert_eq!(
+            expand_path_fmt(fmt2, &config),
+            Ok("/tmp/nick/lamigo-LU_TEST1-MDT000.service".to_string())
+        );
+
+        let fmt3 = "lamigo-{unknown_value}.service";
+        assert_eq!(
+            expand_path_fmt(fmt3, &config),
+            Err(strfmt::FmtError::KeyError(
+                "Invalid key: unknown_value".to_string()
```
ip1981: You are testing what is already tested in `strfmt` itself. If we don't trust `strfmt`, maybe we shouldn't use it ;) And anyway what we are interested in is covered in the below test.

nlinker:
> Should struct `Config` be moved to wire-types and named something like `LamigoConfig`?

This is unexpected for me. If we are going to move this `Config` (renamed properly) to wire-types, then the [similar config for `LPurge`](https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-agent/src/action_plugins/lpurge.rs#L12-L26) is better to be moved too. However the LPurge config has `Display` with very  `lpurge`-specific implementation).
So I think either we do the same for LPurge, or leave as is.

> Does the `TryMount` code serve a purpose anymore?

AFAIK, we can check if a mount point is mounted with `device-scanner`, but we have `TryMount` to actually mount lustre. If there is some other duplicating action out there, I'll remove `TryMount`.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 143 at r9](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r9-143:-LzGIm_76P7Iy2sPLhYh:bdkatvk) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/841315d37468173fcf191de0087fe1f9749ccb92/iml-agent/src/action_plugins/lamigo.rs#L143)):*

You are testing what is already tested in `strfmt` itself. If we don't trust `strfmt`, maybe we shouldn't use it ;) And anyway what we are interested in is covered in the below test.
</blockquote></details>

I disagree, I'm testing the function `expand_path_fmt` that _uses_ `strfmt`. It has specific contents, although pretty simple.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 143 at r9](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r9-143:-LzGJNq-70r9YoZoqr32:br48a8c) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/841315d37468173fcf191de0087fe1f9749ccb92/iml-agent/src/action_plugins/lamigo.rs#L143)):*

I disagree, I'm testing the function `expand_path_fmt` that _uses_ `strfmt`. It has specific contents, although pretty simple.
</blockquote></details>

I wanted to check the failed expansion as well, so I'm doing it in the test.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 92 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r1-92:-LzGJ_n27mxvpuA1J6rv:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/53032dbf1556b4fa082b4bae2d31600427a11339/iml-agent/src/action_plugins/lamigo.rs#L92)):*

Yes, I tried to explicitly state that "here the things we do when force is true, here is otherwise". I'm ok with your proposal.
</blockquote></details>

Done.

iml-agent/src/action_plugins/lamigo.rs
ip1981: While I believe that `KeyError` is documented and [tested](https://docs.rs/crate/strfmt/0.1.6/source/src/tests/strfmt.rs), this part can be better with `.is_err()`, so that you don't depend on arbitrary strings.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 143 at r9](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r9-143:-LzGc9QUA8Sy1EfP-ViS:bbgbz1z) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/841315d37468173fcf191de0087fe1f9749ccb92/iml-agent/src/action_plugins/lamigo.rs#L143)):*

While I believe that `KeyError` is documented and [tested](https://docs.rs/crate/strfmt/0.1.6/source/src/tests/strfmt.rs), this part can be better with `.is_err()`, so that you don't depend on arbitrary strings.
</blockquote></details>

Ok, agree.

nlinker:
:fist_right: :fist_left:

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,171 @@
+// Copyright (c) 2019 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::{agent_error::ImlAgentError, daemon_plugins, env};
+use futures::future::TryFutureExt;
+use std::collections::HashMap;
+use std::path::Path;
+use tokio::fs;
+use tokio::io::AsyncWriteExt;
```
jgrund: nit: imports could be compressed.

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,171 @@
...
+        hot_pool = c.hot_pool,
+        min_age = c.min_age,
+        user = c.user,
+        iml_socket = iml_socket
+    )
+}
+
+pub async fn create_lamigo_service_unit(c: Config) -> Result<(), ImlAgentError> {
+    let path_fmt = env::get_var("LAMIGO_UNIT_PATH");
+    let iml_socket = daemon_plugins::postoffice::socket_name(&c.mailbox); // wants SOCK_DIR
```
jgrund: Not related to this patch, but feels like this should be in `env.rs` It seems strange to reach into another plugin to get some info.

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,171 @@
...
+    let path = expand_path_fmt(&path_fmt, &c)?;
+    create_lamigo_service_unit_internal(path, &c, &iml_socket)
+        .err_into()
+        .await
+}
+
+fn expand_path_fmt(path_fmt: &str, c: &Config) -> strfmt::Result<String> {
+    let mut vars = HashMap::new();
+    let min_age_str = c.min_age.to_string();
+    vars.insert("fs".to_string(), &c.fs);
```
jgrund: Any reason these can't be `&'static str`?

iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -0,0 +1,171 @@
...
+    iml_socket: &str,
+) -> std::io::Result<()> {
+    if let Some(dir_path) = file_path.as_ref().parent() {
+        fs::create_dir_all(&dir_path).await?;
+    };
+    let mut file = fs::File::create(file_path).await?;
+    let cnt = format_lamigo_unit_contents(c, iml_socket);
+    file.write_all(cnt.as_bytes()).await?;
+    Ok(())
+    // fs::write(file, cnt.as_bytes()).await
```
jgrund: Commented code.

iml-agent/src/action_plugins/lamigo.rs
nlinker: Ok, doing right now

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 72 at r11](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r11-72:-LzJ1Q6X97xsrq6HHlBM:b-y7lmuu) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/eabc91c9dd1bcee48bba31f6b9c737214ca6b09d/iml-agent/src/action_plugins/lamigo.rs#L72)):*

Not related to this patch, but feels like this should be in `env.rs` It seems strange to reach into another plugin to get some info.
</blockquote></details>

I propose to create a function `env::get_mailbox_socket` to obtain the socket  in the followup PR.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 82 at r11](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r11-82:-LzJ2qun1gY8xtJQWWIb:b-qclaw6) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/eabc91c9dd1bcee48bba31f6b9c737214ca6b09d/iml-agent/src/action_plugins/lamigo.rs#L82)):*

Any reason these can't be `&'static str`?
</blockquote></details>

I just avoided to create several excessive names just for a tiny optimization for the code, which most probably not being hot.

nlinker:
___
*[iml-agent/src/action_plugins/lamigo.rs, line 106 at r11](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1433#-Ly-rnIXR0_Z9mxFXKck-r11-106:-LzJ4BNY8Vq1XALlDujQ:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/eabc91c9dd1bcee48bba31f6b9c737214ca6b09d/iml-agent/src/action_plugins/lamigo.rs#L106)):*

Commented code.
</blockquote></details>

Done.

PR #1434, opened 2020-01-02
Add iml-orm crate

jgrund:
Add a new crate `iml-orm` that can be used with
[diesel](http://diesel.rs/) to make working with the db much simpler.

The schema has been generated from the IML Db, and structs have been
generated to represent each table in the schema.

The model structs may be used in other contexts for example both the
iml-agent and iml-gui.


mkpankov:
What's the async story with Diesel?

I believe some time ago they wasn't async and weren't planning. Are we going to work with it via a thread pool?

jgrund:
> What's the async story with Diesel?

They aren't async and there are some fundamental issues getting there.

> Are we going to work with it via a thread pool?

[yes](https://github.com/mehcode/tokio-diesel)

iml-manager-env/src/lib.rs
```diff
@@ -138,3 +138,27 @@ pub fn get_db_name() -> Option<String> {
...
+
+    if let Some(x) = get_db_name() {
+        xs.push(format!("dbname={}", x));
+    }
+
+    if let Some(x) = get_db_password() {
+        xs.push(format!("password={}", x));
+    }
+
+    let s = xs.join(" ");
```
johnsonw: Could do:
```
xs.join(" ")
```

jgrund: yeah, missed that lint.

iml-orm/src/schema.rs
```diff
@@ -0,0 +1,1907 @@
+table! {
```
mkpankov: Is this file written by hand?

jgrund: This is generated from a production db.

iml-manager-env/src/lib.rs
```diff
@@ -138,3 +138,25 @@ pub fn get_db_name() -> Option<String> {
...
     empty_str_to_none(get_var("DB_PASSWORD"))
 }
+
+/// Gets a connection string from the IML env
+pub fn get_db_conn_string() -> String {
+    let mut xs = vec![format!("user={}", get_db_user())];
+
+    let host = match get_db_host() {
+        Some(x) => x,
+        None => "/var/run/postgresql".into(),
```
nlinker: Never seen host like this before...

PR #1445, opened 2020-01-08
Setup a timer to handle stratagem scans at specified intervals

### LINKED ISSUE #1413: Setup IML Timer container docker service
We currently use a systemd timer running on the manager node to schedule stratagem runs in IML. Since docker setups will not use the manager node, a new *IML-Timer* container must be created. This container will run systemd in non-privileged mode and will have a small web service running that will listen for requests to add or remove a timer. By doing this, the job scheduler can then send a PUT request containing both the timer config as well as its corresponding service config. When the request is sent to the web service, the timer will be enabled immediately and the daemon will be reloaded.

johnsonw:
GH-1413 - Setup a timer to handle stratagem scans at specified intervals
    
Fixes #1413.
    
Our current solution for running stratagem scans at specified intervals is
to use systemd. Unfortunately, this won't work in the docker
environment. Instead, we should utilize a scheduling container that has
the ability to load a config file with the associated timer data and its
commands. Since the job scheduler will need to generate the timer
config, it will send the config over http to the reverse proxy, where it
will be routed to an http wrapper on the jobber service. Once the
command is received by the http service, it will save the config data
and reload the daemon.
    
---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1445)


docker/setup-jobber.sh
```diff
@@ -0,0 +1,39 @@
...
+# Naive check runs checks once a minute to see if either of the processes exited.
+# This illustrates part of the heavy lifting you need to do if you want to run
+# more than one service in a container. The container exits with an error
+# if it detects that either of the processes has exited.
+# Otherwise it loops forever, waking up every 60 seconds
+
+while sleep 60; do
+  # ps aux |grep jobberd |grep -q -v grep
+  # PROCESS_1_STATUS=$?
+  ps aux |grep iml-jobber |grep -q -v grep
```
ip1981: `pidof`, `pgrep`?

iml-jobber/src/main.rs
```diff
@@ -0,0 +1,80 @@
+use std::process::Command;
+use std::{
+    fs::File,
+    io::prelude::*,
+    str::{from_utf8, Utf8Error},
+};
+use tracing_subscriber::{fmt::Subscriber, EnvFilter};
+use warp::{self, Buf, Filter};
+
+static CONFIG_FILE: &str = "/home/jobberuser/.jobber";
```
ip1981: ?
```suggestion
static CONFIG_FILE: &str = "/etc/jobber.conf";
```

johnsonw: > The jobs are defined in the file /home/jobberuser/.jobber
https://hub.docker.com/_/jobber

docker/setup-jobber.sh
johnsonw: I can play with this a bit if you feel it is necessary. This is one of three options discussed in the docker documentation: https://docs.docker.com/config/containers/multi-service_container/

iml-jobber/src/main.rs
ip1981: Oh, interesting.

johnsonw:
Just a quick update showing that I am able to:
1. Configure the timer from the gui and verify that the timer is set inside the container
2. Verify that the scan is executed after the specified interval
3. Unconfigure the timer from the gui and verify that the timer is no longer set in the container

stratagem_timer_progress

docker/setup-jobber.sh
johnsonw: This file no longer exists.

johnsonw:
Items left:
- [ ] Make sure timer endpoint in nginx requires authorization
- [ ] Change name from iml-jobber to iml-timer

utopiabound:
Systemd timer state is held in: `/var/lib/systemd/timers/`

johnsonw:
I'll add a volume to hold the timer data. Thanks Nate.

docker/iml-timer.dockerfile
```diff
@@ -0,0 +1,14 @@
...
+FROM imlteam/systemd-base
+COPY --from=builder /build/target/release/iml-timer /bin/
+COPY docker/iml-timer/iml-timer.service /etc/systemd/system/
+COPY docker/iml-timer/start-stratagem-scan.py /usr/local/bin
+
+
+RUN yum install -y epel-release \
+  && yum install -y python2 \
+  && pip2 install requests \
+  && systemctl enable iml-timer
```
jgrund: There is no cleanup here for pip, the cache adds around 20MB. N.B. this doesn't matter if the command was rust, as we'd have no external deps.

chroma-manager.conf.template
```diff
@@ -236,6 +236,27 @@ server {
         proxy_pass {{MAILBOX_PROXY_PASS}}/mailbox;
     }
 
+    location /timer/configure {
```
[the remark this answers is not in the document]
johnsonw: Should I leave these entries out of nginx and instead send the request directly to the iml-timer container?

docker/iml-timer/start-stratagem-scan.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python2
```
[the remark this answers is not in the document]
johnsonw: sure, I will update.

iml-timer/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+    let subscriber = Subscriber::builder()
+        .with_env_filter(EnvFilter::from_default_env())
+        .finish();
+
+    tracing::subscriber::set_global_default(subscriber).unwrap();
+
+    // Match a config route
+    let config_route = warp::put()
+        .and(warp::path("configure"))
+        .and(warp::body::content_length_limit(1024 * 16))
```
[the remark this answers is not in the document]
johnsonw: I didn't think it would hurt to limit the length. We aren't expecting a lot of data so it seemed to make sense.

chroma-manager.conf.template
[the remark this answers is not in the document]
jgrund: Either that or add them for docker only.

iml-timer/src/config.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub name: String,
+    pub content: String,
+}
+
+pub struct ConfigFiles {
+    pub timer_file: ConfigFile,
+    pub service_file: ConfigFile,
+}
+
+pub fn unit_name(fid: &str) -> String {
```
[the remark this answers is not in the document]
johnsonw: I was staying consistent with https://github.com/whamcloud/integrated-manager-for-lustre/blob/9d07b6b0612953365274b19c2dcabb5253fd9a3c/chroma_core/models/stratagem.py#L95

iml-timer/src/main.rs
```diff
@@ -0,0 +1,104 @@
...
+                .await
+            }
+        })
+        .map(|_| Ok(StatusCode::NO_CONTENT));
+
+    let routes = config_route.or(unconfigure_route).recover(customize_error);
+
+    tracing::debug!("Serving routes.");
+    warp::serve(routes)
+        .run(iml_manager_env::get_timer_addr())
```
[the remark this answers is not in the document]
johnsonw: Ok, I will add this in.

docker/iml-timer.dockerfile
johnsonw: right, i'll get rid of the python file and can remove pip entirely.

iml-timer/src/config.rs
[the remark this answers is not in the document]
jgrund: What is an example of the id here?

chroma-manager.conf.template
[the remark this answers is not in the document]
johnsonw: I'll look into adding them conditionally.

iml-timer/src/config.rs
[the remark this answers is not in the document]
johnsonw: I will need to fire up another cluster to verify, but if I remember correctly the timer unit filename would look something like: `iml-stratagem-34.timer`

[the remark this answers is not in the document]
jgrund: Ok. I think fid is a misnomer here. Probably should change to just `unit_id`.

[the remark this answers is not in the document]
johnsonw: Looks like that's the correct format:
image

johnsonw:
I've addressed the above comments. Will post further documentation from testing unless there are other things you would like me to address.

johnsonw:
**Testing results:**
1. Configure scanning for 30 days:
```
[root@iml-timer /]# systemctl list-timers
NEXT                         LEFT                LAST                         PASSED      UNIT                         ACTIVATES
Wed 2020-02-12 19:20:48 UTC  22h left            Tue 2020-02-11 19:20:48 UTC  1h 0min ago systemd-tmpfiles-clean.timer systemd-tmpfiles-clean.service
Thu 2020-03-12 20:21:00 UTC  4 weeks 1 days left n/a                          n/a         iml-stratagem-1.timer        iml-stratagem-1.service
```

2. Generated service file:
```
[root@iml-timer /]# systemctl cat iml-stratagem-1.service
# /etc/systemd/system/iml-stratagem-1.service
# This file is part of IML
# This file will be overwritten automatically

[Unit]
Description=Start Stratagem run on 1

[Service]
Type=oneshot
EnvironmentFile=/var/lib/chroma/iml-settings.conf
ExecStart=/usr/bin/start-stratagem-scan --filesystem 1
```

Initial Scan:
image

After adding data:
image

Scanning from MDS node:
image

Scanning from iml-manager-cli container:
```
[root@4b6411870914 /]# iml stratagem scan -f fs -r 0s -p 300s
✔ Stratagem: Scanning all MDT's successful
```

Fids expiring soon file:
```
[root@c1 tmp]# cat 707f0dac-f3b6-43b8-81a7-2b4891c33e25-warn_fids-fids_expiring_soon.txt
/mnt/fs/simple-usage-test-files/test-5k.txt
/mnt/fs/simple-usage-test-files/test-100M.txt
/mnt/fs/simple-usage-test-files/test-1G.txt
/mnt/fs/simple-usage-test-files/test-1T.txt
```

I ran the scan again and the files were removed since it was > 5 minutes.

Removing the timer using the cli:
```
[root@4b6411870914 /]# iml stratagem interval remove -f fs
✔ Remove Stratagem for the given filesystem successful
[root@4b6411870914 /]# iml stratagem interval list
No Stratagem intervals found
[root@iml-timer /]# systemctl list-timers --all
NEXT                         LEFT     LAST                         PASSED    UNIT                         ACTIVATES
Wed 2020-02-12 21:46:05 UTC  23h left Tue 2020-02-11 21:46:05 UTC  15min ago systemd-tmpfiles-clean.timer systemd-tmpfiles-clean.service
n/a                          n/a      n/a                          n/a       systemd-readahead-done.timer systemd-readahead-done.service
```

Setting the timer from the cli:
```
iml stratagem interval add -f fs -r 1s -p 120s -i 65s
✔ Configuring Stratagem scan interval successful
root@c1 simple-usage-test-files]# ls | wc -l                                                                          │
3000
```
After the scan, the report was generated and all 3000 files were listed in the /tmp file. After the scan ran again and the files were > 2 minutes all 3000 files were purged.

**Scale Test:**

Scanning 100000 files:
```
[root@4b6411870914 /]# time iml stratagem scan -f fs -r 1s
✔ Stratagem: Scanning all MDT's successful

real    0m21.706s
user    0m0.268s
sys     0m0.043s
[root@c1 tmp]# cat 90561dda-1179-4c89-a8b9-fac436b047e7-warn_fids-fids_expiring_soon.txt | wc -l                       │1 timers listed.
100000
```

Purging Files:
```
[root@4b6411870914 /]# time iml stratagem scan -f fs -p 1s
✔ Stratagem: Scanning all MDT's successful

real    0m30.934s
user    0m0.365s
sys     0m0.066s

[root@c1 scale-test]# ls | wc -l
0
```

The results show that the docker build succeeded in all stratagem functionality and in performing the scale test just as it does using a standard setup.

chroma_core/models/stratagem.py
```diff
@@ -105,41 +105,113 @@ def service_file(fid):
...
+            # Create timer file
+            timer_config = """# This file is part of IML
+# This file will be overwritten automatically
+
+[Unit]
+Description=Start Stratagem run on {}
+
+[Timer]
+OnActiveSec={}
+OnUnitActiveSec={}
```
ip1981: Nit: maybe use named parameters for readability and robustness. As a side effect, can dedup `config.interval / 1000`

docker/iml-timer/iml-timer.conf
```diff
@@ -0,0 +1,20 @@
...
+
+location /timer/unconfigure/ {
+    auth_request /auth;
+    proxy_set_header Host $host;
+    proxy_set_header X-Real-IP $remote_addr;
+    proxy_set_header X-Forwarded-Host $host;
+    proxy_set_header X-Forwarded-Server $host;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    rewrite ^/timer/unconfigure/(.*)$ /unconfigure/$1 break;
+    proxy_pass http://iml-timer:8892;
```
ip1981: With the above rewrite rule this looks similar to `proxy_pass http://iml-timer:8892/configure;`, no?

docker/setup-nginx
```diff
@@ -43,5 +43,7 @@ with open(conf_template, "r") as f:
 
     config = re.sub(r"{{(.+)}}", r"{{ .Env.\g<1> }}", config)
 
+    config = re.sub(r"^}\s\Z", "\n    include /etc/nginx/conf.d/iml-timer.extras;\n}", config, 0, re.MULTILINE)
```
ip1981: I wonder if nginx's config says something like `include /etc/nginx/conf.d/*`

chroma_core/models/stratagem.py
johnsonw: sure 👍

docker/iml-timer/iml-timer.conf
johnsonw: Yeah this one is kind of tricky. I needed to use the rewrite rule in order to pass the id to the proxied server.

docker/setup-nginx
johnsonw: One thing I could do is create a folder like this:
```
/etc/nginx/conf.d/extras
```
and put in any extra location files in that directory. With that said I don't think it's necessary at the moment.

docker/iml-timer/iml-timer.conf
ip1981: I think a rewrite rule is not necessary according to https://docs.nginx.com/nginx/admin-guide/web-server/reverse-proxy/:

> If the URI is specified along with the address, it replaces the part of the request URI that matches the location parameter. For example, here the request with the /some/path/page.html URI will be proxied to http://www.example.com/link/page.html

johnsonw: I believe this was my first approach actually and it didn't work. I can try changing it back to this and see if it works.

johnsonw:
I tested deleting the configuration after removing the rewrite rule and I get the following:

```
iml_nginx.1.l2gphunmorcm@onyx-24.onyx.whamcloud.com    | 10.0.5.9 - - [13/Feb/2020:15:03:08 +0000] "DELETE /timer/unconfigure/1 HTTP/1.1" 405 23 "-" "python-requests/2.12.5"
```

image

I think the rewrite rule is necessary.

ip1981:
What is in the backend logs? (not nginx)

docker/iml-timer/iml-timer.conf
ip1981: I think it has to be `proxy_pass http://iml-timer:8892/unconfigure`

ip1981: And, maybe, it could be a single entry:

```
location /timer/ {
    auth_request /auth;
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-Server $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_pass http://iml-timer:8892/;
}
```

But I am not sure because I don't know the whole thing :)

johnsonw:
The logs in the iml-timer container don't contain any pertinent information as it hasn't received a valid request. What's happening is that iml-timer is configured to look for "/configure" and "/unconfigure/{id}". But without the rewrite rule, the proxy goes to "/timer/unconfigure/{id}" instead of "/unconfigure/{id}" and it returns a 405 (method not allowed).

docker/iml-timer/iml-timer.conf
jgrund: @johnsonw Have you tried leaving the trailing slash off the location directive? I.E.

`location /timer/unconfigure`

Also, you could probably match inside the location block itself to avoid the rewrite rule:

```
    location ~ /timer/unconfigure/(.+)$ {
        auth_request /auth;
        proxy_set_header Host $host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-Host $host;
        proxy_set_header X-Forwarded-Server $host;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_pass http://iml-timer:8892/unconfigure/$1;
    }
```

johnsonw: The single block worked:
```
location /timer/ {
    auth_request /auth;
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-Server $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_pass http://iml-timer:8892/;
}
```

Configuring Timer:
image

Unconfiguring Timer:
image

```
[root@iml-timer /]# systemctl list-timers
NEXT                         LEFT     LAST                         PASSED    UNIT                         ACTIVATES
Fri 2020-02-14 16:17:31 UTC  23h left Thu 2020-02-13 16:17:31 UTC  17min ago systemd-tmpfiles-clean.timer systemd-tmpfiles-clean.service

1 timers listed.
```
I will push a PR shortly with the update

PR #1447, opened 2020-01-08
Add Table component

### LINKED ISSUE #1446: Implement Table component
Implement a table component that uses the new look and feel.

jgrund:
Fixes #1446.

Add new table component that implements look and feel.

table

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1447)

### DIFF
--- a/iml-gui/crate/src/components/table.rs
+++ b/iml-gui/crate/src/components/table.rs
@@ -1,0 +1,40 @@
+use crate::generated::css_classes::C;
+use seed::{prelude::*, *};
+
+pub fn wrapper_view<T>(children: impl View<T>) -> Node<T> {
+    table![
+        class![C.table_auto, C.w_full, C.text_center],
+        style! {
+            St::BorderSpacing => px(10),
+            St::BorderCollapse => "initial"
+        },
+        children.els()
+    ]
+}
+
+pub fn thead_view<T>(children: impl View<T>) -> Node<T> {
+    thead![style! { St::BorderSpacing => "0 10px"}, tr![children.els()]]
+}
+
+fn th_base<T>(more_attrs: Attrs, children: impl View<T>) -> Node<T> {
+    let mut cls = class![C.px_3, C.text_left, C.text_gray_800, C.font_normal, C.text_center];
+
+    cls.merge(more_attrs);
+
+    th![cls, children.els()]
+}
+
+pub fn th_view<T>(children: impl View<T>) -> Node<T> {
+    th_base(Attrs::empty(), children.els())
+}
+
+pub fn th_sortable_view<T>(children: impl View<T>) -> Node<T> {
+    th_base(class![C.border_b_2, C.border_blue_500], children.els())
+}
+
+pub fn td_view<T>(children: impl View<T>) -> Node<T> {
+    td![
+        class![C.px_3, C.bg_gray_100, C.rounded, C.text_center, C.p_4],
+        children.els()
+    ]
+}
--- a/iml-gui/crate/src/components/table.rs
+++ b/iml-gui/crate/src/components/table.rs
@@ -1,9 +1,17 @@
 use crate::generated::css_classes::C;
-use seed::{prelude::*, *};
+use seed::{prelude::*, Attrs, *};
 
-pub fn wrapper_view<T>(children: impl View<T>) -> Node<T> {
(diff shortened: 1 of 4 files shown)


iml-gui/crate/src/components/table.rs
```diff
@@ -0,0 +1,40 @@
...
+    th_base(Attrs::empty(), children.els())
+}
+
+pub fn th_sortable_view<T>(children: impl View<T>) -> Node<T> {
+    th_base(class![C.border_b_2, C.border_blue_500], children.els())
+}
+
+pub fn td_view<T>(children: impl View<T>) -> Node<T> {
+    td![
+        class![C.px_3, C.bg_gray_100, C.rounded, C.text_center, C.p_4],
```
ip1981: Sometimes it needs to align to the left or right, for example, size of FS.

ip1981: I wonder if we should start using our own CSS classes (basing on tailwind) - they might make things easily compostable.

jgrund: Even with size I'm not sure why it couldn't be center aligned...

jgrund: > I wonder if we should start using our own CSS classes (basing on tailwind) - they might make things easily compostable.

It's a philosophical thing for sure, but this approach is basically using atomic css: https://acss.io/thinking-in-atomic.html

I find this easier to work with (even if classes end up being repeated) as things are consistent and the style logic is co-located within the view.

ip1981: It could, but looks not good:

```
1234
 20
```

Vs

```
1234
  40
```

ip1981: BTW, I think alerts could look better aligned to the left (because of the bell icon).

ip1981: Basically, having a helper functions (components) is equivalent to custom CSS classes, I just get a feeling that it is less composable, probably adding extra classes should override defaults if needed, but in case of `td_view` there is not such way.

iml-gui/crate/src/components/table.rs
```diff
@@ -0,0 +1,40 @@
...
+        children.els()
+    ]
+}
+
+pub fn thead_view<T>(children: impl View<T>) -> Node<T> {
+    thead![style! { St::BorderSpacing => "0 10px"}, tr![children.els()]]
+}
+
+fn th_base<T>(more_attrs: Attrs, children: impl View<T>) -> Node<T> {
+    let mut cls = class![C.px_3, C.text_left, C.text_gray_800, C.font_normal, C.text_center];
```
ip1981: `C.text_left` + `C.text_center`

jgrund: Fixed.

iml-gui/crate/src/components/table.rs
jgrund: Abstracted classes to their own standalone fns, and added overrides to all helper elements.

This should allow for 90% usage, and escape hatches if needed.

iml-gui/crate/src/components/table.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn wrapper_view<T>(more_attrs: Attrs, children: impl View<T>) -> Node<T> {
+    let mut cls = wrapper_cls();
+
+    cls.merge(more_attrs);
+
+    table![
+        cls,
+        style! {
+            St::BorderSpacing => px(10),
+            St::BorderCollapse => "initial"
```
ip1981: Could it be part of `wrapper_cls()`?

iml-gui/crate/src/components/table.rs
```diff
@@ -0,0 +1,61 @@
+use crate::generated::css_classes::C;
+use seed::{prelude::*, Attrs, *};
+
+pub fn wrapper_cls() -> Attrs {
+    class![C.table_auto, C.w_full]
```
ip1981: This is almost the same as defining a custom class like
```css
.wrapper-cls {
  @apply table-auto w-full;
}
```
It is not "atomic", but "molecular" for sure :)
Putting it in a CSS file might have a small advantage for not requiring recompilation Rust, by the way.

PR #1455, opened 2020-01-10
Implement full local provisioning

jgrund:
This patch implements provisioning / installation of both Python and
Rust code from the local repo on both the manager and agent nodes.

Previously only the python code was built / installed.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1455)

### DIFF
--- a/vagrant/scripts/install_iml_local_agent.sh
+++ b/vagrant/scripts/install_iml_local_agent.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+yum install -y yum-plugin-copr
+yum copr enable -y managerforlustre/manager-for-lustre-devel
+mkdir -p /tmp/agent-rpms
+scp adm:/tmp/agent-rpms/*.rpm /tmp/agent-rpms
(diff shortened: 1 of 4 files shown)


vagrant/scripts/install_iml_local_agent.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+yum install -y yum-plugin-copr
+yum copr enable -y managerforlustre/manager-for-lustre-devel
+mkdir -p /tmp/agent-rpms
+scp adm:/tmp/agent-rpms/*.rpm /tmp/agent-rpms
+yum localinstall -y /tmp/agent-rpms/*.rpm
```
johnsonw: No reason to use `yum localinstall` unless on centos 5. It should still work the same as `yum install`.

jgrund: ref?

jgrund: yeah, guess it makes sense that command is deprecated... I liked that it signals intent.

PR #1482, opened 2020-01-14
Add Modal Component

### LINKED ISSUE #1408: Create Modal Component
Create a modal component that can be used in the seed framework. The modal should be:
1. Responsive
2. Should grow to accommodate content
3. Should have a title section, body section, and footer section
4. Should allow buttons in the footer section
5. Should have a close button in the top right corner that is optionally displayed
6. Should be able to be closed by hitting the escape key
7. Each open modal should appear on top of any previously opened modals

jgrund:
Fixes #1408.

modal

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1482)

### DIFF
--- a/iml-gui/crate/src/components/modal.rs
+++ b/iml-gui/crate/src/components/modal.rs
@@ -1,4 +1,4 @@
-use crate::{components::font_awesome, generated::css_classes::C};
+use crate::{components::font_awesome, generated::css_classes::C, key_codes};
 use seed::{prelude::*, *};
 
 #[derive(Default)]
@@ -15,7 +15,7 @@ pub enum Msg {
 pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
     match msg {
         Msg::KeyDown(code) => {
-            if code == 27 {
+            if code == key_codes::ESC {
                 orders.send_msg(Msg::Close);
             }
         }
--- a/iml-gui/crate/src/components/modal.rs
+++ b/iml-gui/crate/src/components/modal.rs
@@ -1,0 +1,103 @@
+use crate::{components::font_awesome, generated::css_classes::C};
+use seed::{prelude::*, *};
+
+#[derive(Default)]
+pub struct Model {
+    pub open: bool,
+}
+
+#[derive(Clone)]
+pub enum Msg {
+    KeyDown(u32),
+    Close,
+}
+
+pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
+    match msg {
+        Msg::KeyDown(code) => {
+            if code == 27 {
+                orders.send_msg(Msg::Close);
+            }
+        }
+        Msg::Close => {
+            model.open = false;
+        }
+    }
+}
+
+pub fn title_view(children: impl View<Msg>) -> Node<Msg> {
+    div![
+        class![C.flex, C.justify_between, C.items_center, C.pb_3,],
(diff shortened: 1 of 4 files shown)


iml-gui/crate/src/components/modal.rs
```diff
@@ -0,0 +1,103 @@
...
+#[derive(Clone)]
+pub enum Msg {
+    KeyDown(u32),
+    Close,
+}
+
+pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg>) {
+    match msg {
+        Msg::KeyDown(code) => {
+            if code == 27 {
```
johnsonw: Is there a constant we can use to indicate what 27 is?

jgrund: Nope. This should really be part of seed, but I'll add a module for keycodes.

jgrund:
> Does this modal account for z-indexing if a modal opens on top of an existing modal?

I'll add an `atomicu32` that will increment the stacking order each time a modal is opened.

iml-gui/crate/src/components/modal.rs
johnsonw: Was just curious. I'll +1.

johnsonw:
If you want to do that in a separate patch I don't see any reason to block it.

jgrund:
> If you want to do that in a separate patch I don't see any reason to block it.

May as well fix here.

jgrund:
> Does this modal account for z-indexing if a modal opens on top of an existing modal?

Actually, I'm not sure I need to do anything if both are absolute and have the same z-index...

jgrund:
Thinking about this more, if we end up needing to keep track of stacks, we should probably have a "global" (top-level of app model) `modal_stack` that bookeeps open modals. This will be needed for handling esc correctly as lower modals won't get autofocused when the top one closes.

johnsonw:
I agree, that makes sense.

iml-gui/crate/src/components/modal.rs
```diff
@@ -0,0 +1,103 @@
...
+            span![class![C.text_sm,], "(Esc)",],
+        ],
+        div![class![C.py_4, C.text_left, C.px_6,], children.els()]
+    ]
+}
+
+pub fn bg_view(open: bool, children: impl View<Msg>) -> Node<Msg> {
+    if !open {
+        return empty![];
+    }
```
nlinker: How do you make sure, that all other controls are not accessible, if `open == true`?

nlinker: How do you make sure, that all other controls on the page are not accessible, if `open == true`?

jgrund: The controls behind the modal?

There is an overlay that blocks any events to elements underneath the overlay.

nlinker: I thought it is going to be a transparent layer with `z_index` like 999 but haven't noticed this.

jgrund: Doesn't need to be transparent... but there is a full size overlay. In terms of z-index, we don't want to specify this as overlays can stack at this point.

nlinker: Ok, thanks for the explanation!

jgrund:
I think for now we can land this. I'm not sure we'll need the stack tracking; and if we do, we can implement it separately.

jgrund:
Landed in baf536b02c70695c7188faaab4e9ee7d07742faa

PR #1484, opened 2020-01-16
Fix: install-iml-local provisioner agent packages are overwritten

### LINKED ISSUE #1483: install-iml-local provisioner agent packages are overwritten
The `install-iml-local` provisioner installs rust packages on the manager and agent.

However, the packages that are generated are production style (no epoch seconds in nevra), so they are overwritten when the agent is deployed / updated by IML.

Instead of generating production-style packages, we should both bump the release and add epoch seconds to the version to ensure our local packages always get chosen.

jgrund:
Fixes #1483.

These changes should ensure the locally built packages get chosen (within a comparable range).

It also enables reinstall capabilities.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1484)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,11 +38,11 @@ all: copr-rpms rpms
 
 rpms:
 	$(MAKE) -f .copr/Makefile iml-srpm outdir=.
-	rpmbuild -D "_topdir $(CURDIR)/_topdir" -bb _topdir/SPECS/python-iml-manager.spec
+	rpmbuild --rebuild -D "_topdir $(CURDIR)/_topdir" _topdir/SRPMS/python-iml-manager-*.src.rpm
 
 copr-rpms:
 	$(MAKE) -f .copr/Makefile srpm outdir=.
-	rpmbuild -D "_topdir $(CURDIR)/_topdir" -bb _topdir/SPECS/rust-iml.spec
+	rpmbuild --rebuild -D "_topdir $(CURDIR)/_topdir" _topdir/SRPMS/rust-iml-*.src.rpm
 
 cleandist:
 	rm -rf dist
--- a/vagrant/scripts/install_iml_local.sh
+++ b/vagrant/scripts/install_iml_local.sh
@@ -5,15 +5,31 @@ yum install -y rpmdevtools git ed epel-release python-setuptools gcc openssl-dev
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 source $HOME/.cargo/env
 cd /integrated-manager-for-lustre
-make copr-rpms
 
+# Bump the release number. This should ensure we get picked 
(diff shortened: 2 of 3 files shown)


jgrund:
___
*[Makefile, line 41 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1484#-LymW6oG08oHrmJb72Bx:-Lyna87n0SLr3dm2df48:b-brnb0s) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/16715db4dfbfcac8fb13f18df272f63b8a846f09/Makefile#L41)):*

What's the caching policy on this? Can one always safely rebuild from SRPM or some changes require rebuild of SRPM itself?
</blockquote></details>

The previous line `$(MAKE) -f .copr/Makefile iml-srpm outdir=.` does the SRPM build. What was previously happening is the SRPM was built on that line, and then it was built again on this line.

Now it's rebuilt from SRPM which is what should have been happening imo.

jgrund:
___
*[vagrant/scripts/install_iml_local.sh, line 13 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1484#-LymWSOU3tusYSIHY6Wo:-Lynb0XC-mUZvy5WUG_u:b-zfzc6b) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/16715db4dfbfcac8fb13f18df272f63b8a846f09/vagrant/scripts/install_iml_local.sh#L13)):*

One thing I don't like currently is that build process modifies git versioned files (these specs) and makes workspace dirty. Would it be a lot of work to address here? Otherwise, I'll make an issue for that.
</blockquote></details>

Changes are rsynced from the host so any changes inside the VM (where this script occurs) do not reflect onto the host. These changes are erased whenever `vagrant rsync adm` is run (which happens automatically on boot).

Given changes are typically coming from outside the VM and rsynced over, I'm not sure it makes sense to cleanup inside the VM afterwards...

jgrund:
___
*[vagrant/scripts/install_iml_local.sh, line 13 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1484#-LymWSOU3tusYSIHY6Wo:-Lynqsxm5zZnjTZ51l84:bpjlva4) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/16715db4dfbfcac8fb13f18df272f63b8a846f09/vagrant/scripts/install_iml_local.sh#L13)):*

I guess I'm talking not about this script specifically, I'll make an issue
</blockquote></details>

Everything done in these scripts is done within the VM, you may be thinking of the docker builder which does alter the spec in the working tree (due to the rw bind mount).

Issue #1486, opened 2020-01-16
iml-client::check_stonith: Add support for ExaScalar


utopiabound:
ES 5.x uses `fence_sfa_vm`, which should be supported.

PR #1488, opened 2020-01-16
check_stonith: Fix untrimmed nodename

utopiabound:
Also test with different realworld Stonith config.

Fixes: #1486 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1488)


iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -376,4 +392,20 @@ mod tests {
             }
         );
     }
+
+    #[test]
+    fn test_stonith_exascalar_static() {
+        let testxml = include_bytes!("snapshots/check_stonith_test_es.xml");
```
jgrund: is snapshots the right place for this? We probably shouldn't write anything there since it's managed by `insta`.

How about `fixtures` instead?

utopiabound: okay.  I think I'll need to change one of the other patches then.

iml-agent/src/action_plugins/check_stonith.rs
```diff
@@ -109,6 +109,8 @@ fn do_check_stonith(xml: &[u8], nodename: &str) -> Result<ComponentState<bool>,
         ..Default::default()
     };
 
+    let nodename = nodename.trim();
```
nlinker: The spot, I guess

PR #1493, opened 2020-01-17
Account for pacemaker template types

### LINKED ISSUE #1485: check_ha::get_resource_list should handle templated resources
Pacemaker Explained - [11.1.1. Configuring Resources with Templates](https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/2.0/html-single/Pacemaker_Explained/index.html#_configuring_resources_with_templates)

`check_ha::get_resource_list()` only supports fully defined resources, but if resources are defined via templates the agent info will be empty.

We should apply the templates to the resources defined by them.

utopiabound:
Fixes #1485 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1493)


iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -38,14 +48,31 @@ fn create<'a>(elem: &Element, group: impl Into<Option<&'a str>>) -> ResourceAgen
 
 fn process_resource_list(output: &[u8]) -> Result<Vec<ResourceAgentInfo>, ImlAgentError> {
     let element = Element::from_reader(output)?;
-
+    let templates: HashMap<String, ResourceAgentType> = element
+        .find_all("template")
+        .map(|elem| {
+            (
+                elem.get_attr("id").unwrap_or("").to_string(),
```
jgrund: nit: use `.unwrap_or_default()` instead:

`elem.get_attr("id").unwrap_or_default().to_string()`

jgrund: More importantly, this has the potential to use an empty `String` as a `HashMap` key.

Shouldn't we `.filter` empty keys rather than pushing an empty key into the map?

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -38,14 +48,31 @@ fn create<'a>(elem: &Element, group: impl Into<Option<&'a str>>) -> ResourceAgen
...
+            )
+        })
+        .collect();
     Ok(element
         .find_all("group")
         .flat_map(|g| {
+            let templates = &templates;
             g.find_all("primitive")
-                .map(move |p| create(p, g.get_attr("id").unwrap_or_default()))
+                .map(move |p| create(p, g.get_attr("id").unwrap_or_default(), templates))
```
jgrund: No need for `unwrap_or_default` here

```suggestion
                .map(move |p| create(p, g.get_attr("id"), templates))
```

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -38,14 +48,31 @@ fn create<'a>(elem: &Element, group: impl Into<Option<&'a str>>) -> ResourceAgen
 
 fn process_resource_list(output: &[u8]) -> Result<Vec<ResourceAgentInfo>, ImlAgentError> {
     let element = Element::from_reader(output)?;
-
+    let templates: HashMap<String, ResourceAgentType> = element
+        .find_all("template")
+        .map(|elem| {
```
jgrund:
```suggestion
        .filter(|elem| elem.get_attr("id").is_some())
        .map(|elem| {
```

iml-agent/src/action_plugins/check_ha.rs
```diff
@@ -38,14 +48,31 @@ fn create<'a>(elem: &Element, group: impl Into<Option<&'a str>>) -> ResourceAgen
...
+            (
+                elem.get_attr("id").unwrap_or("").to_string(),
+                ResourceAgentType::new(
```
jgrund: Is it always safe to assume if we hit a `template` that it will alway have these values?

PR #1495, opened 2020-01-17
Create Vagrant test framework

### LINKED ISSUE #1471: Create vagrant-test binary
Create a rust integration test binary that can interface with vagrant.

At a minimum it should be able to run the following commands:

`vagrant up [hosts]...`

`vagrant halt [hosts]...`

`vagrant snapshot save [host] <name>`

`vagrant snapshot delete [host] <name>`

`vagrant snapshot restore [host] <name>`

`vagrant provision --provision-with=<name> [hosts]...`

### LINKED ISSUE #1472: Create test setup / teardown fns
Abstract setup / teardown fns

- `setup_bare()`
    - `vagrant up iscsi adm mds1 mds2 oss1 oss2 c1`
    - `vagrant provision --provision-with=yum-update`
    - `vagrant halt`
    - `vagrant snapshot save {iscsi,adm,mds1,mds2,oss1,oss2} bare`
    - `vagrant up iscsi adm mds1 mds2 oss1 oss2 c1`

- `setup_iml_install()`
   - run `setup_bare_cluster` within this fn
   - `vagrant provision --provision-with=install-iml-local`
   - `vagrant halt`
   - `vagrant snapshot save {iscsi,adm,mds1,mds2,oss1,oss2} iml-installed`

- `setup_deploy_servers(hostlist:String, profile:String)`
    - run `setup_iml_install` within this fn
    - `ssh adm iml server add -h {hostlist} -p {profile}`
    - `vagrant snapshot save {iscsi,adm,mds1,mds2,oss1,oss2} iml-deployed`

jgrund:
This patch implements a rudimentary vagrant test framework.

Fixes #1471.
Fixes #1472.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1495)


mkpankov:
Will play with it a little tomorrow

jgrund:
To run with std{out,err} in terminal:

`cargo test -p iml-integration-tests -- --nocapture`

mkpankov:
Tried it on Windows, got couple issues. Not 100% we should try solving them, but maybe they aren't Windows-specific. [First](https://gist.github.com/mkpankov/504284dd2e206b36cfd50600f26ae8d3). [second](https://gist.github.com/mkpankov/e7cf046bab8fea301b2c9d9175a305f9).

mkpankov:
On Linux, got an [issue](https://gist.github.com/mkpankov/c16377b7d66de656ca2889ef6de177ab) that looks similar to second one on Windows.

.github/workflows/rust-ci.yml
```diff
@@ -77,6 +77,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
+          args: --exclude iml-integration-tests --workspace
```
ip1981: Locally this might be inconvenient.

iml-integration-tests/tests/integration_test.rs
```diff
@@ -0,0 +1,14 @@
...
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+mod vagrant;
+
+#[tokio::test]
+async fn test_setup() -> Result<(), Box<dyn std::error::Error>> {
+    let config = vagrant::ClusterConfig::default();
+
+    vagrant::setup_deploy_servers(&config, "base_managed_patchless").await?;
```
ip1981: Why a new wrapper binary in Rust instead of a simple shell script in `./vagrant`? Windows?

jgrund: This allows more flexibility in that the specific servers / profile can be specified programmatically.

.github/workflows/rust-ci.yml
jgrund: Open to other suggestions.

ip1981: Maybe it shouldn't be implemented as a test, but a standalone executable?

ip1981: There is also `#[ignored]` and `cargo test -- --ignored`.

jgrund:
___
*[.github/workflows/rust-ci.yml, line 80 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1495#-Lz2dQt5n6lr0XIcIgmo-r1-80:-LzHgd513W-yPsxbMbVW:bm79imz) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/450d2122f263a3dcbf5a4ce6c8660f6efd960674/.github/workflows/rust-ci.yml#L80)):*
I thought about executable initially as well, however I think using `cargo test` will ultimately be better. 

Right now this only has one test, but it will eventually have many tests, at which point having something like cargo test to drive it will be better than a standalone where we write our own test runner.

In addition, we'll want to hook into test output and display test results in Jenkins which again wouldn't really make sense without a test runner imo.

> We don't gain much from hooking into cargo test, and this feels like repurposing.

I'm not sure I understand this, isn't the point of `cargo test` to run tests?


> Our tests aren't actually integration, they're system ones.

Sure, I'll rename the module.

.github/workflows/rust-ci.yml
ip1981: My main concern is the possibility to step on the rake accidentally, in my view having these thing in a single repository and within single cargo workspace and use of the same testing framework make it more likely, but other than this I have no objections. I realized that I don't run tests ~~so I don't care~~ in a batch anyway (only per crate I am working on), thus it is ok :)

nlinker:
___
*[.github/workflows/rust-ci.yml, line 80 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1495#-Lz2dQt5n6lr0XIcIgmo-r1-80:-Lzg4q6RECTjwI-MO2q7:b-di8n4a) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/450d2122f263a3dcbf5a4ce6c8660f6efd960674/.github/workflows/rust-ci.yml#L80)):*

My main concern is the possibility to step on a rake accidentally, in my view having these thing in a single repository and within single cargo workspace and use of the same testing framework make it more likely, but other than this I have no objections. I realized that I don't run tests ~~so I don't care~~ in a batch anyway (only per crate I am working on), thus it is ok :)
</blockquote></details>

My position is close to Joe's, `cargo test` reusage might give us many benefits, e.g. CI support.
An additional variant we might consider is the following:
```
cargo test   # runs the ordinary test suite

cargo test --system-tests   # runs the new tests against locally deployed vagrant vms
```

jgrund:
___
*[.github/workflows/rust-ci.yml, line 80 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1495#-Lz2dQt5n6lr0XIcIgmo-r1-80:-M2P3MXl7DlEpaoGirGq:bcyymyo) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/450d2122f263a3dcbf5a4ce6c8660f6efd960674/.github/workflows/rust-ci.yml#L80)):*

```
cargo test --system-tests
```

That's not possible, I think? `cargo test` has its own flags we don't hook into. We can have `cargo test -- --system-tests` to pass our flags to the test binary, however, these flags are actually for `cargo` itself if they are to work.

Looking at its flags, what we can afford is probably something like
```
cargo test --test system_test_1
```

And now this test should be always run manually separately from other tests, because most of them can be run in parallel, but those with vagrant can't.

In this option, I don't like having to know the test's name.

We can probably also have

```
cargo test system_test_ -- --test-threads 1
```
to run all tests with names containing `system_test_` one after another.

I guess I'm not satisfied with default cargo test runner to begin with, but if we want to spend zero time on developing a test harness, we should make something from above options work.
</blockquote></details>

I've removed iml-system-tests from the workspace. It will not run by default.

I do agree that we should have a custom harness, but right now we have no automated testing for vagrant, and this will be *something* and a step in the right direction. After this lands, we can think about how to write a proper harness.

iml-system-tests/tests/integration_test.rs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2019 DDN. All rights reserved.
```
johnsonw: Copyright year

iml-system-tests/tests/vagrant.rs
```diff
@@ -0,0 +1,227 @@
+// Copyright (c) 2019 DDN. All rights reserved.
```
johnsonw: Year

iml-system-tests/tests/integration_test.rs
jgrund: We'll need to pass through and update everything in one patch.

Issue #1479, opened 2020-01-14
Remove all usages of parking lot from iml-agent


jgrund:
The iml-agent currently uses `parking-lot` in a few places. This will block the entire task when a lock is acquired.

Instead uses a futures-compatible locking mechanism.

utopiabound:
For postman this requires `DaemonPlugin::teardown()` to become async.

mkpankov:
Seems in some places we won't be able to do the replacement.

One of these:
https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-agent/src/daemon_plugins/action_runner.rs#L71

If we switch to `futures::lock::Mutex`, it returns a future, so we have to use `.lock().await`, and `await` can only be used in `async` functions. Here it's a trait method, and it's not supported by compiler yet.

Looking at others now.

mkpankov:
Apparently all of our usages of `Mutex` are in trait methods, so can't replace any of it with futures at the moment.

jgrund:
> If we switch to futures::lock::Mutex, it returns a future, so we have to use .lock().await, and await can only be used in async functions. Here it's a trait method, and it's not supported by compiler yet.

`async` directly isn't supported, but any use of async can be pinned and boxed which can then be used in a trait method.

mkpankov:
Code I was referring to:
https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-agent/src/http_comms/session.rs#L190

`&mut info.lock()` becomes `&mut *(info.lock().await)` and it snowballs from there.

This is with signature converted to `-> Pin<Box<dyn Future ...`.

mkpankov:
Made changes to all `Session`'s methods except `teardown`.

For teardown it's really complicated as we don't have a way to prevent borrowing the `Sessions`, and that borrow needs to live for `'static`.

https://github.com/whamcloud/integrated-manager-for-lustre/pull/1513/commits/d4cc088d49c456418915d6ad14d68df8efea42cb#diff-097c3716dcec898101971ec6913b99edR159

It would be possible if we had a way to strip `Arc<RwLock<...>>` in `teardown` and consume the inner value instead.

PR #1496, opened 2020-01-20
Replace parking_lot::Mutex in postoffice

mkpankov:
This is first part of #1479.

We could also use [`tokio::task::block_in_place`](https://docs.rs/tokio/0.2.9/tokio/task/fn.block_in_place.html) instead of [`futures::executor::block_on`](https://docs.rs/futures/0.3.1/futures/executor/fn.block_on.html) but that would require two features (`rt-threaded` and `blocking`) and it has less convenient API where you need to pass a closure returning a result. `async` closures are even unstable.

This PR improves code by making everything except `teardown` and `Debug::fmt` async. Making `teardown` async is not really viable IMO, as the future needs to outlive the data it refers to, and for teardown, it doesn't seem to make sense. With `fmt`, we're stuck with `std` signature where we can't sneak anything in or out.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1496)

### DIFF
--- a/iml-agent/src/daemon_plugins/postoffice.rs
+++ b/iml-agent/src/daemon_plugins/postoffice.rs
@@ -50,7 +50,11 @@ pub fn socket_name(mailbox: &str) -> String {
 
 impl std::fmt::Debug for PostOffice {
     fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "PostOffice {{ {:?} }}", block_on(self.routes.lock()).keys())
+        write!(
+            f,
+            "PostOffice {{ {:?} }}",
+            block_on(self.routes.lock()).keys()
+        )
     }
 }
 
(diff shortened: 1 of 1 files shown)


mkpankov:
First, we're not making it worse: `teardown` is blocking now, too.

Second, `teardown` being async would mean it produces a `Future` that gets ownership over the data being torn down - so accepting `self` rather than `&mut self`. And that future needs to be processed by someone.

I can try to go that way if you want.

jgrund:
> First, we're not making it worse: teardown is blocking now, too.

Yes, this is an improvement.

> Second, teardown being async would mean it produces a Future that gets ownership over the data being torn down - so accepting self rather than &mut self.

Yes, though I can think of some ways where the param could still be a mutable borrow.

Let's land this and investigate in future work.

PR #1500, opened 2020-01-21
`warp-drive` cache should have `Arc` variant

### LINKED ISSUE #1478: warp-drive cache should have `Arc` variant
Even though we are using `im.rs` cloning a Cache collection still clones it's inner elements.

Instead, we should wrap all collection items in `Arc` so cloning only increases a refcount.

nlinker:
Fixes #1478

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500)


iml-gui/crate/src/components/tree.rs
```diff
@@ -752,6 +756,60 @@ mod tests {
         )
     }
 
+    #[test]
+    fn test_cache_cloning() {
+        let c1 = fixtures::get_cache();
+        let mut c2: Cache = c1.clone();
+        let mut c3: Cache = c2.clone();
+
+        // The hashmap ost_pool_osts is the same for all cache clones c1, c2, c3
```
ip1981: You are basically testing what must be already tested in std library. Probably there is not value in these tests here.

iml-gui/crate/src/components/tree.rs
```diff
@@ -752,6 +756,60 @@ mod tests {
...
+            id: 2,
+            ostpool_id: 2,
+            managedost_id: 2,
+        };
+        let rec18 = OstPoolOstsRecord {
+            id: 18,
+            ostpool_id: 18,
+            managedost_id: 13,
+        };
+        let rec19 = OstPoolOstsRecord {
```
ip1981: This and rec18 look like a copy of the fixture. I think it is better to have only one source of truth.

ip1981:
Out of curiosity, are there places where cloning happens and `im` does not help?

iml-gui/crate/src/components/tree.rs
nlinker: Yep, agree.

iml-gui/crate/src/components/activity_indicator.rs
```diff
@@ -3,8 +3,9 @@ use im::HashMap;
 use iml_wire_types::{Alert, AlertSeverity};
 use seed::{prelude::*, *};
 use std::cmp::max;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
nlinker: Okay

iml-gui/crate/src/components/tree.rs
nlinker: I'd leave it here.
This (among the other things) shows the common usage pattern of the data structure similar to the `Cache` (it took some time for me and I'm not sure I can figure it out easily after a half of a year).

Also it shows the common invariants to expect when working with such data structures.

jgrund:
> Out of curiosity, are there places where cloning happens and im does not help?

Yes, it appears every element of an immutable collection is cloned if it isn't `Copy`  when the parent is cloned.

I tested this by implementing a custom `Clone` impl that printed to stdout when it was cloned. After placing it in an im collection and cloning the top-level collection, the custom clone was called.

ip1981:
That's right that an element is cloned. I meant, do we clone the collection itself? In this PR all touched places only involve taking a reference (`&`). Do I miss something?

jgrund:
> I meant, do we clone the collection itself?

That's the plan, once we have something in place that doesn't clone each element of a collection.

There is a need to pass either entire collections of subsets of one to child components. I think the important part is that each element needs to be wrapped in `Arc` not the collection itself.

nlinker:
___
*[iml-gui/crate/src/lib.rs, line 307 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz6ZIcgEOkAYdeWctFw:-Lz7m1SU9jVK_Zp8lD8V:bh84cv5) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/lib.rs#L307)):*

`make_mut` implements copy-on-write, is it what we want?
</blockquote></details>

Yes, it is intentional. The collections from `im` implement efficient `Clone`.

nlinker:
___
*[iml-gui/crate/src/components/tree.rs, line 790 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz5kzUxCg6kSWg8hdpa-r1-790:-Lz7mTRQ2RmOiSQ2AaFw:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/components/tree.rs#L790)):*

Ok, if you want :-)
</blockquote></details>

Done.

nlinker:
___
*[iml-gui/crate/src/components/tree.rs, line 765 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz5kzUxCg6kSWg8hdpa-r1-765:-Lz7mPJlC948mtoMGG_9:b-ozg8vm) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/components/tree.rs#L765)):*

I'd leave it here.
This (among the other things) shows the common usage pattern of the data structure similar to the `Cache` (it took some time for me and I'm not sure I can figure it out easily after a half of a year).

Also it shows the common invariants to expect when working with such data structures.
</blockquote></details>

@mkpankov @jgrund You also see no value in the test, right?

iml-gui/crate/src/components/tree.rs
ip1981: But that's in [the documentation](https://doc.rust-lang.org/std/sync/struct.Arc.html) :)

nlinker:
___
*[iml-gui/crate/src/lib.rs, line 307 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz6ZIcgEOkAYdeWctFw:-Lz8-_YN9-dN6G9a8FDu:b-3r63xn) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/lib.rs#L307)):*

I mean, it does clone the body so that it becomes two distinct collections. Maybe what was intended is `Arc::get_mut` to avoid splitting the cache by accident?
</blockquote></details>

I see your warning, the name `make_mut` might suggest that it performs the modification in place.
This was one of the reason, why I created the test, that explicitly asserts, that the collection is cloned and modified independently through different `Arc`-s.

nlinker:
___
*[iml-gui/crate/src/components/tree.rs, line 765 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz5kzUxCg6kSWg8hdpa-r1-765:-Lz812ga4ER59TPnWmcC:bsis6u8) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/components/tree.rs#L765)):*

I'm split on this one. On one hand it's not testing the `Arc`, it tests and documents applied usage in conjunction with `im`. On the other hand, there really is no code that we wrote. This is documentation example or standalone example at best.
</blockquote></details>

I will move it in some neutral place, since it is not related with `tree` component indeed.

ip1981:
> I'm split on this one. On one hand it's not testing the `Arc`, it tests and documents applied usage in conjunction with `im`. On the other hand, there really is no code that we wrote. This is documentation example or standalone example at best.

Even more, `im` or not `im` does not make any difference, this is all about the Clone trait and `.clone()` called [at the right time](https://doc.rust-lang.org/src/alloc/sync.rs.html#1006-1056) :) This is how traits, strict static typing, etc. help avoiding writing excessive tests.

nlinker:
___
*[iml-gui/crate/src/lib.rs, line 307 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz6ZIcgEOkAYdeWctFw:-LzD1uphANGSK1l7-aVV:b-s857e1) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/lib.rs#L307)):*

So we actually want distinct clones? I don't get it. Does it have to do with how `im` provides immutability?

I don't understand why do we want independent clones in each and every place that accesses the cache.
</blockquote></details>

This sounds reasonable, however this means I need to wrap everything in `RwLock`. Okay...

jgrund:
___
*[iml-gui/crate/src/lib.rs, line 307 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz6ZIcgEOkAYdeWctFw:-LzDFr6P40A_v2EiL6W1:bacy67i) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/lib.rs#L307)):*

This sounds reasonable, however this means I need to wrap everything in `RwLock`. Okay...
</blockquote></details>

To be clear, it's the elements inside the collection that should be wrapped in `Arc`, not the collection itself.

That is:

`model.records.active_alert.insert(x.id, Arc::new(x));`

nlinker:
___
*[iml-wire-types/src/lib.rs, line 2201 at r3](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lzb-S8D9Wffbgo-5igD:-Lzjka5o67E1J9U_lb6Y:b86e4wt) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/12c39c8bb404fc9a7420f42d64f45ddbf227f5d1/iml-wire-types/src/lib.rs#L2201)):*

Why did you decide to move it to separate file in this PR?
</blockquote></details>

Because the file `wire-types.rs` became over 2500 lines long, and the new cache types added even more. I thought the `warp_drive` related data types are worth to be separated.

Even now `lib.rs` has the `db` submodule, that is also good candidate to pulled out, I think (however this is beyond the scope of the task, of course)

nlinker:
___
*[iml-gui/crate/src/lib.rs, line 631 at r3](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-LzazfvK9M8bnYZIexYB:-LzjkWeZEb1oiaE_9arE:bhoq7ml) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/12c39c8bb404fc9a7420f42d64f45ddbf227f5d1/iml-gui/crate/src/lib.rs#L631)):*

At what point caches become independent? On `.clone`? Isn't that something we wanted to avoid?
</blockquote></details>

Frankly speaking it is not a shared cache (where everyone goes after the data), it is a snapshot, that is loaded into IML-GUI at the moment
https://github.com/whamcloud/integrated-manager-for-lustre/blob/1478-wd-cache-arc/iml-gui/crate/src/lib.rs#L329

Then the model goes through all corners down the components, and if a component wants to change an entry, it makes the change (the changed cache is still local for the component) and then pushes the changes back, e.g.:
```
                warp_drive::Record::Filesystem(x) => {
                    let id = x.id;
                    if model.records.filesystem.insert(x.id, Arc::new(x)).is_none() {
                        orders
                            .proxy(Msg::Tree)
                            .send_msg(tree::Msg::Add(warp_drive::RecordId::Filesystem(id)));
                    };
                }
```

So, probably the better term for this storage is _Snapshot_, but I realise it would be to many changes for it and it seems we are get used to the term _Cache_.

nlinker:
___
*[iml-gui/crate/src/lib.rs, line 631 at r3](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-LzazfvK9M8bnYZIexYB:-Lzk9WyR0cUi3iyA_0BR:bnwxo71) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/12c39c8bb404fc9a7420f42d64f45ddbf227f5d1/iml-gui/crate/src/lib.rs#L631)):*
> .. for it ..
.. to rename it ..

nlinker:
___
*[iml-gui/crate/src/components/tree.rs, line 765 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz5kzUxCg6kSWg8hdpa-r1-765:-Lzk9n99DwE24l5hRsdO:bb6un1j) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/6e5ef2db08d10dbaa6325e6ec10f085026109922/iml-gui/crate/src/components/tree.rs#L765)):*

I will move it in some neutral place, since it is not related with `tree` component indeed.
</blockquote></details>

Done

iml-wire-types/src/lib.rs
```diff
@@ -2138,9 +2138,11 @@ impl fmt::Display for ServiceState {
...
 #[derive(serde::Deserialize, serde::Serialize, PartialEq, Clone, Debug)]
 pub struct ResourceAgentType {
-    pub standard: String,         // e.g. ocf, lsb, stonith, etc..
-    pub provider: Option<String>, // e.g. heartbeat, lustre, chroma
-    pub ocftype: String,          // e.g. Lustre, ZFS
+    pub standard: String,
+    // e.g. ocf, lsb, stonith, etc..
+    pub provider: Option<String>,
+    // e.g. heartbeat, lustre, chroma
+    pub ocftype: String, // e.g. Lustre, ZFS
```
jgrund: Inconsistent formatting.

iml-wire-types/src/warp_drive.rs
```diff
@@ -0,0 +1,335 @@
+use crate::{
```
jgrund: Nice. That file was getting too large :)

iml-gui/crate/src/lib.rs
```diff
@@ -652,3 +657,64 @@ pub fn run() {
 
     log!("App started.");
 }
+
+#[cfg(test)]
+mod tests {
```
jgrund: Seems like this should live in `iml-wire-types/src/warp_drive.rs`

iml-wire-types/Cargo.toml
```diff
@@ -6,12 +6,8 @@ version = "0.2.0"
 authors = ["IML Team <iml@whamcloud.com>"]
 edition = "2018"
 
-[lib]
-path = "wire-types.rs"
-
-
 [dependencies]
-serde = { version = "1", features = ["derive"] }
+serde = { version = "1", features = ["derive", "rc"] }
```
jgrund: We don't want the rc feature anymore.

nlinker:
___
*[iml-gui/crate/src/lib.rs, line 662 at r7](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1500#-Lz5kzUxCg6kSWg8hdpb-r7-662:-Lzr1NCWEpb1cyo-oGGg:b-896fix) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/8cef4f2ac14ebba4bb6040fdd06e65416ce88c10/iml-gui/crate/src/lib.rs#L662)):*

Seems like this should live in `iml-wire-types/src/warp_drive.rs`
</blockquote></details>

Done.

iml-wire-types/src/warp_drive.rs
```diff
@@ -0,0 +1,335 @@
...
+    K: std::hash::Hash + Eq + Copy,
+    V: Clone,
+{
+    hm.iter().map(|(k, v)| (*k, (**v).clone())).collect()
+}
+
+/// The current state of locks based on data from the locks queue
+pub type Locks = HashMap<String, HashSet<LockChange>>;
+
+/// Do not forget to keep [`ArcCache`](iml_wire_types::warp_drive::ArcCache) in sync
```
ip1981: First thing to do is to forget ;-)

iml-wire-types/src/warp_drive.rs
```diff
@@ -0,0 +1,335 @@
...
+                self.volume.insert(x.id, x);
+            }
+            Record::VolumeNode(x) => {
+                self.volume_node.insert(x.id(), x);
+            }
+        }
+    }
+}
+
+impl ArcCache {
```
ip1981: There is a lot of duplication. Is old `Cache` used anywhere?

iml-wire-types/src/warp_drive.rs
```diff
@@ -0,0 +1,394 @@
...
+///     use iml_wire_types::warp_drive::ArcValuesExt;
+///
+///     let hm: HashMap<i32, Arc<String>> = im::hashmap!(
+///         1 => Arc::new("one".to_string()),
+///         2 => Arc::new("two".to_string()),
+///     );
+///     let vals: Vec<&str> = hm.arc_values().map(|s| &s[..]).collect();
+///     println!("values are: {:?}", vals);
+/// ```
+pub trait ArcValuesExt<K, V> {
```
ip1981: Did [Deref behavior](https://doc.rust-lang.org/std/sync/struct.Arc.html#deref-behavior) not work out well?

jgrund:
Looks like the warp drive tests are looking for a missing fixtures module.

ip1981:
> > Did [Deref behavior](https://doc.rust-lang.org/std/sync/struct.Arc.html#deref-behavior) not work out well?
> 
> Not really, one should write something like `map(|x| &**x)` which is cryptic enough to scare people 

If I saw just `&**x` I would be wondering, but with an additional trait I am wondering more ;-)

iml-gui/crate/src/components/activity_indicator.rs
```diff
@@ -1,13 +1,13 @@
...
-use std::cmp::max;
+use std::{cmp::max, sync::Arc};
 
-pub fn update_activity_health(active_alert: &HashMap<u32, Alert>) -> ActivityHealth {
+pub fn update_activity_health(active_alert: &HashMap<u32, Arc<Alert>>) -> ActivityHealth {
     active_alert
-        .values()
-        .filter(|x| x.severity > AlertSeverity::INFO)
+        .arc_values()
+        .filter(|x: &&Alert| x.severity > AlertSeverity::INFO)
```
ip1981: `&&Alert` looks suspicious.

nlinker: What do you propose?

iml-gui/crate/src/components/tree.rs
```diff
@@ -51,24 +52,28 @@ fn get_targets_by_parent_resource(
...
 fn get_targets_by_fs_id(
-    xs: &im::HashMap<u32, Target<TargetConfParam>>,
+    xs: &im::HashMap<u32, Arc<Target<TargetConfParam>>>,
     fs_id: u32,
     kind: TargetKind,
 ) -> Vec<&Target<TargetConfParam>> {
-    xs.values()
-        .filter(|x| match kind {
+    xs.arc_values()
+        .filter(|x: &&Target<TargetConfParam>| match kind {
```
ip1981: It is not related to this PR, but to me it looks like `kind` is matched against `TargetKind` many times, instead of just one.

jgrund: Suggestion?

ip1981:
> I don't think we should be {de},serializing `Arc` as from my reading of it there are some stateful issues doing that.

What issues exactly?

> 
> Instead, I think we'd want to two structs `Cache` and `ArcCache` (or something along those lines), where `ArcCache` is just `Cache` with each collection element wrapped in `Arc` (ex: `pub active_alert: HashMap<u32, Arc<Alert>>,`).
> 
> Then we could implement `From<Cache> for ArcCache` and when the initial load happens in the GUI, we can do the conversion.

Technically, what is currently done just re-implements [serde's](https://github.com/serde-rs/serde/blob/a00aee14950baca7de2e334b895e203b013712da/serde/src/de/impls.rs#L1707), that is: clone and  wrap into `Arc`.

Using serde's `rc` would require much less to code.

jgrund:
> What issues exactly?

https://serde.rs/feature-flags.html#-features-rc

> Opt into impls for Rc<T> and Arc<T>. Serializing and deserializing these types does not preserve identity and may result in multiple copies of the same data. Be sure that this is what you want before enabling this feature.
> 
> Serializing a data structure containing reference-counted pointers will serialize a copy of the inner value of the pointer each time a pointer is referenced within the data structure. Serialization will not attempt to deduplicate these repeated data.
> 
> Deserializing a data structure containing reference-counted pointers will not attempt to deduplicate references to the same data. Every deserialized pointer will end up with a strong count of 1.

ip1981:
> > What issues exactly?
> 
> https://serde.rs/feature-flags.html#-features-rc

I have it read, but can't see how this affects us. When we have just de-serialized `Cache` all its data is separate, without identity preserved, no? Then we just make a clone of each.

jgrund:
> I have it read, but can't see how this affects us. When we have just de-serialized Cache all its data is separate, without identity preserved, no? Then we just make a clone of each.

The part about it that feels iffy to me is that it's serializing / deserializing a smart pointer including it's internal references (according to that blurb). 

At the very least this isn't very well described, and if we rely on the feature, we are subject to any further changes they make in the lib.

Have you tried serializing an Arc Cache value to JSON? It would be useful to see it's wire representation at the very least.

ip1981:
Let me check that. Do we use serialisation of Cache at all? Deserialisation work as expected according to the tests :)

jgrund:
> Do we use serialisation of Cache at all?

Yes, it is serialized from the `iml-warp-drive` service, which runs on the manager node and sent to `iml-gui` on page load.

Another facet of this is that using `Arc` is not needed at all on the server-side. It's only the client that needs it. That means we would only be using the rc feature for less code, at the expense of the cache being wrapped in Arc on the server side (which it's currently not).

ip1981:
> Another facet of this is that using `Arc` is not needed at all on the server-side. It's only the client that needs it. That means we would only be using the rc feature for less code, at the expense of the cache being wrapped in Arc on the server side (which it's currently not).

Caramba! :) This really can justify existence of  two structs.

iml-gui/crate/src/components/tree.rs
ip1981: https://github.com/whamcloud/integrated-manager-for-lustre/pull/1572 :)

PR #1506, opened 2020-01-21
Create standalone iml-api service

jgrund:
Create a standalone iml-api service written in rust.

To start it will return some useful configuration info but should be
usable for any generic API task.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1506)


jgrund:
Screen Shot 2020-01-21 at 4 32 59 PM

Working call

jgrund:
___
*[chroma-manager.conf.template, line 184 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1506#-LzGgy-n1bbhyJzCM60s:-LzHfenW0xkE2L0NjhFG:b-euwyrz) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/3b6ef7c3c56d43c84367215294b046f9b77f0436/chroma-manager.conf.template#L184)):*

This is for new API service, and `/api` below is for older `chroma` API? Feels like special-casing `/conf` endpoint, maybe point to something like `/api/v2/` instead and `/conf` should be handled automatically as a sub-path?
</blockquote></details>

Then consumers  would need to know what lives at /api and what lives at api/v2 and call the right one. What benefit does that bring besides de-duping the overloads in the nginx conf?

johnsonw:
I've seen other services like github have multiple API versions but I think that was for backwards compatibility. We don't support rolling back to previous versions so I think going forward it doesn't benefit to have multiple api versions.

iml-api/src/main.rs
```diff
@@ -0,0 +1,29 @@
...
+        is_release: iml_manager_env::get_is_release(),
+    };
+
+    let subscriber = Subscriber::builder()
+        .with_env_filter(EnvFilter::from_default_env())
+        .finish();
+
+    tracing::subscriber::set_global_default(subscriber).unwrap();
+
+    let routes = warp::path("conf").map(move || warp::reply::json(&conf));
```
johnsonw: So in this first stage the api is just returning basic values from the manager environment, correct? Will this eventually start talking to the database via diesel?

docker/iml-api.dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM rust-iml-base as builder
+
+FROM ubuntu
+COPY --from=builder /build/target/release/iml-api /usr/local/bin
+
+RUN apt-get update \
+    && apt-get install -y curl
+
+COPY docker/wait-for-dependencies.sh /usr/local/bin/
+ENTRYPOINT [ "wait-for-dependencies.sh" ]
```
johnsonw: I didn't see it in here yet but I figure this service will eventually use postgres. If that's the case should we instead change this to:
```
ENTRYPOINT [ "wait-for-dependencies-postgres.sh" ]
```
?

johnsonw:
I think this is missing an entry in docker-compose.override.yml

docker/iml-api.dockerfile
jgrund: I think we can wait until we introduce a endpoint that depends on postgres.

iml-api/src/main.rs
jgrund: Yes, that's the plan.

docker/docker-compose.yml
```diff
@@ -253,6 +253,18 @@ services:
       - PROXY_HOST=iml-action-runner
       - ACTION_RUNNER_PORT=8009
       - RUST_LOG=info
+  iml-api:
+    image: "imlteam/iml-api:5.1"
```
johnsonw: should this be 5.1.1-dev?

docker/iml-api.dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM rust-iml-base as builder
+
+FROM ubuntu
```
johnsonw: I haven't looked at the size difference but may want to consider using centos:7

docker/docker-compose.yml
jgrund: yup, I'll update before landing.

docker/iml-api.dockerfile
jgrund: yes, waiting on your changes :)

jgrund:
Landed as 6ee4185c2c64a0b615cb99edad5b83c3c18413ee

PR #1510, opened 2020-01-21
Account for missing FS in iml-ostpool service

### LINKED ISSUE #1509: Ostpool manager service does not account for existing fs
When the ostpool daemon plugin detects a fs, it sends it over to the manager side plugin. 

That plugin ends up calling this fn to resolve the fs name to a fs id:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/3875f5dbce20618643eb6da62396dd3fd52f4be3/iml-services/iml-ostpool/src/db.rs#L56

However, what isn't accounted for is that a fs may already exist but IML does not yet know about it (monitored mode setup for instance).

In this case the fn ends up hitting this case and returning an error:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/3875f5dbce20618643eb6da62396dd3fd52f4be3/iml-services/iml-ostpool/src/db.rs#L68

That error is not handled and is propagated up to the main fn where it ends up crashing the process:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/3875f5dbce20618643eb6da62396dd3fd52f4be3/iml-services/iml-ostpool/src/main.rs#L34

Since the IML target will fail if one of it's services fail, this causes all services to restart again.

And since this will always happen in monitored mode with an existing fs, the services restart over and over in a loop.

utopiabound:
Fixes #1509 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1510)

### DIFF
--- a/iml-services/iml-ostpool/src/main.rs
+++ b/iml-services/iml-ostpool/src/main.rs
@@ -31,7 +31,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             let client = client.clone();
 
             async move {
-                let fsid = client.fsid(&fsname).await?;
+                // If fsid finds no match, assume FS hasn't been added to IML and skip
+                let fsid = match client.fsid(&fsname).await {
+                    Ok(id) => id,
+                    Err(Error::NotFound) => {
+                        tracing::info!("Filesystem {} not found in DB", &fsname);
+                        return Ok(vec![]);
+                    }
+                    Err(e) => return Err(e),
+                };
                 tracing::debug!("{}: {:?}", fsname, newpoolset);
 
                 let oldpoolset = client.poolset(fsname, fsid).await?;


iml-services/iml-ostpool/src/main.rs
```diff
@@ -31,7 +31,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
             let client = client.clone();
 
             async move {
-                let fsid = client.fsid(&fsname).await?;
+                // If fsid finds no match, assume FS hasn't been added to IML and skip
+                let fsid = match client.fsid(&fsname).await {
+                    Ok(id) => id,
+                    Err(Error::NotFound) => {
```
jgrund: nit. `Err(Error::NotFound)` seems like `None` to me, I wonder if this fn should return `Option<Result<.., ..>>` instead.

PR #1513, opened 2020-01-22
Remove parking_lot from Session

mkpankov:
It doesn't currently compile

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1513)

### DIFF
--- a/iml-agent/src/http_comms/session.rs
+++ b/iml-agent/src/http_comms/session.rs
@@ -3,7 +3,7 @@
 // license that can be found in the LICENSE file.
 
 use crate::{
-    agent_error::{ImlAgentError, Result},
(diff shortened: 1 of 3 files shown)


mkpankov:
So this is kinda what it gets to look
https://github.com/whamcloud/integrated-manager-for-lustre/pull/1513/files#diff-097c3716dcec898101971ec6913b99edR194

And it doesn't even compile because `async move { }` creates a future we don't want to resolve inside of `map`. Only way I see is make `addon_info` operate on future of session info instead of reference. @jgrund Do you think I should take that path?

jgrund:
Superseded by #1846

PR #1519, opened 2020-01-23
Add media breakpoint tracking

jgrund:
Add a new module that will track current breakpoint size via the window resize event.

This can be used to handle responsive rendering changes in Rust code.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1519)

### DIFF
--- a/iml-gui/crate/src/breakpoints.rs
+++ b/iml-gui/crate/src/breakpoints.rs
@@ -6,26 +6,21 @@
 
 use seed::window;
 
-pub(crate) const SM: f64 = 569.;
-pub(crate) const MD: f64 = 769.;
-pub(crate) const LG: f64 = 1025.;
-pub(crate) const XL: f64 = 1701.;
-
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
 pub(crate) enum Size {
-    XS,
-    SM,
-    MD,
-    LG,
-    XL,
+    XS = 0,
+    SM = 569,
+    MD = 769,
+    LG = 1025,
+    XL = 1701,
 }
 
-fn inner_width() -> f64 {
+fn inner_width() -> u64 {
     window()
         .inner_width()
         .expect("Could not get inner_width")
         .as_f64()
-        .expect("Could not parse inner_width to f64")
+        .expect("Could not parse inner_width to f64") as u64
 }
 
 /// Returns *maximum* matching breakpoint based on `window.inner_width`.
@@ -33,13 +28,13 @@ fn inner_width() -> f64 {
 pub(crate) fn size() -> Size {
     let w = inner_width();
 
-    if w >= XL {
+    if w >= Size::XL as u64 {
         Size::XL
-    } else if w >= LG {
+    } else if w >= Size::LG as u64 {
         Size::LG
-    } else if w >= MD {
+    } else if w >= Size::MD as u64 {
         Size::MD
-    } else if w >= SM {
+    } else if w >= Size::SM as u64 {
         Size::SM
     } else {
         Size::XS
--- a/iml-gui/crate/src/breakpoints.rs
+++ b/iml-gui/crate/src/breakpoints.rs
@@ -1,0 +1,47 @@
+//! # Breakpoints
+//!
+//! This module lists the existing media breakpoints we use and contains some helper functions for working with them.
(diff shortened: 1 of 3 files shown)


jgrund:
This also fixes a bug where after clicking the hamburger button with the md breakpoint, the menu would not show with the lg breakpoint.

resize

iml-gui/crate/src/breakpoints.rs
```diff
@@ -0,0 +1,47 @@
...
+        .as_f64()
+        .expect("Could not parse inner_width to f64")
+}
+
+/// Returns *maximum* matching breakpoint based on `window.inner_width`.
+/// This models the behavior of breakpoints in tailwind: https://tailwindcss.com/docs/responsive-design/
+pub(crate) fn size() -> Size {
+    let w = inner_width();
+
+    if w >= XL {
```
johnsonw: Any reason not to use a pattern match?

jgrund: How? It's doing an ordering comparision, not an equal one.

johnsonw: meh, you could use a match guard but I guess it doesn't buy anything.

jgrund: Yeah, I thought that might be what you meant. That would look noisier than just an if / else imo.

johnsonw: yes, I agree.

iml-gui/crate/src/breakpoints.rs
```diff
@@ -0,0 +1,47 @@
...
+use seed::window;
+
+pub(crate) const SM: f64 = 569.;
+pub(crate) const MD: f64 = 769.;
+pub(crate) const LG: f64 = 1025.;
+pub(crate) const XL: f64 = 1701.;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
+pub(crate) enum Size {
+    XS,
```
ip1981: I wonder if this could be `enum Size { XS = 0, SM = 569,...}` and so on.

jgrund: Hrrm, didn't consider that, looking.

jgrund: So, this *does* work:

```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd)]
pub(crate) enum Size {
    XS = 0,
    SM = 569,
    MD = 769,
    LG = 1025,
    XL = 1701,
}
```

However any usage of the enum then needs to be double casted to compare with the f64 of `inner_width`:

`w >= Size::XL as u64 as f64`

Not sure it's worth the cost of callers having to do that each time over the current `const` bindings.

ip1981: Well, then update `inner_width` to return an int. Width is integer after all.

jgrund: yeah, ok.

iml-gui/crate/src/breakpoints.rs
```diff
@@ -6,40 +6,35 @@
...
+        .expect("Could not parse inner_width to f64") as u64
 }
 
 /// Returns *maximum* matching breakpoint based on `window.inner_width`.
 /// This models the behavior of breakpoints in tailwind: https://tailwindcss.com/docs/responsive-design/
 pub(crate) fn size() -> Size {
     let w = inner_width();
 
-    if w >= XL {
+    if w >= Size::XL as u64 {
```
ip1981: Doesn't it work without type casting? O_o

jgrund: Nope

jgrund: Which is what I was referring to earlier. However, the compiler / rust-analyzer will tell you if you don't cast, so I think it's ok.

ip1981: I feel something like regret :)

ip1981: Oh, maybe use `#[repr]` for the enum?

jgrund: Nope, still same issue with `repr` attribute.

PR #1521, opened 2020-01-23
Responsive fixups

jgrund:
Fixup header responsiveness

The header would previously break layout before it reached the md
breakpoint.

This patch fixes that.

responsive-fix

This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1521)


jgrund:
Landed as: 1773e4567ca3598d663781a490024bf82b38da73

PR #1526, opened 2020-01-25
Suppress excessive warnings in IML-GUI

nlinker:
There are lot of warnings from clippy and rustc, that are just noise, because they are mostly false positives. E.g `unused function` warning is emitted just because all the calls are inside `seed` macros.

This patch supresses these warnings, the only warning is left is
```
2230 |         pub fn remove_record(&mut self, x: &RecordId) -> bool {
     |                                            ^^^^^^^^^ help: consider passing by value instead: `RecordId`
```
which I'm going to fix in `Arc` PR.

I understand this is rather effect fix (not the reason), so rejecting is okay.

Signed-off-by: Nick Linker <nlinker@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1526)

### DIFF
--- a/iml-gui/crate/Cargo.toml
+++ b/iml-gui/crate/Cargo.toml
@@ -17,6 +17,7 @@ wasm-bindgen-test = "0.3"
 
 [dependencies]
 chrono = { version = "0.4", features = ["wasmbind"] }
+futures = "0.3"
 im = { version = "14.1", features = ["serde"] }
 iml-wire-types = { path = "../../iml-wire-types", version = "0.2" }
 js-sys = "0.3"
@@ -40,6 +41,7 @@ features = [
   "NotificationOptions",
   "NotificationPermission",
   "ServiceWorkerContainer",
+
   "ServiceWorkerRegistration",
   "Window",
 ]
--- a/iml-gui/crate/src/components/dropdown.rs
+++ b/iml-gui/crate/src/components/dropdown.rs
@@ -1,8 +1,7 @@
 use crate::{components::Placement, generated::css_classes::C};
 use seed::{prelude::*, virtual_dom::Attrs, Style, *};
 
-#[allow(unused)]
-pub(crate) fn wrapper_view<T>(attrs: Attrs, placement: Placement, open: bool, children: impl View<T>) -> Node<T> {
+pub fn wrapper_view<T>(attrs: Attrs, placement: Placement, open: bool, children: impl View<T>) -> Node<T> {
     if !open {
         return empty![];
     }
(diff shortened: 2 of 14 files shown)


iml-gui/crate/src/components/dropdown.rs
```diff
@@ -1,7 +1,7 @@
 use crate::{components::Placement, generated::css_classes::C};
 use seed::{prelude::*, virtual_dom::Attrs, Style, *};
 
-pub(crate) fn wrapper_view<T>(attrs: Attrs, placement: Placement, open: bool, children: impl View<T>) -> Node<T> {
+pub fn wrapper_view<T>(attrs: Attrs, placement: Placement, open: bool, children: impl View<T>) -> Node<T> {
```
ip1981: I think is it not used _yet_, but will be  soon.

iml-wire-types/wire-types.rs
```diff
@@ -2280,6 +2268,7 @@ pub mod warp_drive {
         }
     }
 
+    #[allow(clippy::large_enum_variant)]
```
jgrund: I feel like we should probably fix the `large_enum_variant` warnings, but I'm not pushing a change because I know you're working in this area.

jgrund:
As part of this patch, we should probably fail on clippy warnings as well.

iml-gui/crate/src/components/dropdown.rs
nlinker: Already
https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-gui/crate/src/components/paging.rs#L167

nlinker:
> I did not verify, but I hope that`pub(crate)` will help us spot and eliminate dead code (while plain `pub` won't).

The problem is that `pub(crate)` doesn't help either, because of these false positives.

nlinker:
> As part of this patch, we should probably fail on clippy warnings as well.

You mean to fire compilation error? And for now temporarily suppress the `RecordId` warning either?

jgrund:
> You mean to fire compilation error?

Yeah, to both fail early in the github action clippy check, and in the `cargo make` task that is run locally.

There is a flag you can pass to clippy to make warnings report as errors.

> And for now temporarily suppress the RecordId warning either?

Either that or fix it.

jgrund:
> The problem is that pub(crate) doesn't help either, because of these false positives.

In a way, they aren't false positives as they are tracking whether their usages are also being used (which is sort of confusing). I.E. it's tracking usage till the root of the dependency tree.

jgrund:
Just force-pushed to latest master and turned clippy warnings into errors.

iml-gui/crate/Cargo.toml
```diff
@@ -40,6 +41,7 @@ features = [
   "NotificationOptions",
   "NotificationPermission",
   "ServiceWorkerContainer",
+
```
ip1981:
```suggestion

```

jgrund:
Landed as: 52d1122a58ebf219b42272020af83b55fce4184b

PR #1538, opened 2020-01-30
Add lock indicator component

ip1981:
Use `:focus-within` pseudo-class to show a popover.

lock

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1538)

### DIFF
--- a/iml-gui/configs/tailwind.config.js
+++ b/iml-gui/configs/tailwind.config.js
@@ -211,8 +211,16 @@ module.exports = {
     }
   },
   variants: {
-    display: ["group-hover", "responsive"],
+    display: ["group-hover", "group-focus", "responsive"],
     textColor: ["hover", "group-hover", "focus"]
   },
-  plugins: []
+  plugins: [
+    function({ addVariant }) {
+      addVariant('group-focus', ({ container, separator }) => {
+        container.walkRules(rule => {
+          rule.selector = `.group:focus-within .group-focus\\:${rule.selector.slice(1)}`
+        })
+      })
+    }
+  ]
 };
--- a/iml-gui/configs/webpack.config.js
+++ b/iml-gui/configs/webpack.config.js
@@ -55,7 +55,8 @@ module.exports = (env, argv) => {
       }),
       // Inline the critical part of styles, preload remainder.
       new Critters({
-        logLevel: "warn",
+        // Values below error produce some noise => https://github.com/GoogleChromeLabs/critters/issues/51
+        logLevel: "error",
         // https://github.com/GoogleChromeLabs/critters/issues/34
         pruneSource: false
       }),
--- a/iml-gui/crate/src/components/lock_indicator.rs
+++ b/iml-gui/crate/src/components/lock_indicator.rs
@@ -1,0 +1,62 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::{
+    components::{font_awesome, popover, tooltip, Placement},
+    generated::css_classes::C,
+};
+use im::HashSet;
+use iml_wire_types::{warp_drive::Locks, LockChange, LockType, ToCompositeId};
+use seed::{prelude::*, *};
+
+pub(crate) fn lock_indicator<T>(x: &dyn ToCompositeId, all_locks: &Locks) -> Node<T> {
+    match all_locks.get(&x.composite_id().to_string()) {
+        Some(locks) if !locks.is_empty() => indicator(locks),
+        _ => empty![],
+    }
+}
+
+fn indicator<T>(lcks: &HashSet<LockChange>) -> Node<T> {
+    let (rw, ro): (HashSet<&LockChange>, HashSet<&LockChange>) =
+        lcks.iter().partition(|l| l.lock_type == LockType::Write);
+
+    let mut tooltip = vec![];
+    let mut popup = vec![];
+
+    let rw_c = rw.len();
+    if rw_c > 0 {
+        if rw_c > 1 {
+            tooltip.push(format!("{} write locks", rw_c));
+        } else {
+            tooltip.push("1 write lock".to_string());
+        }
+        popup.push(popover::title_view("Write locks"));
+        popup.push(popover::content_view(mk_lock_list(&rw)));
+    }
+
+    let ro_c = ro.len();
+    if ro_c > 0 {
+        if ro_c > 1 {
+            tooltip.push(format!("{} read locks", ro_c));
+        } else {
+            tooltip.push("1 read lock".to_string());
+        }
+        popup.push(popover::title_view("Read locks"));
+        popup.push(popover::content_view(mk_lock_list(&ro)));
+    }
+
+    span![
+        class![C.group, C.cursor_pointer, C.relative],
+        attrs! {At::TabIndex => 0},
+        font_awesome(class![C.inline, C.w_4, C.h_4], "lock"),
+        tooltip::view(&tooltip.join(", "), Placement::Top),
+        popover::view(popup, Placement::Bottom)
+    ]
+}
+
+fn mk_lock_list<T>(locks: &HashSet<&LockChange>) -> Node<T> {
+    let mut items: Vec<&String> = locks.iter().map(|x| &x.description).collect();
+    items.sort_unstable();
+    ul![class![C.list_disc, C.px_4], items.iter().map(|x| li![x])]
+}
--- a/iml-gui/crate/src/generated/css_classes.rs
+++ b/iml-gui/crate/src/generated/css_classes.rs
@@ -3913,6 +3913,15 @@ pub struct CssClasses<'a> {
         display: table-row;
         display: table-cell;
         display: none;
+        display: block;
+        display: inline-block;
+        display: inline;
+        display: flex;
+        display: inline-flex;
+        display: table;
+        display: table-row;
+        display: table-cell;
+        display: none;
         color: transparent;
         color: #000;
         color: #FFF;
@@ -4018,6 +4027,24 @@ pub struct CssClasses<'a> {
         display: table-row;    @media (min-width: 569px)
         display: table-cell;    @media (min-width: 569px)
         display: none;    @media (min-width: 569px)
+        display: block;    @media (min-width: 569px)
+        display: inline-block;    @media (min-width: 569px)
+        display: inline;    @media (min-width: 569px)
+        display: flex;    @media (min-width: 569px)
+        display: inline-flex;    @media (min-width: 569px)
+        display: table;    @media (min-width: 569px)
+        display: table-row;    @media (min-width: 569px)
+        display: table-cell;    @media (min-width: 569px)
+        display: none;    @media (min-width: 569px)
+        display: block;    @media (min-width: 769px)
+        display: inline-block;    @media (min-width: 769px)
+        display: inline;    @media (min-width: 769px)
+        display: flex;    @media (min-width: 769px)
+        display: inline-flex;    @media (min-width: 769px)
+        display: table;    @media (min-width: 769px)
+        display: table-row;    @media (min-width: 769px)
+        display: table-cell;    @media (min-width: 769px)
+        display: none;    @media (min-width: 769px)
         display: block;    @media (min-width: 769px)
         display: inline-block;    @media (min-width: 769px)
         display: inline;    @media (min-width: 769px)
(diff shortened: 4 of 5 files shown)


iml-gui/configs/tailwind.config.js
```diff
@@ -211,8 +211,16 @@ module.exports = {
...
-    display: ["group-hover", "responsive"],
+    display: ["group-hover", "group-focus", "responsive"],
     textColor: ["hover", "group-hover", "focus"]
   },
-  plugins: []
+  plugins: [
+    function({ addVariant }) {
+      addVariant('group-focus', ({ container, separator }) => {
+        container.walkRules(rule => {
+          rule.selector = `.group:focus-within .group-focus\\:${rule.selector.slice(1)}`
```
ip1981: There is some noise:
```
⚠ ｢Critters｣: 45 rules skipped due to selector errors:
  .group:focus-within .group-focus\ -> '.group:focus-within .group-focus\' is not a valid selector
  .group:focus-within .group-focus\ -> '.group:focus-within .group-focus\' is not a valid selector
  .group:focus-within .group-focus\ -> '.group:focus-within .group-focus\' is not a valid selector
  .group:focus-within .group-focus\ -> '.group:focus-within .group-focus\' is not a valid selector
  .group:focus-within .group-focus\ -> '.group:focus-within .group-focus\' is not a valid selector
  .group:focus-within .group-focus\ -> '.group:focus-within .group-focus\' is not a valid selector
  .group:focus-within .group-focus\ -> '.group:focus-within .group-focus\' is not a valid selector
...
```

iml-gui/crate/src/generated/css_classes.rs
```diff
@@ -3904,6 +3906,15 @@ pub struct CssClasses<'a> {
     pub hidden: &'a str,
 
     /**
+        display: block;
```
jgrund: Some sort of bug in the codgen? A lot of these look repeated.

ip1981: I think so.

iml-gui/configs/tailwind.config.js
jgrund: If you update plugins to look like this, it will remove the warnings:

```
  plugins: [
    ({ addVariant, config }) => {
      const prefixClass = function(className) {
        const prefix = config("prefix");
        const getPrefix = typeof prefix === "function" ? prefix : () => prefix;
        return `${getPrefix(`.${className}`)}${className}`;
      };

      const groupPseudoClassVariant = function(pseudoClass) {
        return ({ modifySelectors, separator }) => {
          return modifySelectors(({ selector }) => {
            return selectorParser(selectors => {
              selectors.walkClasses(classNode => {
                classNode.value = `group-${pseudoClass}${separator}${classNode.value}`;
                classNode.parent.insertBefore(
                  classNode,
                  selectorParser().astSync(
                    `.${prefixClass("group")}:${pseudoClass} `
                  )
                );
              });
            }).processSync(selector);
          });
        };
      };

      addVariant("group-focus", groupPseudoClassVariant("focus"));
    }
  ]
```

You will need to add `postcss-selector-parser` as a devDependency and require it at the top of this file as well:

`const selectorParser = require("postcss-selector-parser");`

iml-gui/crate/src/components/lock_indicator.rs
```diff
@@ -0,0 +1,61 @@
...
+                } else {
+                    tooltip.push("1 read lock".to_string());
+                }
+                popup.push(popover::title_view("Read locks"));
+                popup.push(popover::content_view(mk_lock_list(&ro)));
+            }
+
+            return span![
+                class![C.group, C.cursor_pointer, C.relative],
+                attrs! {At::TabIndex => 0},
```
jgrund: perhaps roll this and the above class block into a single fn that returns these in popover.rs. That way we won't need to repeat them each time. ref:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/1b14eced92c102fddc333232706eb78f14964c97/iml-gui/crate/src/components/tooltip.rs#L7-L12

iml-gui/crate/src/generated/css_classes.rs
jgrund: Is there anything we can do on our end to not have these added?

iml-gui/configs/tailwind.config.js
ip1981: No, the warning is triggered by `focus-within`. It looks like you used `focus` and it worked well. If I change my code to use `focus` the warning will be gone too.

Tailwind uses `focus-within` in other places without [visible] issues. Something else is broken, probably Critters => https://github.com/GoogleChromeLabs/critters/issues/47

iml-gui/crate/src/components/lock_indicator.rs
```diff
@@ -0,0 +1,61 @@
...
+    components::{font_awesome, popover, tooltip, Placement},
+    generated::css_classes::C,
+};
+use im::HashSet;
+use iml_wire_types::{warp_drive::Locks, LockChange, LockType, ToCompositeId};
+use seed::{prelude::*, *};
+
+pub(crate) fn lock_indicator<T>(x: &dyn ToCompositeId, locks: &Locks) -> Node<T> {
+    if let Some(lcks) = locks.get(&x.composite_id().to_string()) {
+        if !lcks.is_empty() {
```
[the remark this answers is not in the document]
ip1981: Can't wait for https://github.com/rust-lang/rust/issues/53667 :)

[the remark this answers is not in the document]
ip1981: Refactored. I generally try to have the least possible number of exist points. In this case it is two.

ip1981:
> > Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
> > _[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1538)_ status: all files reviewed, 4 unresolved discussions (waiting on @ip1981 and @jgrund)
> 
> _[iml-gui/crate/src/components/lock_indicator.rs, line 15 at r2](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1538#-LzqieQDVhrdsiR_8-DY-r2-15:-M-F2TpJE3PZzOr1_1lJ:b2vb1uz) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/7056ce4d375265133442d05faa6db408b634f9e9/iml-gui/crate/src/components/lock_indicator.rs#L15)):_
> Previously, jgrund (Joe Grund) wrote…
> 
> I'd rename also `lcks` to `locks`:
> 
> ```rust
> if let Some(locks) = locks.get(&x.composite_id().to_string()) { ... }
> ```
> 
> There is no need to mangle the name since the only inner `locks` is used inside `if` body.
> The similar code seems to be used in std, e.g. https://github.com/rust-lang/rust/blob/master/src/librustc/traits/select.rs#L2850-L2859

No, in that code there are simple `if`s, not `if lets`s. But I renamed to `all_locks` and `locks` :)

iml-gui/configs/tailwind.config.js
jgrund: Hrrm, why do we need focus-within?

iml-gui/crate/src/components/lock_indicator.rs
ip1981: The classes can be required for both a tooltip and a popup. I felt that it would be confusing and fragile to have only `tooltip::container()` or only `popup::container()`, or both, when the component have a popup and a tooltip, like in case of the lock indicator. I don't know what will be the best, at the moment I put it explicitly here.

jgrund: So, how about extracting to a common location and using for both. something like `group_container`.

iml-gui/configs/tailwind.config.js
ip1981: Because we need show one element, when the other have focus:

```html
<div class="parent">
<span class="input">element to get focus</span>
<span class="popup">popup</span>
</div>
```

I can think of two reliable ways of doing that:

```css
.parent:focus-within .popup {
//...
}
```

or:
```css
.input:focus ~ .popup {
//...
}
```

The latter AFAIK is not possible with  Tailwind. And `:focus` does not propagates to parents.

It is possible to use `.input:focus .popup`, but we would need to make a popup a child of the input, which is odd in general: elements that can be "focus" cannot have children.

ip1981:
> It seems you find shadowing to be dangerous

I don't like my mental cache reset ;)

iml-gui/configs/tailwind.config.js
jgrund: hrrm, What's strange is I tested this with just focus and it seemed to work fine. I'll try it again.

jgrund: Yeah, it works fine with just focus on the parent. 

focus-parent


`:focus-within` is for dealing with focused child selectors anyway (i.e. applying styles to a form when an input is focused), which is the opposite of what we want.

iml-gui/crate/src/components/lock_indicator.rs
```diff
@@ -0,0 +1,61 @@
...
+            let (rw, ro): (HashSet<&LockChange>, HashSet<&LockChange>) =
+                lcks.iter().partition(|l| l.lock_type == LockType::Write);
+
+            let mut tooltip = vec![];
+            let mut popup = vec![];
+
+            let rw_c = rw.len();
```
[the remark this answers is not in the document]
ip1981: I tried a lot to reduce duplicate, and I believe the result is the least evil, because what is happening here is plural forms with hypothetical i18n plus modifying two "global" structs. Even now the plural forms is not trivial (there are two of them :)) and, I think, this has to be done somehow in general.

Most importantly, tearing off some words, like "read" and "write" from the complete phrases is not worthwhile whatsoever. This could be something like "Write locks: 123", etc., but I prefer human-friendly UI :)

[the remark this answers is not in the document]
jgrund: Even if you were extract the whole strings (which would be the case in i18n), you'd still have some common code here. 

"{} write locks" / "{} read locks"
"1 write lock" / "1 read lock" 
"Write locks" / "Read locks"

are the variant strings, which would need to be pulled from some i18n lookup. If this were abstracted you would be pushing in the lookups instead of the strings based on r/w type. 

That said, I think DRYing this is a nit anyway.

iml-gui/crate/src/generated/css_classes.rs
ip1981: Only to have a local modified copy of the generator? => https://github.com/MartinKavik/postcss-typed-css-classes/blob/3ad711589d475f07daa28b1581d8aaade394f71a/generators/rust_generator.js#L129

iml-gui/crate/src/components/lock_indicator.rs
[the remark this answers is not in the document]
jgrund: I do agree we need to start thinking about i18n in general fwiw.

iml-gui/crate/src/generated/css_classes.rs
jgrund: Can you open an issue on their repo about it?

iml-gui/configs/tailwind.config.js
ip1981: Yes, but this feels like a hack, because not all elements, that can be focused, can have children, `input` can't, for instance. Although the only issue we have is that some tool is not updated yet for CSS4 :) `:focus-within` was introduced to overcome limitations of `:focus`, so that the "focus" state of a child were known to parents. And of course it can be used to style sibling, e. g. to highlight an input's label.

jgrund: > Yes, but this feels like a hack, because not all elements, that can be focused, can have children, input can't, for instance.

Sure, but I don't think we're going to be adding a popover to a input element that only opens on focus anytime soon anyway (and if we do, we can tackle that problem when it occurs).

This gets us over the hump of the functionality working the same for popover (which is the purpose of this patch) while not spamming the console with tons of warnings everytime we save a file.

ip1981: There is a logLevel option for Critters and we can suppress warnings, this will not affect functionality. I don't want to taint one thing just because another is lame :) I could question that Critters is useful at all in our case, but it is fine as is.

ip1981: So that chain looks this way: Critters -> [JSDOM](https://github.com/jsdom/jsdom/issues/2026#issuecomment-339050204) -> [CSSOM](https://github.com/NV/CSSOM/).

jgrund: It's sort of trading one workaround for another, but I can live with it. Should probably put the issue next to the logLevel bump so it can be tracked and revered in the future

iml-gui/configs/webpack.config.js
```diff
@@ -55,7 +55,7 @@ module.exports = (env, argv) => {
       }),
       // Inline the critical part of styles, preload remainder.
       new Critters({
-        logLevel: "warn",
+        logLevel: "error",
```
jgrund:
```suggestion
        // https://github.com/GoogleChromeLabs/critters/issues/47
        logLevel: "error",
```

iml-gui/crate/src/generated/css_classes.rs
ip1981: https://github.com/MartinKavik/postcss-typed-css-classes/issues/3

iml-gui/configs/tailwind.config.js
ip1981: I think I will open an issue for Critters.

iml-gui/crate/src/components/lock_indicator.rs
ip1981: Next thing => https://github.com/whamcloud/integrated-manager-for-lustre/issues/1558

iml-gui/crate/src/components/lock_indicator.rs
```diff
@@ -0,0 +1,62 @@
...
+fn indicator<T>(lcks: &HashSet<LockChange>) -> Node<T> {
+    let (rw, ro): (HashSet<&LockChange>, HashSet<&LockChange>) =
+        lcks.iter().partition(|l| l.lock_type == LockType::Write);
+
+    let mut tooltip = vec![];
+    let mut popup = vec![];
+
+    let rw_c = rw.len();
+    if rw_c > 0 {
+        if rw_c > 1 {
```
johnsonw: Nothing wrong with this but it might be nice to have a `pluralize` function somewhere:
```
tooltip.push(pluralize("{} write {}", rw_c, "lock"));
```

iml-gui/crate/src/components/lock_indicator.rs
```diff
@@ -0,0 +1,62 @@
...
+        } else {
+            tooltip.push("1 write lock".to_string());
+        }
+        popup.push(popover::title_view("Write locks"));
+        popup.push(popover::content_view(mk_lock_list(&rw)));
+    }
+
+    let ro_c = ro.len();
+    if ro_c > 0 {
+        if ro_c > 1 {
```
johnsonw: Another use case for pluralize

iml-gui/crate/src/components/lock_indicator.rs
ip1981: This should be something similar to ngettext.

PR #1542, opened 2020-01-30
Keep sosreports in archive format

jgrund:
Try to work around Jenkins archive step hanging by never expanding the
sosreports.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1542)

### DIFF
--- a/tests/integration/utils/chroma_log_collector.py
+++ b/tests/integration/utils/chroma_log_collector.py
@@ -177,32 +177,12 @@ class ChromaLogCollector(object):
 
         errors = []
 
-        # Copy and expand the diagnostics locally, so they will be
-        # able to be read in browser in Jenkins.
         errors.append(self.fetch_log(server, "/var/tmp/%s" % diagnostics, ""))
 
-        if diagnostics.endswith("tar.xz"):
-            if shell_run(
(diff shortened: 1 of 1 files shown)


tests/integration/utils/chroma_log_collector.py
```diff
@@ -177,33 +177,13 @@ def fetch_iml_diagnostics(self, server):
...
-                ["tar", "-xvzf", "%s/%s" % (self.destination_path, diagnostics), "-C", self.destination_path]
-            ).rc:
-                errors.append("Error tar -xvzf the iml diagnostics file")
-        else:
-            errors = "Didn't recognize iml-diagnostics file format"
-
-        diagnostics_dir = re.compile("(sosreport-.*)\.tar\..*").search(diagnostics).group(1)
-
-        if shell_run(["chmod", "-R", "777", "%s/%s" % (self.destination_path, diagnostics_dir)]).rc:
+        if shell_run(["chmod", "777", "%s/%s" % (self.destination_path, diagnostics)]).rc:
```
johnsonw: With this being diagnostic info this should be fine, just wondering if something like 655 would work?

jgrund: it might, but I'd rather not change it right now.

PR #1543, opened 2020-01-30
Add page enum

### LINKED ISSUE #1539: Implement Page enum
Our current approach to having multiple pages in iml-gui is to store each model as a top-level struct. Instead, we should model the current page as an enum and only store the model of the currently active page.

This is beneficial because:

  - It gives a hook into page initialization, where we can show a loading indicator if a given subpage needs a network resource.
  - It allows for items to be destroyed easily on page navigation. For example, consider a subpage component that needs to poll a network resource. If we always persist the subpage model on the top-level model, we would need to explicitly check and stop the poll on any route change. This way, we can stop polling as soon as the page has changed, as it's model will no longer exist.

jgrund:
Fixes #1539.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1543)

### DIFF
--- a/iml-gui/crate/src/auth.rs
+++ b/iml-gui/crate/src/auth.rs
@@ -35,7 +35,6 @@ impl Default for State {
 
 #[allow(clippy::large_enum_variant)]
 #[derive(Clone)]
-#[allow(clippy::large_enum_variant)]
 pub enum Msg {
     Fetch,
     Fetched(fetch::ResponseDataResult<Session>),
--- a/iml-gui/crate/src/auth.rs
+++ b/iml-gui/crate/src/auth.rs
@@ -35,6 +35,7 @@ impl Default for State {
 
 #[allow(clippy::large_enum_variant)]
 #[derive(Clone)]
+#[allow(clippy::large_enum_variant)]
 pub enum Msg {
     Fetch,
     Fetched(fetch::ResponseDataResult<Session>),
--- a/iml-gui/crate/src/lib.rs
+++ b/iml-gui/crate/src/lib.rs
@@ -4,6 +4,7 @@
 
 pub mod components;
 pub mod key_codes;
+pub mod page;
 
 mod auth;
 mod breakpoints;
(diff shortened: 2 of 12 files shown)


ip1981:
It will require some manual work to keep the global and local models coherent, won't it?

jgrund:
> It will require some manual work to keep the global and local models coherent, won't it?

In what way?

ip1981:
> > It will require some manual work to keep the global and local models coherent, won't it?
> 
> In what way?

For example, when a server is deleted, you need to update the local model of the servers page. This is done [here](https://github.com/whamcloud/integrated-manager-for-lustre/blob/288138a583030f21745eb3b054a885fe33acd8de/iml-gui/crate/src/lib.rs#L438), here](https://github.com/whamcloud/integrated-manager-for-lustre/blob/288138a583030f21745eb3b054a885fe33acd8de/iml-gui/crate/src/lib.rs#L375).

Later, with more pages and models, it may become difficult to track.

jgrund:
> For example, when a server is deleted, you need to update the local model of the servers page. This is done here, here, here.

Sure, but that's an existing issue and orthogonal to this change. State flows from top-down so anything new needs to be propagated to the children.

jgrund:
One thing I've thought about a little bit is moving the shared state (cache / session) between the pages so they own it. That would mean some top-level stuff moves along with it (the wrapper view around most subpages), but should also mean you don't need to check which model is active in the top-level. 

I need to mull over a bit how that should work, I don't think it should be part of this change unless we can clearly define it.

iml-gui/crate/src/auth.rs
```diff
@@ -35,6 +35,7 @@ impl Default for State {
 
 #[allow(clippy::large_enum_variant)]
 #[derive(Clone)]
+#[allow(clippy::large_enum_variant)]
```
ip1981: There is the one above already.

jgrund: oops

jgrund:
```suggestion
```

PR #1549, opened 2020-01-31
Save docker images and install with RPM

jgrund:
We can automate most of the docker setup process (and eventually all of
it) for RHEL hosts by creating an RPM of the docker stack images.

1. Build / pull all relevant images locally
2. use `docker save` to build a tar archive with their contents
3. Create a RPM with image archive
4. Have RPM move iml-cli alias to the correct spot
5. Have RPM run `docker load` to load image archive
6. Create systemd unit to start / stop `docker stack` with `--resolve-image` flag

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1549)


jgrund:
Verified I can at least load the GUI by installing the RPM from this patch.

docker/Makefile
```diff
@@ -0,0 +1,19 @@
...
+	docker system prune -f --all --volumes
+	rm -rf iml-images.tar.gz
+
+build: clean
+	DOCKER_BUILDKIT=1 docker build -t imlteam/manager-service-base:5.1 -f base.dockerfile ../
+	DOCKER_BUILDKIT=1 docker build -t rust-iml-base -f rust-base.dockerfile ../
+	COMPOSE_DOCKER_CLI_BUILD=1 docker-compose build
+
+save: build
+	COMPOSE_DOCKER_CLI_BUILD=1 docker-compose pull postgres rabbit realtime view-server device-aggregator update-handler
```
johnsonw: I'll add in the grafana and influxdb images once I rebase on top of this. Should we also add:
```
iml-agent-comms
iml-warp-drive
iml-action-runner
iml-stratagem
iml-manager-cli
iml-ostpoool
iml-postoffice
```
?

docker/iml-action-runner.dockerfile
```diff
@@ -4,7 +4,8 @@ FROM ubuntu
 COPY --from=builder /build/target/release/iml-action-runner /usr/local/bin
 
 RUN apt-get update \
-  && apt-get install -y curl
+  && apt-get install -y curl \
```
johnsonw: There is some commonality between some of these images. I'll take care of moving the common pieces to a base image.

iml-docker.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=iml-docker
+Requires=docker.service
+After=docker.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+StandardError=null
```
johnsonw: We don't want to send output to journal?

iml-docker.service
```diff
@@ -0,0 +1,18 @@
...
+StandardOutput=null
+WorkingDirectory=/etc/iml-docker
+ExecStart=docker stack deploy -c /etc/iml-docker/docker-compose.yml -c /etc/iml-docker/docker-compose.overrides.yml iml --resolve-image=never
+# @FIXME: Need to poll service removal since docker stack rm is async
```
johnsonw: This should work:
```
ExecStop=docker stack rm iml && while [ "$(docker stack ps iml 2>&1 | grep 'nothing found in stack') == 1" ]; do echo "Waiting on docker stack to finish removing" && sleep 1; done;
```

.gitignore
```diff
@@ -1,49 +1,51 @@
...
 chroma_unit_test_env/
-tests/framework/utils/defaults.sh
-tests/framework/chroma_support.repo
-tests/framework/services/runner.sh
-tests/framework/integration/shared_storage_configuration/full_cluster/cluster_setup
+chroma-config.1.gz
+copr-local.mk
+database.db
+deps
+dist
```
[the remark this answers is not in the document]
jgrund: guess we should have sorted these earlier :). Updating.

docker/Makefile
```diff
@@ -0,0 +1,19 @@
+COMPOSE_IMAGES := $(shell docker-compose config 2>/dev/null | awk '{if ($$1 == "image:") print $$2;}')
+
+.PHONY: clean build save 
+
+clean:
+	docker system prune -f --all --volumes
+	rm -rf iml-images.tar.gz
+
+build: clean
+	DOCKER_BUILDKIT=1 docker build -t imlteam/manager-service-base:5.1 -f base.dockerfile ../
```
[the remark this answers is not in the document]
jgrund: the `build` task depends on `clean` which runs `docker system prune -f --all --volumes`. That prunes everything from the system. 

In terms of running `docker stack rm` I think that's probably orthogonal to this which is a build task.

In the save step (`docker save ...`) I explicitly pull in the parsed compose images to not include anything unrelated

docker/Makefile
jgrund: This is only to pull things that are built in other repos. All of the above is already created in the `build` task.

iml-docker.service
jgrund: I don't think we want to echo anything during service shutdown.

iml-docker.service
jgrund: I'm not sure yet. I need to play around with a bit to figure out the best thing to do.

iml-docker.service
jgrund: hrrm, we could also look at the exit code (which returns 1 when the stack is empty). Do we want to do that instead of grepping?

johnsonw: Yeah I think that should work.

jgrund: My concern is there could be other scenarios where `docker stack ps iml` could return 1. 

On the flip side if `docker stack ps iml` ever changes it's text when there are no services running, then this will loop forever.

jgrund: Ended up going with exit code.

iml-docker.service
jgrund: Changed to journal output.

PR #1551, opened 2020-02-02
Put cluster in maintence mode before changing port

### LINKED ISSUE #1548: Put cluster in maintenance mode when changing ports
It looks like we can update mcast port by doing the following:

1. Run `pcs property set maintenance-mode=true`
1. Edit /etc/corosync/corosync.conf, change mcastport to a new value, and bump config_version.
1. Run `pcs cluster sync`
1. Run `systemctl restart corosync` on all cluster nodes
1. Run `pcs property set maintenance-mode=false`

jgrund:
Fixes #1548.

This patch changes the behavior of updating the mcast port.

Previously, changing mcast port needed to be done one at a time for each node in a cluster. This meant nodes would flip between online / offline, and if there was a fencing-agent, they may have ended up STONITHing each other as mcast ports would be different. 

Now, changing the mcast port on one node will update it across the cluster. Nodes will no longer STONITH each other, as we put the cluster into and out of maintenance mode.

Here is what the new series of commands look like:

Screen Shot 2020-02-05 at 6 03 44 PM

And here is what the update job looks like:

Screen Shot 2020-02-05 at 6 04 58 PM

In addition to these changes, we now update the `config_version` in the `corosync.conf`, so the cluster can tell which version is newer.

This patch also moves a number of action plugins to rust.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1551)


chroma_core/models/corosync2.py
```diff
@@ -313,18 +314,47 @@ def get_steps(self):
...
+        )
 
-        self.invoke_agent_expect_result(
-            corosync_configuration.host,
-            "change_mcast_port",
-            {"old_mcast_port": kwargs["old_mcast_port"], "new_mcast_port": kwargs["new_mcast_port"]},
+
+class ActivateClusterMaintenceModeStep(Step):
+    def run(self, kwargs):
+        self.invoke_rust_agent_expect_result(kwargs["fqdn"], "pcs", ["property", "set", "maintenance-mode=true"])
```
utopiabound: More directly `crm_attribute --name maintenance-mode --update true`

chroma_core/models/corosync2.py
```diff
@@ -313,18 +314,47 @@ def get_steps(self):
...
+
+
+class DeactivateClusterMaintenceModeStep(Step):
+    def run(self, kwargs):
+        self.invoke_rust_agent_expect_result(kwargs["fqdn"], "pcs", ["property", "unset", "maintenance-mode"])
```
utopiabound: More directly `crm_attribute --name maintenance-mode --delete`

iml-agent/src/action_plugins/high_availability/corosync_conf.rs
```diff
@@ -0,0 +1,257 @@
...
+        if x.trim().starts_with("mcastport:") {
+            RE.replace(&x, |caps: &regex::Captures| {
+                format!("{}{}", &caps[1], new_port)
+            })
+            .to_string()
+        } else {
+            x
+        }
+    })
+}
```
utopiabound: This assumes mcastport is specified in config file

iml-agent/src/action_plugins/high_availability/corosync_conf.rs
```diff
@@ -0,0 +1,257 @@
...
+            mcastaddr: x.get("mcastaddr (str)").map(|x| (*x).to_string()),
+            broadcast: x.get("broadcast (str)").map(|x| (*x).to_string()),
+            mcastport: try_get(&x, "mcastport (u16)")?.parse()?,
+        })
+    }
+}
+
+/// View the current totem and interface information from `corosync-cmapctl`.
+async fn corosync_cmapctl() -> Result<(Totem, Vec<Interface>), ImlAgentError> {
+    let x = Command::new("corosync-cmapctl").checked_output().await?;
```
utopiabound: If we only care about the cmap sections that appear in the config: 
```corosync-cmapctl totem logging quorum nodelist qb resources```
(though I've never seen a deployment with the last two)

File: tests/integration/core/remote_operations.py [hunk 5]
```diff
@@ -937,14 +937,19 @@ def clear_ha(self, server_list):
...
 
-            # clear_ha_el7.sh
-            result = self._ssh_address(address, "if crm_mon -b1; then crm_resource -l|xargs pcs resource disable; fi")
+            result = self._ssh_address(address, "if crm_mon -b1; then pcs property set maintenance-mode=true; fi")
+            logger.debug("CMD OUTPUT:\n%s" % result.stdout)
+
+            result = self._ssh_address(address, "if crm_mon -b1; then crm_resource -l | xargs pcs resource disable; fi")
+            logger.debug("CMD OUTPUT:\n%s" % result.stdout)
+
+            result = self._ssh_address(address, "if crm_mon -b1; then pcs property unset maintenance-mode; fi")
```
utopiabound: `crm_attribute --name maintenance-mode --delete`

tests/integration/core/remote_operations.py
```diff
@@ -937,14 +937,19 @@ def clear_ha(self, server_list):
 
             firewall = RemoteFirewallControl.create(address, self._ssh_address_no_check)
 
-            # clear_ha_el7.sh
-            result = self._ssh_address(address, "if crm_mon -b1; then crm_resource -l|xargs pcs resource disable; fi")
+            result = self._ssh_address(address, "if crm_mon -b1; then pcs property set maintenance-mode=true; fi")
```
utopiabound: `crm_attribute --name maintenance-mode --delete`

tests/integration/core/remote_operations.py
```diff
@@ -937,14 +937,19 @@ def clear_ha(self, server_list):
...
+            logger.debug("CMD OUTPUT:\n%s" % result.stdout)
+
+            result = self._ssh_address(address, "if crm_mon -b1; then crm_resource -l | xargs pcs resource disable; fi")
```
utopiabound: `crm_resource -l | xargs -n 1 crm_resource --set-parameter target-role --meta --parameter-value Started --resource`

iml-agent/src/action_plugins/high_availability/corosync_conf.rs
jgrund: Where else would it be specified?

If the `mcastport` is missing from the file, the regex replace is never run and this fn is effectively a noop.

iml-agent/src/action_plugins/high_availability/corosync_conf.rs
jgrund: can just be `corosync-cmapctl totem` for now I think.

tests/integration/core/remote_operations.py
jgrund: hrrm, I think this should be `--parameter-value Disabled`?

jgrund: meh, `Stopped` works since we are in maintenance-mode.

iml-agent/src/action_plugins/high_availability/corosync_conf.rs
utopiabound: That's what I mean.  If it's missing in the file, it would show in `corosync-cmapctl totem.interface` output.  Defaults to port 5405.  This will fail silently.  Sort of an odd corner case, but if we're managing random corosync configs...

jgrund: I see... I'll fix in a followup

PR #1559, opened 2020-02-05
Update to newer Vagrantfile

jgrund:
Even though it's not available on vagrant cloud yet, we can pull the
newest CentOS 7 Vbox directly from cloud.centos.org.

This should allow for us to skip updating the kernel on each vm before
installing Lustre.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1559)


ip1981:
Maybe it is worth a checksum.

jgrund:
> Maybe it is worth a checksum.

Done

ip1981:
I wasn't even seeing http (without `S`) before the checksum was added :)

PR #1560, opened 2020-02-06
Synchronize file processing with incoming streams

jgrund:
If an incoming stream outpaces processing in the mailbox, it will close
before all lines have been flushed to disk.

If this occurs, we may end up trying to process things (ex report /
purge of fids) with an incomplete list.

This patch synchonizes writing all lines to disk such that a connection
will not close until the EOF has been processed and the data is flushed
to disk.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1560)

### DIFF
--- a/iml-mailbox/src/lib.rs
+++ b/iml-mailbox/src/lib.rs
@@ -3,7 +3,11 @@
 // license that can be found in the LICENSE file.
 
 use bytes::Buf;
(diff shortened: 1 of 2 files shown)


jgrund:
Been a while since I've timed it, but here are the latest timings for 100,000 fids:

report time:
real    0m23.958s
user    0m0.300s
sys     0m0.080s

purge time:
real    0m16.383s
user    0m0.234s
sys     0m0.042s

PR #1562, opened 2020-02-07
Bump docker tags

jgrund:
Bump docker tags to 5.1.1-dev, which is not in the registry.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1562)


docker/base.dockerfile
```diff
@@ -22,6 +22,7 @@ RUN yum update -y \
   && yum install -y python python-pip python-devel postgresql openssl gcc-c++ \
   && pip install -r requirements.txt \
   && yum autoremove -y gcc-c++ python-pip python-devel \
+  && rm -rf /root/.cache/pip \
```
ip1981: Nit: `/root/.cache` or even `/root/` might be good :)

jgrund: Not sure I'm comfortable removing everything under root. Removing the pip cache here saves about 22MB.

jgrund:
Confirmed this works with a docker stack deploy.

Images that are loaded and not in the registry:

Screen Shot 2020-02-07 at 9 07 15 AM

Stack running from the local images.

Screen Shot 2020-02-07 at 9 07 01 AM

johnsonw:
One thing you are probably missing is `iml-wasm-components` inside the nginx.dockerfile

jgrund:
> One thing you are probably missing is `iml-wasm-components` inside the nginx.dockerfile

I think it's fine to keep that as is. I don't see us furthering development there.

johnsonw:
Ok, everything else looks good. I'll +1

PR #1563, opened 2020-02-07
Create a module with common attributes

### LINKED ISSUE #1558: Create shared CSS classes module
While CSS properties should belong to respective components, sometimes they can be common among different components. For example, a "button" could be an `a` or a `button`; or some set of classes does not really belong to any component like defining parent's properties which affect the children.

The proposed module name - `components::classes`. Example class:

```rust
fn container() -> Attrs {
  class![C.relative, C.group]
}
```
This would be used in components which need a tooltip or a popup or both.

Here could be various aliases, for example, colours.

ip1981:
Closes #1558.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1563)

### DIFF
--- a/iml-gui/crate/src/components/attrs.rs
+++ b/iml-gui/crate/src/components/attrs.rs
@@ -1,0 +1,16 @@
+/**
+ A module for common attributes.
+
+ While CSS classes and other attributes should belong to respective components,
+ sometimes they can be common among different components. For example, a
+ "button" could be an `<a>` or a `<button>`.
+*/
+use crate::generated::css_classes::C;
+use seed::{prelude::*, *};
+
+/// Call this fn within the element wrapping the tooltip or popup (or both)
+/// It will add the needed styles so the tooltip and popup will render
+/// in the expected position and behaviour.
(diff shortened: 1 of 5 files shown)


ip1981:
I am thinking about lazy static too:

```rust
lazy_static! {
    pub(crate) static ref CONTAINER: Attrs = class![C.relative, C.group];
}
```

jgrund:
I'd consider using lazy static sparingly, in this case a fn seems just as good and more straightforward.

iml-gui/crate/src/components/attrs.rs
```diff
@@ -0,0 +1,16 @@
+/**
```
jgrund: Hrrm, does this show up as a module-level doc comment?

ip1981:
> Yes, it can be done as the global object, but I forsee the need to pass some additional parameters there, and then it will need to be converted to the function back.
> 
 Oh, no :) Everything here is supposed to be dumb simple.

PR #1565, opened 2020-02-07
Use Centos:7 for rust-base and systemd-base

jgrund:
Use Centos:7 for rust-base and systemd-base.

This will remove the Centos:8 image from being bundled and should
hopefully save a few hundred MBs in the final RPM.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1565)


jgrund:
Looks like this change brings us from 796MB -> 727MB.

PR #1570, opened 2020-02-11
Call rustup update after loading environment

mkpankov:
Previously it wouldn't update the toolchain - my setup still had 1.40 because of it.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1570)


mkpankov:
Ok it seems the true reason is that I hadn't merged the master yet where `rustup update` is added, but I think this change is still worth it.

PR #1574, opened 2020-02-12
Move /action api endpoint to rust.

jgrund:
This patch moves the /action api endpoint to rust.

Endpoint auth is provided via the nginx proxy.

This endpoint should be a bit faster than before because:

1. It's in rust
1. We perform multiple rpc calls in parallel using Futures.

- Remove python /action endpoint
- Abstract rabbitmq client filter to it's own feature under iml-rabbit.
- Add TryFrom implementation for `CompositeId`
- Add try_from deserializer for `CompositeId`
- Add constructor for `ApiList`

COPR Module Manager: managerforlustre/action-api

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1574)


iml-job-scheduler-rpc/src/lib.rs
```diff
@@ -126,5 +127,49 @@ pub async fn call<I: Debug + serde::Serialize, T: serde::de::DeserializeOwned>(
         return Err(e.into());
     }
 
-    Ok(resp.result)
+    if let Some(x) = resp.result {
+        Ok(x)
+    } else {
+        Err(ImlJobSchedulerRpcError::RpcError(
+            "RPC response was unexpectedly empty".into(),
+        ))
```
ip1981:
```rust
resp.result.ok_or(ImlJobSchedulerRpcError::RpcError(
            "RPC response was unexpectedly empty".into()
        ))
```
?

jgrund: probably should be `ok_or_else` so it doesn't allocate unless needed. I'll update.

jgrund:
Landed as: a79521660e8ce89b0413e7a871e7eecf1bddb5bb

PR #1176, opened 2019-08-23
Pull device-scanner directly into IML plugin architecture

jgrund:
Fixes #1175.

Signed-off-by: Joe Grund <jgrund@whamcloud.io>

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1176)

PR #1597, opened 2020-02-20
Add action dropdown to fs and fs detail pages.

jgrund:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1597)


iml-gui/crate/src/page/filesystem.rs
```diff
@@ -1,102 +1,183 @@
...
+                format!("Filesystem {}", &fs.label)
             ]
         ],
         T::wrapper_view(vec![
-            tr![T::th_right(plain!("Space Used / Total")), T::td_view(size_view(f))],
+            tr![T::th_left(plain!("Space Used / Total")), T::td_view(size_view(fs))],
             tr![
-                T::th_right(plain!("Files Created / Maximum")),
-                T::td_view(files_view(f))
+                T::th_left(plain!("Files Created / Maximum")),
```
ip1981: I aligned to the right, because I found it easier to read when the header and the value are closer to each other.

iml-gui/crate/src/page/filesystem.rs
```diff
@@ -203,27 +290,15 @@ fn server_link<I>(uri: Option<&String>, txt: &String) -> Node<I> {
...
-            " ",
-            size
-        ]
-    } else {
-        plain!("N/A")
-    }
+    a![
+        class![C.whitespace_no_wrap, C.text_blue_500, C.hover__underline, C.block],
+        attrs! {At::Href => Route::Volume(RouteId::from(vol_id)).to_href()},
+        t.volume_name,
```
ip1981: `volume_name` is something like "360014058cba15f6bfe7408090247fd69", not a device path?

iml-gui/crate/src/page/filesystems.rs
```diff
@@ -1,57 +1,130 @@
...
+                            fs_link(f),
+                            lock_indicator::view(all_locks, f).merge_attrs(class![C.mr_2]),
+                            alert_indicator(&cache.active_alert, f, true, Placement::Right)
+                        ])
+                        .merge_attrs(class![C.text_center]),
+                        T::td_view(filesystem::mgs(&cache.target.arc_values().cloned().collect(), f))
+                            .merge_attrs(class![C.text_center]),
+                        T::td_right(plain![f.mdts.len().to_string()]).merge_attrs(class![C.text_center]),
+                        T::td_right(filesystem::clients_view(f)),
+                        T::td_view(filesystem::size_view(f)).merge_attrs(class![C.text_center]),
```
ip1981: We could use `td_center`  :D

iml-gui/crate/src/page/filesystems.rs
```diff
@@ -1,57 +1,130 @@
...
+                    Some(row) => tr![
+                        T::td_view(vec![
+                            fs_link(f),
+                            lock_indicator::view(all_locks, f).merge_attrs(class![C.mr_2]),
+                            alert_indicator(&cache.active_alert, f, true, Placement::Right)
+                        ])
+                        .merge_attrs(class![C.text_center]),
+                        T::td_view(filesystem::mgs(&cache.target.arc_values().cloned().collect(), f))
+                            .merge_attrs(class![C.text_center]),
+                        T::td_right(plain![f.mdts.len().to_string()]).merge_attrs(class![C.text_center]),
```
ip1981: The content is a number, I think it is better be aligned to the right. On the other side, I can't imagine someone want to compare these number for different filesystems.

P. S. We definitely need `td_center` :)

jgrund:
___
*[iml-gui/crate/src/lib.rs, line 547 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1597#-M0XCAMv7piREhcmjzho:-M0XcJ0_CafsKBpoUIkn:b-ocswx4) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/b67b4c1baf119f093783304dd1f187310751b55e/iml-gui/crate/src/lib.rs#L547)):*
> We performing two consecutive clicks on one page and then immediately on another page. This looks strange.

Can you post a gif of what you're seeing?

iml-gui/crate/src/page/filesystem.rs
```diff
@@ -1,145 +1,237 @@
...
+pub struct Row {
+    dropdown: action_dropdown::Model,
+}
+
+#[derive(Default)]
+pub struct Model {
+    pub id: u32,
+    pub mdts: Vec<Target<TargetConfParam>>,
+    pub mgt: Vec<Target<TargetConfParam>>,
+    pub osts: Vec<Target<TargetConfParam>>,
```
ip1981: Why do we need a local copy?

iml-gui/crate/src/page/filesystem.rs
jgrund: I think we should be focusing on visual alignment between components rather than differing alignments dependant on type.

iml-gui/crate/src/page/filesystem.rs
jgrund: Yes, unfortunately there is a bug where a target can point to a no longer existent volume.

However, this situation should basically never occur, you can't have a target if the backing storage doesn't exist.

This will be fixed by #1176 which uses the devices scsi id (or other unique id for virtual devices) so that even if a device is removed and re-added in the db, the pointer will still be consistent.

iml-gui/crate/src/page/filesystems.rs
jgrund: yeah

iml-gui/crate/src/page/filesystem.rs
jgrund: The `rows` field needs to be updated whenever a target changes (add / remove). Because of this, Since we have to go through and update the rows for each table anyway, it makes sense to cache these values in the update.

We don't necessarily have to do this, but it also has the effect of being more performant. By doing it in the update, we split fields and sort only when a target has changed. If we did this in the view, the splitting and sorting would occur every render, regardless if anything relating to targets has changed.

jgrund:
___
*[iml-gui/crate/src/lib.rs, line 547 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1597#-M0XCAMv7piREhcmjzho:-M0wqjOFEeVzlHsn8DiB:bjgicmy) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/b67b4c1baf119f093783304dd1f187310751b55e/iml-gui/crate/src/lib.rs#L547)):*

Sorry, I've got troubles with vagrant, I am to rebuild it from clean state.

Anyways, if you are sure that the scenario is right, then I'm ok too.
</blockquote></details>

These messages proxy to the corresponding page message. Since only one page can be loaded at a time, all the other messages will effectively be a noop (and they all may be noops if none of these pages are currently active).

jgrund:
___
*[iml-gui/crate/src/lib.rs, line 698 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1597#-M0XE0em4AUGbRPX4mgT:-M0wsbuf0PYSyGToCngy:b-m4tx29) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/b67b4c1baf119f093783304dd1f187310751b55e/iml-gui/crate/src/lib.rs#L698)):*

> ```
>             model,
>             page::filesystems::view(&model.records, page, &model.locks)
>                 .els()
>                 .map_msg(Msg::FilesystemsPage),
>         )
>         .els(),
> ```
The same could be moved into `page::filesystems::view` using a function parameter, like you did for [modals](src/components/modal.rs:30):
```rust
pub fn content_view<T>(p_msg: fn(Msg) -> T, children: impl View<T>) -> Node<T> {...}
```

It seems, that using the function as the parameter the pattern matching case could be written then as
```rust
Page::Filesystems(page) => main_panels(model, page::filesystems::view(Msg::FilesystemsPage, page, &model.records, &model.locks).els(),
// or
Page::Filesystems(page) => main_panels(model, page::filesystems::view(Msg::FilesystemsPage, page, model).els(),
```
(although in the second variant it could be unwanted dependency on the parent model)
Have I forgotten something?

</blockquote></details>

Yes, It could be done that way, however, the `map_msg` call is preferred because it doesn't "leak" knowledge of the parent msg into the child component.

I passed the parent message into the modal because in that case we need to leak the parent msg as we don't know all possible variants of things people will want to do inside a modal.

iml-gui/crate/src/page/filesystem.rs
ip1981: That's true, but this table is in fact a vertical one in opposite to all others :)

iml-gui/crate/src/page/filesystem.rs
```diff
@@ -1,145 +1,237 @@
...
+        }
+        Msg::SetTargets(xs) => {
+            let old_keys: HashSet<u32> = model.rows.keys().copied().collect();
+            let new_keys: HashSet<u32> = xs.iter().map(|x| x.id).collect();
+
+            let to_remove = &old_keys - &new_keys;
+            let to_add = &new_keys - &old_keys;
+
+            for x in to_remove {
+                model.rows.remove(&x);
```
ip1981: It looks like we are passing all the targets (for all filesystems). Should we pre-filter them first like it is done below (`filter(|t| is_fs_target(model.id, t))`)?

PR #1600, opened 2020-02-20
Conditionally skip yum check

### LINKED ISSUE #1581: Support Deployments where all packages are pre-installed
There are appliance environments where all packages are pre-installed during the kickstart phase and there are no repos available during runtime.

This is incompatible with the current state of IML, as at a bare minimum it expects a repo available for local installations.

This ticket is to add a new code-path where IML can skip any yum install commands and instead enable / start units that have already been installed.

utopiabound:
If base packages (ServerProfile::base_packages) are installed and no yum repos are enabled
Assume everything is pre-installed

Fixes #1581

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1600)

### DIFF
--- a/chroma_core/models/host.py
+++ b/chroma_core/models/host.py
@@ -1021,9 +1021,10 @@ class SetHostProfileStep(Step):
         old_session_id = AgentRpc.get_session_id(host.fqdn)
         self.invoke_agent(host, "restart_agent")
         AgentRpc.await_restart(host.fqdn, timeout=settings.AGENT_RESTART_TIMEOUT, old_session_id=old_session_id)
-        self.invoke_rust_local_action_expect_result(
-            "await_next_session", (host.fqdn, old_rust_session_id, settings.AGENT_RESTART_TIMEOUT)
-        )
+        if old_rust_session_id:
+            self.invoke_rust_local_action_expect_result(
+                "await_next_session", (host.fqdn, old_rust_session_id, settings.AGENT_RESTART_TIMEOUT)
+            )
 
     @classmethod
     def describe(cls, kwargs):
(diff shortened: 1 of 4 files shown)


chroma_core/models/host_jobs/test_host_contact.py
```diff
@@ -108,9 +109,14 @@ def _test_yum_sanity(self, agent_ssh, auth_args, address):
             check_yum = """
 python -c "from yum import YumBase
 yb = YumBase()
+baselist = %s
+if len(yb.repos.listEnabled()) == 0 and len([x for x in yb.doPackageLists(pkgnarrow='installed', patterns=baselist)]) == len(baselist):
```
ip1981: Nit: they say `if not list:` will be more idiomatic Python than `if len(list) == 0`.
Nit2: if `baselist` contained wildcards, the lists might be different :)

chroma_core/models/host.py
```diff
@@ -1021,9 +1021,10 @@ def run(self, kwargs):
         old_session_id = AgentRpc.get_session_id(host.fqdn)
         self.invoke_agent(host, "restart_agent")
         AgentRpc.await_restart(host.fqdn, timeout=settings.AGENT_RESTART_TIMEOUT, old_session_id=old_session_id)
-        self.invoke_rust_local_action_expect_result(
-            "await_next_session", (host.fqdn, old_rust_session_id, settings.AGENT_RESTART_TIMEOUT)
-        )
+        if old_rust_session_id:
```
jgrund: I'm guessing you hit a case where this was `None`?

johnsonw: I'm not sure if this is related but i've noticed if I remove a server and then add it back it doesn't have the same session and will not allow me to add the server. Is this related to that?

jgrund:
> I'm not sure if this is related but i've noticed if I remove a server and then add it back it doesn't have the same session and will not allow me to add the server. Is this related to that?

@johnsonw can you provide more details on that? are there any specific errors you're seeing?

PR #1603, opened 2020-02-21
Package new GUI as RPM and docker image

jgrund:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1603)

### DIFF
--- a/chroma-manager.conf.template
+++ b/chroma-manager.conf.template
@@ -86,7 +86,7 @@ server {
     }
 
     location ~ ^/$ {
-        return 301 https://$http_host/ui2;
+        return 307 https://$http_host/ui2;
     }
 
     location /help {
--- a/chroma-manager.conf.template
+++ b/chroma-manager.conf.template
@@ -8,6 +8,12 @@ map $status $loggable {
     default 1;
 }
 
+map $uri $cache_control {
+    default no-cache;
+    ~\.html$ public,must-revalidate;
+    ~\.(jpg|jpeg|png|gif|ico|css|js|wasm|woff2|svg)$ public,max-age=31536000,immutable;
+}
+
 server {
     listen {{HTTPS_FRONTEND_PORT}} ssl http2;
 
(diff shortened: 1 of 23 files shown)


chroma-manager.conf.template
```diff
@@ -78,7 +86,7 @@ server {
     }
 
     location ~ ^/$ {
-        return 301 https://$http_host/ui;
+        return 301 https://$http_host/ui2;
```
utopiabound: Should this be 302 until it can be changed back to `/ui`?

docker/iml-gui.dockerfile
```diff
@@ -0,0 +1,18 @@
...
+    curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add - && \
+    echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list && \
+    curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
+    apt-get install -y yarn nodejs && \
+    rustup target add wasm32-unknown-unknown && \
+    cargo install wasm-pack && \
+    yarn install && \
+    yarn build:release;
+
+FROM ubuntu
```
johnsonw: Can this not be centos?

chroma-manager.conf.template
jgrund: Maybe 307?

jgrund:
```suggestion
        return 307 https://$http_host/ui2;
```

jgrund: Updated to 307

docker/iml-gui.dockerfile
jgrund: It doesn't matter, we don't use it in production. The files are copied out of this container and into the nginx one.

johnsonw: Ok, just wanted to make sure we don't introduce a new image if we can avoid it.

PR #1608, opened 2020-02-25
Refactor number-formatter to use integers

### LINKED ISSUE #1567: number-formatter should be more precise
Number formatter is used to format cardinal numbers like number of files or bytes in a human friendly way. For example, `2000 B` will become `1.95 KiB`. In case of smaller numbers the formatter produces unsatisfactory results due to its configurable precision feature: if precision is set to 2, the number `123` will be formatted as `123.00` which is confusing.

The suggestion is:

1. Switch to unsigned integers (`u64`)
2. Ignore precision (assume 0) when power is 0.

ip1981:
Make it print precise numbers if they are small enough.

number-formatter

Closes #1567.

### DIFF
--- a/iml-gui/crate/src/components/tree.rs
+++ b/iml-gui/crate/src/components/tree.rs
@@ -700,7 +700,7 @@ fn tree_volume_collection_view(cache: &ArcCache, model: &Model, parent_address: 
                     let v = cache.volume.values().find(|v| v.id == x.volume_id).unwrap();
 
                     let size = match v.size {
-                        Some(x) => format!(" ({})", number_formatter::format_bytes(x as f64, None)),
+                        Some(x) => format!(" ({})", number_formatter::format_bytes(x as u64, None)),
                         None => "".into(),
                     };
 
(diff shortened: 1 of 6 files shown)


ip1981:
1. The log disappeared because it shouldn't had sneaked in :)
2. AFAIK we are already limited to u64 in other places, and f64 could not be used to represent the exact number > max of u64, so the test didn't make much sense anyway (I think). We _might_ use u128, but to my knowledge u64 is _a lot_ and still enough.
3. As for the TODO test, it is [complicated](https://programming.guide/worlds-most-copied-so-snippet.html). We might just ignore that, or stop using from logarithms and float number altogether. I put the test here as as reminder.

Reviewable is down at the time of writing :)

ip1981:
For what I've seen in the API part, we are using Django's `IntegerField` which is i32. It becomes f64 because JS only supports f64.

jgrund:
> For what I've seen in the API part, we are using Django's IntegerField which is i32. It becomes f64 because JS only supports f64 (and max integer 2^53 - 1).

Keep in mind that within the next week we will be using influx to get these numbers.

jgrund:
> because JS only supports f64 (and max integer 2^53 - 1).

We are not restricted to JS here, but Rust instead since this ultimately is compiled to WebAssembly.

ip1981:
> Keep in mind that within the next week we will be using influx to get these numbers.

Still `u64` is two times more than enough (`i64`) :)

ip1981:
> > because JS only supports f64 (and max integer 2^53 - 1).
> 
> We are not restricted to JS here, but Rust instead since this ultimately is compiled to WebAssembly.

I was trying to say that so far  we were limited by JS and were not actually using large enough numbers, or somewhere there were incorrect appearance.

PR #1612, opened 2020-02-26
Update alert_indicator to take composite ids

jgrund:
It's easier to lookup alerts by composite ids than resource uris,
especially when fetching them from active alerts.

Update alert_indicator to take anything that implements
`ToCompositeId`, and update the /alert endpoint
to return the composite ids of effected items.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1612)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.8"
+version = "0.7.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "743ad5a418686aad3b87fd14c43badd828cf26e214a00f92a384291cf22e1811"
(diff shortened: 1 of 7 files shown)


jgrund:
> Could you remove ResourceUri if it's unused now?

Probably want it to stick around so it can be used to lookup an api resource when it's not on the resource object itself.

ip1981:
> > Could you remove ResourceUri if it's unused now?
> 
> Probably want it to stick around so it can be used to lookup an api resource when it's not on the resource object itself.

I added `ResourceUri` for the alert indicator specifically, so it might be orphan now. And then, what do you mean by "not on the resource object itself."? How could a trait be useful?

jgrund:
> I added ResourceUri for the alert indicator specifically, so it might be orphan now. And then, what do you mean by "not on the resource object itself."

Whoops, I was getting confused with `EndpointName` I guess. I'll remove.

PR #1614, opened 2020-02-26
Add Private key and Password auth types to server add

utopiabound:
Move all StructOpt versions to 0.3

Issue #1487 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1614)

### DIFF
--- a/iml-manager-cli/src/server.rs
+++ b/iml-manager-cli/src/server.rs
@@ -23,7 +23,10 @@ use std::{
     time::Duration,
 };
 use structopt::StructOpt;
-use tokio::time::delay_for;
+use tokio::{
+    io::{stdin, AsyncReadExt},
+    time::delay_for,
+};
 
 #[derive(StructOpt, Debug)]
 pub struct RemoveHosts {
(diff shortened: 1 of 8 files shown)


iml-manager-cli/src/server.rs
```diff
@@ -61,17 +73,61 @@ pub enum ServerCommand {
...
+    // For auth_type == "private_key_choice"
+    private_key: Option<String>,
+}
+
+impl<'a> TestHostConfig<'a> {
+    fn new(address: &'a str, auth: &AuthType) -> Self {
+        Self {
+            address,
+            auth_type: auth.into(),
+            root_password: {
```
johnsonw: Not going to hold up the PR but I think we can simplify this by doing the following:

```
impl From<&AuthType> for Option<String> {
    fn from(auth: &AuthType) -> Self {
        match auth {
            AuthType::Password(p) => Some(p.into()),
            AuthType::PrivateKey(p) => Some(p.into()),
            _ => None
        }
    }
}

impl<'a> TestHostConfig<'a> {
    fn new(address: &'a str, auth: &AuthType) -> Self {
        Self {
            address,
            auth_type: auth.into(),
            root_password: auth.into(),
            private_key: auth.into()
        }
    }
}
```
Let me know your thoughts.

jgrund: Seems reasonable, I think this change highlights how the API struct is not really designed correctly.

I.E. since root_password and private key are disjoint, it really doesn't make sense to have two fields there. 

They should probably be collapsed into a single field which could hold either one of them.

PR #1626, opened 2020-03-03
Switch the default port

mkpankov:
To what the defaults everywhere else are

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1626)

### DIFF
--- a/iml-gui/configs/webpack.config.js
+++ b/iml-gui/configs/webpack.config.js
@@ -37,7 +37,7 @@ module.exports = (env, argv) => {
       proxy: [
         {
           context: ["/api", "/help"],
-          target: "https://localhost:7444/",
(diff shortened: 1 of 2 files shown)


iml-gui/configs/webpack.config.js
```diff
@@ -37,7 +37,7 @@ module.exports = (env, argv) => {
       proxy: [
         {
           context: ["/api", "/help"],
-          target: "https://localhost:7444/",
+          target: "https://localhost:8443/",
```
jgrund: I know it's not part of this patch, but we could extract the port to a `const` and use it both here and as a prop to `HtmlWebpackPlugin` so we wouldn't have to update two places.

PR #1629, opened 2020-03-03
Return empty list if resource list is empty in get_ha_resource_list

utopiabound:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1629)

### DIFF
--- a/iml-agent/src/action_plugins/high_availability/mod.rs
+++ b/iml-agent/src/action_plugins/high_availability/mod.rs
@@ -68,7 +68,13 @@ fn process_resource(output: &[u8]) -> Result<ResourceAgentInfo, ImlAgentError> {
 }
 
 pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAgentError> {
-    let resources = crm_resource(&["--list-raw"]).await?;
+    let resources = match crm_resource(&["--list-raw"]).await {
+        Ok(res) => res,
+        Err(e) => {
+            tracing::info!("Failed to get resource list: {:?}", e);
+            return Ok(vec![]);
+        }
+    };
 
     let xs = resources
         .trim()


iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -68,7 +68,13 @@ fn process_resource(output: &[u8]) -> Result<ResourceAgentInfo, ImlAgentError> {
 }
 
 pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAgentError> {
-    let resources = crm_resource(&["--list-raw"]).await?;
+    let resources = match crm_resource(&["--list-raw"]).await {
+        Ok(res) => res,
+        Err(e) => {
+            tracing::info!("Failed to get resource list: {:?}", e);
+            return Ok(vec![]);
```
johnsonw: So before this was short circuiting because of the failure case which was causing the error right?

utopiabound: Yes.  This was returning an error because `crm_resource` was missing.

jgrund: Should we be more specific here and only handle the os error 2 case?

johnsonw: That's a good question. If there was another type of error, would we always want to return an empty vector?

utopiabound: Any errors we would WANT to pass up?  Or if you ask for list of ha devices and something goes wrong so we say "none we can find"

utopiabound: If we're more specific we'll need to figure out the error return for `crm_resource` installed, but not running

PR #1662, opened 2020-03-05
Calculate relative time using server time

### LINKED ISSUE #1633: Implement relative offset in GUI
When showing relative timestamps in the GUI ("happened x time ago") we need to account for time diff between the local (browser) and remote (server) nodes.

To do this, we can take the offset by looking at the current date on the local node and the `Date` header from the server node and subtracting the difference.

ip1981:
Closes #1633.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1662)

### DIFF
--- a/iml-gui/crate/src/server_date.rs
+++ b/iml-gui/crate/src/server_date.rs
@@ -1,0 +1,15 @@
+#[derive(Default)]
+pub struct ServerDate {
+    date: Option<chrono::DateTime<chrono::offset::FixedOffset>>,
+}
+
+impl ServerDate {
+    pub(crate) fn set(self: &mut Self, date: chrono::DateTime<chrono::offset::FixedOffset>) {
+        self.date = Some(date);
+    }
+    pub(crate) fn timeago(&self, date: chrono::DateTime<chrono::offset::FixedOffset>) -> String {
+        self.date
+            .map(|sd| format!("{}", chrono_humanize::HumanTime::from(date - sd)))
+            .unwrap_or_else(|| date.to_rfc2822())
+    }
+}
--- a/iml-gui/crate/src/auth.rs
+++ b/iml-gui/crate/src/auth.rs
@@ -37,7 +37,7 @@ impl Default for State {
 #[derive(Clone)]
 pub enum Msg {
     Fetch,
-    Fetched(fetch::ResponseDataResult<Session>),
+    Fetched(fetch::FetchObject<Session>),
     SetSession(Session),
     Loop,
     Stop,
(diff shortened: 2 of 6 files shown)


iml-gui/crate/src/server_date.rs
```diff
@@ -0,0 +1,15 @@
...
+    date: Option<chrono::DateTime<chrono::offset::FixedOffset>>,
+}
+
+impl ServerDate {
+    pub(crate) fn set(self: &mut Self, date: chrono::DateTime<chrono::offset::FixedOffset>) {
+        self.date = Some(date);
+    }
+    pub(crate) fn timeago(&self, date: chrono::DateTime<chrono::offset::FixedOffset>) -> String {
+        self.date
+            .map(|sd| format!("{}", chrono_humanize::HumanTime::from(date - sd)))
```
jgrund: Just curious, does this work the same regardless of which value is larger? I.E. if `date > sd` or if `sd > date`

ip1981: I think no, it can be "ago", "in" depending on which is large.

jgrund: Since it always will be in the past (since these events aren't happening in the future), should we always subtract the larger number from the smaller number?

ip1981: I first wrote `sd - date` but that displayed as "in 2 weeks" :)  Then I checks [the sources](https://docs.rs/chrono-humanize/0.0.11/src/chrono_humanize/humantime.rs.html#147) and it looks like it assumes a negative duration is related to a moment in the past. Which makes sense IMHO. Thus `date - sd` seems correct.

ip1981:
> Just curious, what FetchObject has that ResponseDataResult doesn't have?

`FetchObject` allows access to the whole response including headers.

ip1981:
> The only thing I'd like to see is a test on the parsing

Parsing is out of scope here on its own, It is done by the chrono crate and the input comes from the backend. So IMHO testing here will add no value or confidence, but #1663 will.

iml-gui/crate/src/server_date.rs
jgrund: > Thus date - sd seems correct.

But what happens if `sd` is greater than `date` (which is possible). In that case it would show the event as happening in the future, correct?

ip1981: That would be interesting :) Yes, it will be in the future, unless the difference in small (10 secs) - then it will be "now". Maybe we should log such cases and show some fallback.

jgrund: Hrrm, nm I think it will still be ok. 

Since we are calculating offsets, the result (even if positive) should still yield an event time in the past.

PR #1668, opened 2020-03-10
Set Retention Policy for Influx and downsample data

### LINKED ISSUE #1661: Influx: Downsample with continuous query
Create retention rules and continuous query to down sample data
Determine amount of space needed for retention

utopiabound:
Retain influx data for 90 days
Downsample iml_stats to 30m average, keep raw (per-second) data for 1 day.
This should keep total data size fairly small.

Fixes #1661 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1668)


chroma_core/lib/service_config.py
```diff
@@ -317,7 +317,43 @@ def _setup_influxdb(self):
         log.info("Creating InfluxDB database...")
         self.try_shell(["influx", "-execute", "CREATE DATABASE {}".format(settings.INFLUXDB_IML_DB)])
         self.try_shell(["influx", "-execute", "CREATE DATABASE {}".format(settings.INFLUXDB_STRATAGEM_SCAN_DB)])
+        self.try_shell(
```
ip1981: I wonder if this needs to be updated for docker too.

jgrund: Given we only store the most recent scan result for stratagem, I wonder if we need to have a retention policy for it.

chroma_core/lib/service_config.py
```diff
@@ -317,7 +317,43 @@ def _setup_influxdb(self):
...
+                'CREATE RETENTION POLICY "a_quarter" ON "{}" DURATION 90d REPLICATION 1 SHARD DURATION 5d'.format(
+                    settings.INFLUXDB_IML_STATS_DB
+                ),
+            ]
+        )
+        self.try_shell(
+            [
+                "influx",
+                "-execute",
+                'CREATE CONTINUOUS QUERY "downsample" ON "{}" BEGIN SELECT mean(*) INTO "{}"."a_quarter".:MEASUREMENT FROM /.*/ GROUP BY time(30m) END'.format(
```
jgrund: Is `mean` the correct aggregation function for every stat being stored?

jgrund:
Since we are now creating a new measurement `iml_stats.a_quarter`, we will need to dynamically switch  between the default measurement and this measurement when the date range in grafana is > 1 day.

Will this cause an issue in grafana, i.e. is there a way we can do this easily with our current charts?

chroma_core/lib/service_config.py
utopiabound: Still need to add mean of differences for several data sets

utopiabound:
screen

chroma_core/lib/service_config.py
johnsonw: What would be the downside of using the 90 day retention policy?

jgrund: Stats get dropped after 90 days. Though they would probably be stale by that point anyway.

PR #1687, opened 2020-03-15
Api config

johnsonw:
Add ability to include a config file in `/etc/iml-docker/setup/config`. This config should (at least for now) allow the installer to include: `BRANDING` and `STRATRAGEM_ENABLED`. These values will be loaded into environment variables along with the rest of the variables specified in `print-settings.py`. The gui app will additionally wait for these values to be loaded before the gui is loaded.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1687)

### DIFF
--- a/chroma_core/management/commands/print-settings.py
+++ b/chroma_core/management/commands/print-settings.py
@@ -1,10 +1,11 @@
-# Copyright (c) 2018 DDN. All rights reserved.
+# Copyright (c) 2020 DDN. All rights reserved.
 # Use of this source code is governed by a MIT-style
 # license that can be found in the LICENSE file.
 
 import settings
 import glob
 import json
+import os
 from django.db import connection
 from django.core.management.commands.runserver import Command as BaseCommand
 
@@ -16,6 +17,18 @@ class Command(BaseCommand):
 
     help = """Prints out selected settings in env variable format"""
 
+    def _load_config(self, config_file):
+        config = {}
+        if os.path.exists(config_file):
+            with open(config_file) as file:
+                content = file.read()
+
+                for line in content.splitlines():
+                    key_val = line.split("=", 1)
+                    config[key_val[0]] = key_val[1]
+
+        return config
+
     def handle(self, *args, **kwargs):
         cursor = connection.cursor()
         cursor.execute("SELECT * from api_key()")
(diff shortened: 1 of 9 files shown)


chroma_core/management/commands/print-settings.py
```diff
@@ -16,6 +17,21 @@ class Command(BaseCommand):
 
     help = """Prints out selected settings in env variable format"""
 
+    def _load_config(self, config_file):
```
ip1981: https://pypi.org/project/python-dotenv/ ?

iml-wire-types/src/lib.rs
```diff
@@ -1509,5 +1511,33 @@ impl PartialEq for OstPool {
...
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum Branding {
+    Whamcloud,
+    Ddn,
+    DdnAi400,
+}
+
+impl From<String> for Branding {
+    fn from(x: String) -> Self {
+        match x.as_str() {
```
ip1981: Maybe convert to lower case?

iml-wire-types/src/lib.rs
```diff
@@ -1509,5 +1511,33 @@ impl PartialEq for OstPool {
     }
 }
 
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub enum Branding {
+    Whamcloud,
+    Ddn,
+    DdnAi400,
```
ip1981: Is an enum really needed? Could it be a plain string? We can style it on display.

chroma_core/management/commands/print-settings.py
johnsonw: I'd rather not bring in any extra dependencies if we can avoid it.

iml-wire-types/src/lib.rs
johnsonw: sure, sounds good 👍

iml-wire-types/src/lib.rs
johnsonw: Personally I think the enum makes sense.

ip1981: I don't :) First of all you are limiting possible branding, adding new will be difficult, will require recompilation, etc. Then, you are basically reading a string, convert it to a enum and back to a string. Why not omit intermediate representation?

johnsonw: Because the type system checks the values. What if someone creates a config file and specifies:
"BRANDING=ddm"

Without the enum, it will simply read it as `ddm` and then they will spend time trying to figure out why it's loading the whamcloud logo. By using the enum, it will immediately fail giving an error message that it couldn't parse the branding, allowing you to fix the problem much faster. While it may seem like more boilerplate, I would much rather use the types. If a new logo is added in the future, it still has to be updated in the code anyway.

iml-gui/crate/src/lib.rs
```diff
@@ -863,7 +865,7 @@ fn view(model: &Model) -> Vec<Node<Msg>> {
         .els(),
         Page::Filesystem(page) => main_panels(
             model,
-            page::filesystem::view(&model.records, page, &model.locks, model.auth.get_session())
+            page::filesystem::view(&model.records, page, &model.locks, model.auth.get_session(), model.conf.use_stratagem)
```
ip1981: Maybe we should start passing the whole `model`.

johnsonw: The thought did cross my mind :) I think 4 parameters are probably ok. If it ends up going beyond that then I think it makes sense to just pass the model. For now I think this is ok though.

jgrund: In general I think it's better to pass in params then expose the parent model to a child component.

chroma_core/management/commands/print-settings.py
```diff
@@ -16,6 +17,21 @@ class Command(BaseCommand):
...
+        if os.path.exists(config_file):
+            file = open(config_file, "r")
```
jgrund: use the `with` context handler for opening files. ref: https://www.pythonforbeginners.com/files/with-statement-in-python

iml-wire-types/src/lib.rs
jgrund: Brandings are very much hardcoded right now. If we get to the point where we allow for any arbitrary branding (loading svg files from outside the binary instead of ones compiled in) then we should revisit this.

johnsonw:
Screenshots:

Whamcloud:
image

image

DDN AI 400:
image

image

DDN:

settings.py
```diff
@@ -96,6 +96,10 @@
 
 INCLUDES = ""
 
+BRANDING = "Whamcloud"
```
jgrund: nit, could be lowercase.

johnsonw:
Just a note on how to setup the config. 

For Docker:
In `/etc/iml-docker/setup/config`:
```
BRANDING=ddn
USE_STRATAGEM=True
```

For RPM Based setups:
In `/var/lib/chroma-setup/config`:
```
BRANDING=ddn
USE_STRATAGEM=True
```

Values can be changed in the config file but will need to be repopulated by the settings populator. This is straight forward in an RPM based approach as you can simply run `systemctl restart iml-settings-populator`. We are still looking at a good solution for docker, not only for this but also for profiles. For now, that should not hold up this patch.

johnsonw:
Just wanted to comment on possible values for `BRANDING`:
1. whamcloud
2. ddn
3. ddnai400

PR #1693, opened 2020-03-16
Remove volume links and other tree-view links that do not need to be shown

johnsonw:
- remove link from volumes on filesystem detail, mgt, individual volumes in tree view
- Remove OST pools parent link in tree view
- Remove Target parent links (except MGT)

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1693)


jgrund:
> Why don't they need to be shown?

They will be, but in a future version.

PR #1696, opened 2020-03-17
Update stats to utilize batch writes

johnsonw:
the stats service currently writes each point to influx serially. This results in one http request / response for each point to write it to influx.

Instead, we should do a single write by batching all points in one http request / response. ref: https://docs.influxdata.com/influxdb/v1.7/guides/writing_data/#writing-multiple-points

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1696)


iml-gui/crate/src/components/chart/fs_usage.rs
```diff
@@ -96,7 +96,7 @@ pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg, GMsg>)
     SELECT LAST("bytes_total") AS bytes_total,
     LAST("bytes_free") as bytes_free,
     LAST("bytes_avail") as bytes_avail
-    FROM "target" WHERE "kind" = '"OST"' {} GROUP BY target
+    FROM "target" WHERE "kind" = 'OST' {} GROUP BY target
```
ip1981: I think we need such an update in few other places.

PR #1697, opened 2020-03-18
Replace "N/A" with "---"

ip1981:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1697)

### DIFF
--- a/iml-gui/crate/src/components/stratagem/inode_table.rs
+++ b/iml-gui/crate/src/components/stratagem/inode_table.rs
@@ -137,7 +137,7 @@ pub(crate) fn view(model: &Model) -> Node<Msg> {
             p![
                 class![C.py_4, C.text_gray_600],
                 "Last Scanned: ",
-                model.last_known_scan.as_ref().unwrap_or(&"N/A".to_string())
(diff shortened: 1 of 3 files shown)


iml-gui/crate/src/components/stratagem/inode_table.rs
```diff
@@ -137,7 +137,7 @@ pub(crate) fn view(model: &Model) -> Node<Msg> {
             p![
                 class![C.py_4, C.text_gray_600],
                 "Last Scanned: ",
-                model.last_known_scan.as_ref().unwrap_or(&"N/A".to_string())
+                model.last_known_scan.as_ref().unwrap_or(&"---".to_string())
```
ip1981: Note a `unwrap_or()`, but it is not clear how to avoid it (and use `unwrap_or_else()`).

jgrund: You could do:

```rust
model.last_known_scan.as_deref().unwrap_or_else(|| "---")
```

The key is to deref the `String` from `last_known_scan` into a `&str`.

PR #1704, opened 2020-03-19
Update functionality of date picker on charts

johnsonw:
The charts should adjust their range based on the selected period.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1704)

### DIFF
--- a/iml-gui/crate/src/components/grafana_chart.rs
+++ b/iml-gui/crate/src/components/grafana_chart.rs
@@ -3,13 +3,10 @@
 // license that can be found in the LICENSE file.
 
 use seed::{attrs, iframe, prelude::*};
-use serde_with::with_prefix;
-use std::collections::HashMap;
+use std::collections::BTreeMap;
 
 pub static IML_METRICS_DASHBOARD_ID: &str = "8Klek6wZz";
 pub static IML_METRICS_DASHBOARD_NAME: &str = "iml-metrics";
-
-with_prefix!(grafana_var "var-");
 
 #[derive(serde::Serialize)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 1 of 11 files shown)


iml-gui/crate/src/components/grafana_chart.rs
```diff
@@ -3,29 +3,37 @@
...
     panel_id: u16,
     vars: impl IntoIterator<Item = (impl ToString, impl ToString)>,
 ) -> GrafanaChartData<'a> {
-    let hm: HashMap<String, String> = vars.into_iter().map(|(x, y)| (x.to_string(), y.to_string())).collect();
+    let hm: BTreeMap<String, String> = vars
+        .into_iter()
+        .map(|(x, y)| {
+            let var = x.to_string();
+            let key = match var.as_str() {
+                "to" | "from" => var.to_string(),
```
ip1981: It looks like an abuse of vars. They all are supposed to be prefixed with "var-". Why not add to and from to other parameters like panel id?

johnsonw: I'm not sure why grafana made the decision to not use the `var-` prefix like everything else when using`to` and `from`, but because they are part of the query parameters I think they belong here.

johnsonw: Just to follow up, all query parameters are being passed in and then serialized. If we were to take these two variables out we would have to work them into the query string outside of the serialized string. Furthermore, we would need to pass the variables individually to each function that would use it (`create_chart_params`, `performance_container` and any other function that would use it in the future). Grouping all query parameters together makes more sense to me:
```
create_chart_params(
   32,
    vec![
      ("fs_name", &model.fs_name),
      ("from", &model.mdt_usage_date_picker.from),
      ("to", &model.mdt_usage_date_picker.to)
   ]
),
```

johnsonw:
date_picker

iml-gui/crate/src/components/grafana_chart.rs
ip1981: I mean vars and non-vars are different entities. And since we have only one way to pass them all in a query, it is decided to distinguish them by a var- prefix.

ip1981: Also any exceptions are bad, to and from are now exceptions, one will not be able to add "real" vars (to or from) easily. And correct me: any graph can be parameterized by to and from parameter, but each graph may have zero or more specific vars like fs_name.

jgrund: All charts can be parameterized by from and to, however it won't make sense for all charts (for example OST balance, where we always take latest).

In the interest of time, I'll land this and then extract the from and to params in a follow-on patch.

ip1981: Yeah, the "var-" variable are "template" variables, totally different thing from the data set variables :)

PR #1706, opened 2020-03-19
Fix grafana charts

johnsonw:
The grafana charts are not updating. This is because the `var-` prefix
that was being automatically injected by serde was removed (to allow for
    the `from` and `to` parameters). This patch updates the stratagem
charts to use the `create_chart_params` function instead of passing a
struct.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1706)

### DIFF
--- a/iml-gui/crate/src/components/stratagem/mod.rs
+++ b/iml-gui/crate/src/components/stratagem/mod.rs
@@ -394,22 +394,12 @@ pub(crate) fn view(model: &Model, all_locks: &Locks) -> Node<Msg> {
                 caption_wrapper(
                     "inode Usage Distribution",
                     Some(&last_scan),
-                    stratagem_chart(grafana_chart::GrafanaChartData {
-                        org_id: 1,
-                        refresh: "1m",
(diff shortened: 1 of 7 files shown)


iml-gui/crate/src/components/stratagem/mod.rs
```diff
@@ -394,22 +394,12 @@ pub(crate) fn view(model: &Model, all_locks: &Locks) -> Node<Msg> {
                 caption_wrapper(
                     "inode Usage Distribution",
                     Some(&last_scan),
-                    stratagem_chart(grafana_chart::GrafanaChartData {
-                        org_id: 1,
-                        refresh: "1m",
-                        panel_id: 2,
-                        vars: config.grafana_vars.clone()
-                    })
+                    stratagem_chart(grafana_chart::create_chart_params(2, "1m", config.grafana_vars.clone()))
```
ip1981: What is odd here now is the question why vars do not include `panel_id` and `refresh` :)

PR #1707, opened 2020-03-20
Show page details in title and in breadcrumbs

### LINKED ISSUE #1577: Could show specific server/filesystem/etc. in breadcrumbs
Currently breadcrumbs looks like `/ Filesystems / Filesystem Details`, it might be better as `/ Filesystem / qwerty` with the name of the filesystem currently viewed. The same for servers and other similar entities.

ip1981:
Derive page title from the page's model rather
than from the route. Closes #1577.

title

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1707)

### DIFF
--- a/iml-gui/crate/src/page/mod.rs
+++ b/iml-gui/crate/src/page/mod.rs
@@ -72,6 +72,36 @@ pub(crate) enum Page {
     Volume(volume::Model),
 }
 
+impl Page {
+    pub(crate) fn title(self: &Self) -> String {
+        match self {
+            Self::About => "About".into(),
+            Self::AppLoading => "Loading...".into(),
+            Self::Filesystems(_) => "Filesystems".into(),
+            Self::Filesystem(m) => format!("Filesystem: {}", &m.fs.name),
+            Self::Dashboard(_) => "Dashboard".into(),
+            Self::FsDashboard(m) => format!("{} Filesystem Dashboard", &m.fs_name),
+            Self::ServerDashboard(m) => format!("{} Server Dashboard", &m.host_name),
+            Self::TargetDashboard(m) => format!("{} Target Dashboard", &m.target_name),
+            Self::Jobstats => "Jobstats".into(),
+            Self::Login(_) => "Login".into(),
+            Self::Mgts(_) => "MGTs".into(),
+            Self::NotFound => "Page not found".into(),
+            Self::OstPools => "OST Pools".into(),
+            Self::OstPool(m) => format!("OST Pool: {}", &m.id),
+            Self::PowerControl => "Power Control".into(),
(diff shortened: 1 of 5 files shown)


iml-gui/crate/src/page/mod.rs
```diff
@@ -72,6 +72,36 @@ pub(crate) enum Page {
...
+            Self::Dashboard(_) => "Dashboard".into(),
+            Self::FsDashboard(m) => format!("{} Filesystem Dashboard", &m.fs_name),
+            Self::ServerDashboard(m) => format!("{} Server Dashboard", &m.host_name),
+            Self::TargetDashboard(m) => format!("{} Target Dashboard", &m.target_name),
+            Self::Jobstats => "Jobstats".into(),
+            Self::Login(_) => "Login".into(),
+            Self::Mgts(_) => "MGTs".into(),
+            Self::NotFound => "Page not found".into(),
+            Self::OstPools => "OST Pools".into(),
+            Self::OstPool(m) => format!("OST Pool: {}", &m.id),
```
jgrund:
```suggestion
            Self::OstPool(m) => format!("OST Pool: {}", &m.name),
```

iml-gui/crate/src/page/mod.rs
```diff
@@ -72,6 +72,36 @@ pub(crate) enum Page {
...
+            Self::PowerControl => "Power Control".into(),
+            Self::Servers(_) => "Servers".into(),
+            Self::Server(m) => format!("Server: {}", &m.server.fqdn),
+            Self::Targets => "Targets".into(),
+            Self::Target(m) => format!("Target: {}", &m.target.name),
+            Self::Users => "Users".into(),
+            Self::User(m) => format!("User: {}", &m.user.username),
+            Self::Volumes(_) => "Volumes".into(),
+            Self::Volume(m) => format!("Volume: {}", &m.id),
```
jgrund:
```suggestion
            Self::Volume(m) => format!("Volume: {}", &m.label),
```

iml-gui/crate/src/page/mod.rs
ip1981: It is not ready yet :) Like other places with ids.

jgrund: Still wouldn't hurt to add it now. We'll need to remember to update this when we implement the page otherwise.

ip1981: Well, I have just a rough idea that those models will contain respective entities. I could add them, but then add code to initialize them, etc. And all it just to fix this small issue instead of working on those features for real. On the other side, I am sure the first thing to do will be removal the dummy ids from those model which will break this code remind to update it.

PR #1709, opened 2020-03-20
Address immediate GUI visual bugs

### LINKED ISSUE #1708: Address immediate GUI visual bugs
There are a couple of gui bugs that need immediate fixing:
1. [x] - The right panel does not display properly in reponsive mode. For now, we should remove the right panel unless we are in a large view.
2. [x] - The volume table cell is rendering as a single column of text when the window is scaled down. We should fix the volume cell and also remove both the primary and failover server columns unless the browser is open in a medium view or larger.

johnsonw:
Fixes #1708.

Description:
There are a couple of gui bugs that need immediate fixing:

 - The right panel does not display properly in reponsive mode. For now, we should remove the right panel unless we are in a large view.
 - The volume table cell is rendering as a single column of text when the window is scaled down. We should fix the volume cell and also remove both the primary and failover server columns unless the browser is open in a medium view or larger.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1709)

### DIFF
--- a/iml-gui/crate/src/page/filesystem.rs
+++ b/iml-gui/crate/src/page/filesystem.rs
@@ -412,7 +412,7 @@ fn targets(
             h3![class![C.py_4, C.font_normal, C.text_lg], title]
         ],
         table![
-            class![C.table_fixed, C.w_full],
+            class![C.table_auto, C.w_full],
             style! {
                 St::BorderSpacing => px(10),
                 St::BorderCollapse => "initial"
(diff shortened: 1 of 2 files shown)


iml-gui/crate/src/page/filesystem.rs
```diff
@@ -421,8 +421,8 @@ fn targets(
                 t::thead_view(vec![
                     t::th_left(plain!["Name"]).merge_attrs(class![C.w_32]),
                     t::th_left(plain!["Volume"]),
-                    t::th_left(plain!["Primary Server"]).merge_attrs(class![C.w_48]),
-                    t::th_left(plain!["Failover Server"]).merge_attrs(class![C.w_48]),
+                    t::th_left(plain!["Primary Server"]).merge_attrs(class![C.w_48, C.hidden, C.md__table_cell]),
+                    t::th_left(plain!["Failover Server"]).merge_attrs(class![C.w_48, C.hidden, C.md__table_cell]),
                     t::th_left(plain!["Started on"]).merge_attrs(class![C.w_48]),
```
ip1981: I was thinking how we could put all these servers in one cell. E. g. show only "started on" and mark it somehow whether it's the primary or a failover. Maybe show all servers in a popup.

johnsonw: Yeah it's a good thought. Once things slow down perhaps we could look through the mocks a bit and make changes where necessary.

PR #1711, opened 2020-03-20
Add baseless profile

### LINKED ISSUE #1710: Initial server deployment should not add repos if server profile does not need them
IML deploys servers in two phases.

1. Add the agent and start it, deploying a "default" profile
2. deploy the server profile on top

The default profile will always have a base repo. However, there are cases where we don't want to deploy any repo as packages are pre-installed.

To address this, we should add another starting variant with no base repo and look at the server profile. If the profile we plan to deploy in step 2 has no repos, then the profile we use should also have no repos

jgrund:
Fixes #1710

Add a new default (phase 1) profile called default_baseless.

It has no base repos and is used iff the server_profile (phase 2)
has no base repos as well.

In addition add a list endpoint for server profiles.

  - Add `default_baseless.profile`
  - Update bootstrap to not create repo or install agent if generated
repofile is empty
  - Ensure iml cli returns correct error codes when commands fail.
  - Ensure iml cli returns error when there are no hosts to add.
  - Add some more debug to warp-drive.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1711)

### DIFF
--- a/iml-manager-cli/src/error.rs
+++ b/iml-manager-cli/src/error.rs
@@ -80,7 +80,7 @@ impl std::fmt::Display for ImlManagerCliError {
             ImlManagerCliError::DoesNotExist(ref err) => write!(f, "{} does not exist", err),
             ImlManagerCliError::FailedCommandError(ref xs) => {
                 let failed_msg = xs.iter().fold(
-                    String::from("The Following commands have failed:\n"),
+                    String::from("The following commands have failed:\n"),
                     |acc, x| format!("{}{}\n", acc, x.message),
                 );
 
--- a/iml-manager-cli/src/error.rs
+++ b/iml-manager-cli/src/error.rs
@@ -2,6 +2,8 @@
 // Use of this source code is governed by a MIT-style
 // license that can be found in the LICENSE file.
 
+use iml_wire_types::Command;
+
 #[derive(Debug)]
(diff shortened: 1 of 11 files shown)


iml-manager-cli/src/error.rs
```diff
@@ -54,35 +56,44 @@ impl std::error::Error for DurationParseError {}
...
         match *self {
+            ImlManagerCliError::ApiError(ref err) => write!(f, "{}", err),
             ImlManagerCliError::ClientRequestError(ref err) => write!(f, "{}", err),
-            ImlManagerCliError::TokioTimerError(ref err) => write!(f, "{}", err),
-            ImlManagerCliError::TokioJoinError(ref err) => write!(f, "{}", err),
+            ImlManagerCliError::CombineEasyError(ref err) => write!(f, "{}", err),
+            ImlManagerCliError::DoesNotExist(ref err) => write!(f, "{} does not exist", err),
+            ImlManagerCliError::FailedCommandError(ref xs) => {
+                let failed_msg = xs.iter().fold(
+                    String::from("The Following commands have failed:\n"),
```
jgrund:
```suggestion
                    String::from("The following commands have failed:\n"),
```

PR #1712, opened 2020-03-20
Update system integration test framework

johnsonw:
- Update the test framework to support a docker setup

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1712)


iml-system-tests/tests/common/vagrant.rs
```diff
@@ -190,6 +162,174 @@ pub async fn setup_deploy_servers(
...
+
+    Ok(x.success())
+}
+
+async fn restart_ntpd(hosts: &[&str]) -> Result<(), io::Error> {
+    vagrant()
+        .await?
+        .arg("ssh")
+        .arg("-c")
+        .arg("'sudo systemctl restart ntpd.service'")
```
ip1981: Are single quotes really needed?

johnsonw: I can try removing them. Not sure to be honest. In general, they are used after `-c`

ip1981: I thought the quotes were only needed when constructing a shell command. But this looks more like `execve()`.

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,126 @@
...
+    Ok(())
+}
+
+pub async fn volume_prune() -> Result<(), io::Error> {
+    let mut x = docker().await?;
+
+    x.arg("volume")
+        .arg("prune")
+        .arg("--force")
+        .checked_status()
```
ip1981: Will it prune all docker volumes, even the ones unrelated to IML?

iml-system-test-utils/src/iml.rs
```diff
@@ -0,0 +1,51 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::{try_command_n_times, CheckedStatus};
+use std::io;
+use tokio::{fs, process::Command};
+
+const IML_DOCKER_PATH: &str = "/etc/iml-docker";
```
ip1981: Just FYI: duplicates https://github.com/whamcloud/integrated-manager-for-lustre/pull/1712/files#diff-58bc3ec70ec0819d14c433040c8e99e7R10

iml-system-test-utils/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    let mut r = cmd.status().await?;
+
+    // try to run the command max_tries times until it succeeds. There is a delay of 1 second.
+    while !r.success() && count < max_tries {
+        count += 1;
+
+        thread::sleep(one_sec);
+
+        r = cmd.status().await?;
```
ip1981: It could be a "do..while" loop, so you don't need to repeat `cmd.status().await?;`. Ref. https://github.com/rust-lang/rfcs/issues/1313#issuecomment-146665147

iml-system-test-utils/src/docker.rs
johnsonw: Yes

iml-system-test-utils/src/lib.rs
johnsonw: I think we would still need to call `cmd.status().await?` on each iteration right? If it succeeds on the first call it shouldn't enter the while loop.

iml-system-test-utils/src/iml.rs
johnsonw: You're right, i'll make it public so it's not duplicated.

iml-system-tests/tests/common/vagrant.rs
johnsonw: I'm going to remove them and run a test to see if it works.

johnsonw: I added this to the Vagrantfile so we can configure ntp using a provisioning script so it's no longer an issue :)

iml-system-test-utils/src/lib.rs
ip1981: Yes, but you have the same call before entering the loop, no?

johnsonw: I think we can do it this way but i'm not sure how I feel about it. Something feels a little off to me and it might be that I have to declare the mutable variable outside of the loop regardless. This is what I came up with:

```
let mut r: ExitStatus;
while {
    r = cmd.status().await?;
    count += 1;

    thread::sleep(one_sec);

    !r.success() && count < max_tries
} {};

if r.success() {
    Ok(())
} else {
    Err(io::Error::new(io::ErrorKind::Other, format!("Command {:?} failed to succeed after {} attempts.", cmd, max_tries)))
}
```
I'm not seeing much of a gain honestly and it feels a bit hacky to me. Do you still think it's better to do it this way?

ip1981: It's ok. Existing code looks better :)

johnsonw: Ok cool, i'll leave it as is then.

iml-system-docker-tests/tests/docker_integration_test.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use std::collections::{HashMap, hash_map::RandomState};
+use iml_system_test_utils::{docker, iml, vagrant, get_local_server_names};
+use iml_systemd;
+use iml_wire_types;
```
jgrund:
```suggestion
```

You don't need to explicitly import these here, things will work the same without them.

iml-system-docker-tests/tests/docker_integration_test.rs
```diff
@@ -0,0 +1,89 @@
...
+    vagrant::create_fs(fs_type, &config.storage_servers()[..]).await?;
+
+    iml::detect_fs().await?;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_docker_ldiskfs_setup() -> Result<(), Box<dyn std::error::Error>> {
+    let server_names = get_local_server_names(&config.storage_servers());
```
jgrund: where is config supposed to come from?

iml-system-docker-tests/tests/docker_integration_test.rs
```diff
@@ -0,0 +1,89 @@
...
+        vec![("base_monitored".into(), &server_names[..])].into_iter().collect::<HashMap<String, &[String], RandomState>>(),
+        vagrant::FsType::LDISKFS
+    ).await?;
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_docker_zfs_setup() -> Result<(), Box<dyn std::error::Error>> {
+    let server_names = get_local_server_names(&config.storage_servers());
```
jgrund: where is config supposed to come from?

iml-system-docker-tests/tests/docker_integration_test.rs
```diff
@@ -0,0 +1,89 @@
...
+async fn test_docker_stratagem_setup() -> Result<(), Box<dyn std::error::Error>> {
+    run_fs_test(
+        &docker::DockerSetup {
+            use_stratagem: true,
+            branding: iml_wire_types::Branding::DdnAi400
+        },
+        vec![
+            ("stratagem_server".into(), &get_local_server_names(&config.get_mds_servers())[..]),
+            ("base_monitored".into(), &get_local_server_names(&config.get_oss_servers())[..]),
+            ("stratagem_client".into(), &get_local_server_names(&config.get_client_servers())[..]),
```
jgrund: where is config supposed to come from?

johnsonw:
What kind of compiler error are you seeing? It's compiling and running the test without any issues on my onyx node.

jgrund:
> What kind of compiler error are you seeing? It's compiling and running the test without any issues on my onyx node. 

I was looking at the revision before the one that was just pushed. It's fixed now.

johnsonw:
Oh interesting. I'm pretty sure that one was running without errors as well. I'll stop force-pushing at this point.

iml-system-docker-tests/tests/docker_integration_test.rs
johnsonw: removed

iml-system-docker-tests/tests/docker_integration_test.rs
jgrund: Fixed in most recent push.

iml-system-docker-tests/tests/docker_integration_test.rs
johnsonw: Ah now I understand the compiler error you were seeing.

iml-systemd/src/lib.rs
```diff
@@ -103,7 +107,7 @@ async fn get_unit_states(unit_name: &str) -> Result<(UnitFileState, ActiveState)
 /// # Arguments
 ///
 /// * `unit_name` - The unit to start
-pub async fn start_unit(unit_name: String) -> Result<(), ImlAgentError> {
+pub async fn start_unit_with_time(unit_name: String, time: u32) -> Result<(), SystemdError> {
```
jgrund:
```suggestion
pub async fn start_unit_and_wait(unit_name: String, time: u32) -> Result<(), SystemdError> {
```

Would better reflect intent I think.

johnsonw: sure 👍

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    let path = canonicalize(IML_DOCKER_PATH).await?;
+
+    x.current_dir(path);
+
+    Ok(x)
+}
+
+pub async fn deploy_iml_stack() -> Result<(), io::Error> {
+    iml_systemd::start_unit_and_wait("iml-docker.service".into(), 400).await?;
```
jgrund: 400 seconds?

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+
+    Ok(())
```
jgrund:
```suggestion
    iml_systemd::start_unit_and_wait("iml-docker.service".into(), 400).await
```

This is already returning the type needed, no need for `?` operator.

iml-system-test-utils/src/docker.rs
johnsonw: I experimented with this a bit and 400 is probably excessive but sometimes it can take a while for everything to load up and all of the docker services to be ready.

iml-system-test-utils/src/docker.rs
johnsonw: Ah right, good call

iml-systemd/src/systemd_error.rs
```diff
@@ -0,0 +1,95 @@
...
+            SystemdError::Io(ref err) => Some(err),
+            SystemdError::Utf8Error(ref err) => Some(err),
+            SystemdError::RequiredError(ref err) => Some(err),
+            SystemdError::CmdError(ref err) => Some(err),
+            SystemdError::UnexpectedStatusError => None,
+        }
+    }
+}
+
+impl From<SystemdError> for std::io::Error {
```
jgrund: This conversion seems odd, I don't think we need this.

johnsonw: Let me see what happens if I remove it.

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+}
```
jgrund: I got a bit ahead of myself, it should be something like this:

```suggestion
pub async fn deploy_iml_stack() -> Result<(), iml_systemd::SystemdError> {
    iml_systemd::start_unit_with_time("iml-docker.service".into(), 400).await
}
```

The `io::Error` conversion is not needed.

iml-systemd/src/lib.rs
```diff
@@ -2,7 +2,9 @@
 // Use of this source code is governed by a MIT-style
 // license that can be found in the LICENSE file.
 
-use crate::agent_error::{ImlAgentError, RequiredError};
+mod systemd_error;
+
+use crate::systemd_error::{RequiredError, SystemdError};
```
jgrund:
```suggestion
use crate::systemd_error::RequiredError;
pub use crate::systemd_error::SystemdError;
```

We should re-export `SystemdError` so it can be easily imported in other crates.

iml-systemd/src/systemd_error.rs
johnsonw: Seems like it's ok

iml-system-test-utils/src/docker.rs
johnsonw: 👍

iml-systemd/src/lib.rs
johnsonw: Makes sense 👍

iml-systemd/src/systemd_error.rs
johnsonw: The issue I'm running into at the moment is here:
```
rror[E0277]: `?` couldn't convert the error to `std::io::Error`
  --> src/docker.rs:42:62
   |
42 |     iml_systemd::stop_unit("iml-docker.service".into()).await?;
   |                                                              ^ the trait `std::convert::From<iml_systemd::systemd_error::SystemdError>` is not implemented for `std::io::Error`
```

jgrund: yeah, `remove_iml_stack` (the code where that's being called) should be like `pub async fn remove_iml_stack() -> Result<(), iml_systemd::SystemdError> {`

There is no need for it to return `io::Error`.

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+
+  let services = x.arg("service").arg("ls").output().await?;
+
+  let output = str::from_utf8(&services.stdout).expect("Couldn't read docker service list.");
+  let cnt: u16 = output.lines().count() as u16 - 1; // subtract the header
+
+  Ok(cnt)
+} 
+
+pub async fn remove_iml_stack() -> Result<(), io::Error> {
```
jgrund:
```suggestion
pub async fn remove_iml_stack() -> Result<(), iml_systemd::SystemdError> {
```

This will compile, and while not technically the best, it's still better than downsampling to io::Error from SystemdError.

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+pub async fn get_docker_service_count() -> Result<u16, io::Error> {
+  let mut x = docker().await?;
+
+  let services = x.arg("service").arg("ls").output().await?;
+
+  let output = str::from_utf8(&services.stdout).expect("Couldn't read docker service list.");
+  let cnt: u16 = output.lines().count() as u16 - 1; // subtract the header
+
+  Ok(cnt)
+}
```
jgrund:
```suggestion
pub async fn get_docker_service_count() -> Result<usize, io::Error> {
  let mut x = docker().await?;

  let services = x.arg("service").arg("ls").output().await?;

  let output = str::from_utf8(&services.stdout).expect("Couldn't read docker service list.");
  let cnt = output.lines().skip(1).count(); // subtract the header

  Ok(cnt)
}
```

casting to u16 is not needed, using `skip` means the result never tries to be negative.

iml-systemd/src/systemd_error.rs
johnsonw: I was thinking I could just map the error, something like this:
```
iml_systemd::stop_unit("iml-docker.service".into()).await
      .map_err(|x| io::Error::new(io::ErrorKind::Other, x.to_string()))?;
```
I thought we would still need the error type to be io::Error since we have a couple of commands that follow. For example:
```
x.arg("stack").arg("rm").arg("iml").checked_status().await?;
```
Those will use io::Error won't they?

jgrund: > Those will use io::Error won't they? 

Yes, but `SystemdError` wraps io::Error, so `?` will cast appropriately.

johnsonw: nvm,  looks like it compiles just fine.

johnsonw: ah that makes sense

iml-system-test-utils/src/docker.rs
jgrund: Why do we need this fn?

The systemd unit should not yield back until there are no more services: 

https://github.com/whamcloud/integrated-manager-for-lustre/blob/fc8a24bf662790622dd13ead7c2279dd88fcd257/iml-docker.service#L16

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+    let mut x = docker().await?;
+
+    x.arg("system")
+        .arg("prune")
+        .arg("--force")
+        .arg("--all")
+        .checked_status()
+        .await?;
+
+    Ok(())
```
jgrund:
```suggestion
    let mut x = docker().await?;

    x.arg("system")
        .arg("prune")
        .arg("--force")
        .arg("--all")
        .checked_status()
        .await
```

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+    let mut x = docker().await?;
+
+    x.arg("volume")
+        .arg("prune")
+        .arg("--force")
+        .checked_status()
+        .await?;
+
+    Ok(())
+}
```
jgrund:
```suggestion
pub async fn volume_prune() -> Result<(), io::Error> {
    let mut x = docker().await?;

    x.arg("volume")
        .arg("prune")
        .arg("--force")
        .checked_status()
        .await
}
```

iml-system-test-utils/src/docker.rs
johnsonw: Awesome, I was looking for `head` to skip the first line. Good to know `skip` exists.

iml-system-test-utils/src/docker.rs
```diff
@@ -0,0 +1,223 @@
...
+
+pub async fn iml_stack_loaded() -> Result<bool, io::Error> {
+    let x = iml::list_servers().await?.status().await?;
+
+    Ok(x.success())
+}
+
+pub async fn wait_for_iml_stack() -> Result<(), io::Error> {
```
jgrund: Why do we need this fn? the systemd service should wait until services are loaded and the api is accessible:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/fc8a24bf662790622dd13ead7c2279dd88fcd257/iml-docker.service#L14

iml-system-test-utils/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+    let one_sec = time::Duration::from_millis(1000);
+
+    let mut r = cmd.status().await?;
+
+    // try to run the command max_tries times until it succeeds. There is a delay of 1 second.
+    while !r.success() && count < max_tries {
+        println!("Trying command: {:?} - Attempt #{}", cmd, count + 1);
+        count += 1;
+
+        thread::sleep(one_sec);
```
jgrund: We should not be mixing threads and futures together. Sleeping can be done with tokio instead.

iml-system-test-utils/src/docker.rs
johnsonw: Right, and in most cases that works fine. There are cases, however, where running `systemctl stop iml-docker.service` immediately exits and docker services are still running. I think this happens if there is an error with one or more of the docker services. When this happens, you have to manually remove the stack and wait for all of the services to come down.

iml-system-test-utils/src/iml.rs
```diff
@@ -0,0 +1,52 @@
...
+        let mut x = iml().await?;
+        let mut cmd = x
+            .arg("server")
+            .arg("add")
+            .arg("-h")
+            .arg(hosts.join(","))
+            .arg("-p")
+            .arg(profile);
+
+        try_command_n_times(3, &mut cmd).await?;
```
jgrund: What issue is this trying to solve?

iml-system-test-utils/src/docker.rs
jgrund: I haven't seen this occur. We're probably better off failing the test if this happens so we can diagnose the error, rather than working around it in test.

iml-system-test-utils/src/docker.rs
johnsonw: Yeah I don't think we need this function. I'll remove.

iml-system-test-utils/src/docker.rs
johnsonw: In that case i'll take out the extra code and it will become:
```
pub async fn remove_iml_stack() -> Result<(), iml_systemd::SystemdError> {
    iml_systemd::stop_unit("iml-docker.service".into()).await
}
```

iml-system-test-utils/src/lib.rs
johnsonw: sure, I updated to tokio's `delay_for`:
```
 while !r.success() && count < max_tries {
        println!("Trying command: {:?} - Attempt #{}", cmd, count + 1);
        count += 1;

        delay_for(Duration::from_secs(1)).await;

        r = cmd.status().await?;
    }
```

iml-system-test-utils/src/iml.rs
johnsonw: Sometimes running a command fails if it's done immediately following another command. This happens many times when attempting to add a server, for example. The steps are:
1. Configure the docker network on each host
2. Add the host to iml

When moving from step 1 to step 2 immediately it seems the servers aren't quite ready yet and validation will fail. You could argue in this case that maybe we should use a delay instead and i'm not opposed to trying that, but I find this to be more successful.  It always succeeds on the second attempt. Another example is when destroying vagrant nodes. Sometimes when you destroy all the vagrant nodes it will fail. Re-running a second time usually succeeds.

jgrund: Given this is around deploying hosts, we probably want to know when something goes wrong instead of trying again. I think we should probably wait for n seconds after setting the network instead of retrying this on failure.

johnsonw: Sure thing. I'll find all the spots where i'm using this and instead wait three seconds before running the command.

iml-system-docker-tests/tests/docker_integration_test.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    docker::deploy_iml_stack().await?;
+
+    vagrant::setup_deploy_docker_servers(&config, server_map).await?;
+    vagrant::configure_ntp_for_host_only_if(&config).await?;
+    vagrant::create_fs(fs_type, &config.storage_servers()[..]).await?;
+
+    // Wait three seconds before detecting the filesystem
+    let three_secs = time::Duration::from_millis(3000);
+    thread::sleep(three_secs);
```
jgrund: We should not be mixing threads and futures, use tokio instead.

johnsonw: updated to use `delay_for`

docker/docker-compose.yml
```diff
@@ -320,6 +320,10 @@ services:
   setup:
     image: "imlteam/manager-setup:5.1.1-dev"
     command: ["tail", "-f", "/dev/null"]
+    restart_policy:
+      condition: on-failure
+      delay: 5s
+      window: 5s
```
jgrund:
```suggestion
```

This needs to be under the deploy section.

docker/docker-compose.yml
```diff
@@ -320,6 +320,10 @@ services:
   setup:
     image: "imlteam/manager-setup:5.1.1-dev"
     command: ["tail", "-f", "/dev/null"]
```
ip1981: I wonder if any  command is needed. Even if it is, could be "`true`" ?

iml-system-docker-tests/tests/docker_integration_test.rs
```diff
@@ -0,0 +1,130 @@
...
+use tokio::time::delay_for;
+
+async fn setup() -> Result<(), Box<dyn std::error::Error>> {
+    // remove the stack if it is running and clean up volumes and network
+    docker::remove_iml_stack().await?;
+    docker::system_prune().await?;
+    docker::volume_prune().await?;
+    docker::configure_docker_overrides().await?;
+    docker::stop_swarm().await?;
+    iml_systemd::restart_unit("docker.service".into()).await?;
```
jgrund, line 19: Why restart here?

iml-system-docker-tests/tests/docker_integration_test.rs
```diff
@@ -0,0 +1,130 @@
...
+) -> Result<(), Box<dyn std::error::Error>> {
+    setup().await?;
+    docker::configure_docker_setup(&docker_setup).await?;
+
+    docker::deploy_iml_stack().await?;
+
+    vagrant::setup_deploy_docker_servers(&config, server_map).await?;
+    vagrant::create_fs(fs_type, &config.storage_servers()[..]).await?;
+
+    // Wait three seconds before detecting the filesystem
```
jgrund, line 52:
```suggestion
    // Wait ten seconds before detecting the filesystem
```

jgrund: What happens if we don't wait here?

iml-system-test-utils/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+    } else {
+        let err = io::Error::new(
+            io::ErrorKind::Other,
+            format!("process exited with code: {:?}", x.code()),
+        );
+        Err(err)
+    }
+}
+
+pub trait CheckedStatus {
```
jgrund, line 21: This exists in the iml-cmd crate. You should be able to reuse that.

iml-system-docker-tests/tests/docker_integration_test.rs
ip1981, line 52:
```suggestion
    // Wait few seconds before detecting the filesystem
```

:)

ip1981: Or just delete the comment.

iml-system-test-utils/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+        self.status()
+            .and_then(|x| async move { handle_status(x) })
+            .boxed()
+    }
+}
+
+pub async fn try_command_n_times(max_tries: u32, cmd: &mut Command) -> Result<(), io::Error> {
+    let mut count = 0;
+    let mut r = cmd.status().await?;
```
jgrund, line 38: Because of this call, this fn actually tries the command n + 1 times. 

`count` should start at `1`, which will account for this.

docker/docker-compose.yml
johnsonw: We might be able to change it but I'd rather do it in a separate patch if that's ok.

iml-system-docker-tests/tests/docker_integration_test.rs
johnsonw: After cleaning up the swarm, volumes, etc, I figured this would be a good place to restart docker in case there are any network hiccups like we've seen in the past. It's not entirely necessary so I can remove it if you think it's not necessary, but it just ensures the docker service is clean before we start.

iml-system-docker-tests/tests/docker_integration_test.rs
johnsonw: There's a chance it will hang detecting the filesystem (doesn't happen every time).

iml-system-test-utils/src/lib.rs
johnsonw: sure 👍

vagrant/scripts/configure_ntp.sh
```diff
@@ -0,0 +1,20 @@
...
+# Append ntp server address 
+sudo sed -i -e "$ a server $NTP_SERVER iburst" /etc/ntp.conf
+sudo systemctl restart ntpd.service
+sudo systemctl enable ntpd.service
+
+sleep 30
+
+until ntpstat;
+do
+  systemctl restart ntpd.service
```
jgrund, line 18: Instead of restarting ntpd in a loop, we should do something like:

`ntpdate -qu 10.73.10.1`

vagrant/scripts/wait_for_ntp.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+until ntpstat;
+do
+  systemctl restart ntpd.service
```
jgrund, line 5: Instead of restarting ntpd in a loop, we should do something like:

`ntpdate -qu 10.73.10.1`

vagrant/scripts/configure_ntp.sh
johnsonw: this command would be run inside the loop?

iml-system-test-utils/src/lib.rs
```diff
@@ -0,0 +1,144 @@
...
+    servers
+        .iter()
+        .map(move |x| format!("{}.local", x))
+        .collect()
+}
+
+pub fn server_map_to_server_set<'a, S: std::hash::BuildHasher>(
+    server_map: &'a HashMap<String, &[&str], S>,
+) -> Vec<&'a str> {
+    let mut set = server_map
```
jgrund, line 133: It seems like a [`BTreeSet`](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html) is what you want here.

It is ordered and a set (deduped).

iml-system-test-utils/src/pdsh.rs
```diff
@@ -0,0 +1,125 @@
...
+pub async fn sed(base_path: &str) -> Result<Command, CmdError> {
+    let mut x = Command::new("sed");
+
+    let path = canonicalize(base_path).await?;
+    x.current_dir(path);
+
+    Ok(x)
+}
+
+pub async fn pdsh(hosts: &[&str], cmd: &str) -> Result<Command, CmdError> {
```
jgrund, line 20: Consider using `dshbak` or the output will be jumbled.

johnsonw: Updated.

iml-system-test-utils/src/lib.rs
jgrund: In that case you could just convert the `BTreeSet` to a `Vec` and return that.

Cargo.toml
```diff
@@ -34,5 +35,7 @@ members = [
 
 exclude = [
     'iml-gui/crate',
-    'iml-system-tests',
+    'iml-system-test-utils',
```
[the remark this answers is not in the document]
jgrund: Specifically, the `iml-system-test-utils` crate can be included into general compilation, as it doesn't run any tests on it's own.

iml-system-test-utils/src/pdsh.rs
```diff
@@ -0,0 +1,145 @@
...
+
+use iml_cmd::{CheckedCommandExt, CmdError};
+use std::{process::Stdio, str};
+use tokio::{
+    fs::canonicalize,
+    io::{AsyncReadExt, AsyncWriteExt},
+    process::Command,
+};
+
+// sed "s/\$1/2.12.4/" scripts/install_ldiskfs_no_iml.sh | pdsh -R ssh -u root -w 10.73.10.11,10.73.10.12
```
jgrund, line 13: cruft?

iml-systemd/src/systemd_error.rs
```diff
@@ -78,3 +78,12 @@ impl From<CmdError> for SystemdError {
         SystemdError::CmdError(err)
     }
 }
+
+impl serde::Serialize for SystemdError {
```
jgrund, line 82: Can we remove this?

iml-system-test-utils/src/pdsh.rs
johnsonw: I'll remove

iml-systemd/src/systemd_error.rs
johnsonw: removed

iml-wire-types/src/lib.rs
```diff
@@ -1394,10 +1394,10 @@ pub enum ActiveState {
 pub enum RunState {
```
jgrund, line 1394: This enum implements `Ord`, Activating needs to be where it was previously.

johnsonw: I think I had it in my patch and it was added to another patch which was merged into master so when I rebased it probably moved it to the bottom where I originally put it. I'll update.

PR #1713, opened 2020-03-23
Add route to show per-host volumes

### LINKED ISSUE #1688: Add route to show per-host volumes
Clicking the volumes folder under a host should show just the volumes for that host.

Screen Shot 2020-03-16 at 10 52 08 AM

We should probably create a new route `/host_volumes/RouteID` that will show the subset. We should be able to reuse the existing volumes page code.

ip1981:
The route is `/servers/<server-id>/volumes`.
It opens a page listing the volumes on the specified server.

Closes #1688.

server-volumes

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1713)

### DIFF
--- a/iml-gui/crate/src/components/tree.rs
+++ b/iml-gui/crate/src/components/tree.rs
@@ -5,6 +5,7 @@
 use crate::{
     components::{alert_indicator, font_awesome, paging, Placement},
     generated::css_classes::C,
+    route::RouteId,
     GMsg, Route,
 };
 use iml_wire_types::{
@@ -578,7 +579,7 @@ fn tree_host_item_view(cache: &ArcCache, model: &Model, host: &Host) -> Option<N
         ],
         alert_indicator(&cache.active_alert, &host, true, Placement::Bottom),
         if tree_node.open {
-            tree_volume_collection_view(cache, model, &address)
+            tree_volume_collection_view(cache, model, &address, &host)
         } else {
             empty![]
         }
@@ -714,11 +715,17 @@ fn tree_pools_collection_view(cache: &ArcCache, model: &Model, parent_address: &
     .unwrap_or(empty![])
 }
 
-fn tree_volume_collection_view(cache: &ArcCache, model: &Model, parent_address: &Address) -> Node<Msg> {
+fn tree_volume_collection_view(cache: &ArcCache, model: &Model, parent_address: &Address, host: &Host) -> Node<Msg> {
     tree_collection_view(
         model,
         parent_address.extend(Step::VolumeCollection),
-        |x| item_view("folder", &format!("Volumes ({})", x.paging.total()), Route::Volumes),
+        |x| {
+            item_view(
+                "folder",
+                &format!("Volumes ({})", x.paging.total()),
+                Route::ServerVolumes(RouteId::from(&host.id)),
+            )
+        },
         |x| {
             slice_page(&x.paging, &x.items)
                 .filter_map(|x| cache.volume_node.get(x))
--- a/iml-gui/crate/src/components/tree.rs
+++ b/iml-gui/crate/src/components/tree.rs
@@ -579,7 +579,7 @@ fn tree_host_item_view(cache: &ArcCache, model: &Model, host: &Host) -> Option<N
         ],
         alert_indicator(&cache.active_alert, &host, true, Placement::Bottom),
         if tree_node.open {
-            tree_volume_collection_view(cache, model, &address, &host)
+            tree_volume_collection_view(cache, model, &address, host)
         } else {
             empty![]
         }
--- a/iml-gui/crate/src/page/mod.rs
+++ b/iml-gui/crate/src/page/mod.rs
@@ -69,11 +69,13 @@ pub(crate) enum Page {
     Users,
     User(user::Model),
     Volumes(volumes::Model),
+    ServerVolumes(volumes::Model),
     Volume(volume::Model),
 }
 
 impl Page {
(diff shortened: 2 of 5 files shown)


iml-gui/crate/src/page/volumes.rs
```diff
@@ -65,15 +66,24 @@ impl RecordChange<Msg> for Model {
...
+        self.rows = build_rows(cache, &self.host);
 
         orders.proxy(Msg::Page).send_msg(paging::Msg::SetTotal(self.rows.len()));
 
         orders.send_msg(Msg::Sort);
     }
 }
 
-fn build_rows(cache: &ArcCache) -> Vec<Row> {
+impl From<&Arc<Host>> for Model {
```
ip1981: Just an experiment.

iml-gui/crate/src/page/mod.rs
```diff
@@ -96,7 +97,13 @@ impl Page {
             Self::Target(m) => format!("Target: {}", &m.target.name),
             Self::Users => "Users".into(),
             Self::User(m) => format!("User: {}", &m.user.username),
-            Self::Volumes(_) => "Volumes".into(),
+            Self::Volumes(m) => {
```
jgrund: It's probably worth making a new page variant rather than switching over `Page::Volumes` and checking the internal model. 

I.E. creating `Page::ServerVolumes`, which parallels `Route::ServerVolumes`.

iml-gui/crate/src/page/volumes.rs
```diff
@@ -65,15 +66,24 @@ impl RecordChange<Msg> for Model {
...
+    fn from(host: &Arc<Host>) -> Self {
+        Self {
+            host: Some(host.clone()),
```
jgrund:
```suggestion
            host: Some(Arc::clone(host)),
```

It's worth explicitly doing the clone this way to show we are not actually cloning the Host, but instead incrementing a refcount

iml-gui/crate/src/page/mod.rs
ip1981: I was thinking about `Self::User(m) => m.title()` (in the comment above). This would let make all fields private.

jgrund: Independent of making something like a `Title` trait that page models implement, I still think we should not reuse pages for different routes.

ip1981: I am 50/50 on this. I agree that this is a kind of different page, but on the other side it is the same page with the same model and the code. Technically, we reuse other pages (like Server) for different URLs (server id). And there is another extreme example - the NotFound page :)

jgrund: Where does server page get reused?

ip1981: I mean we use it with different models with different server id :) Just like other pages: user, volume, filesystem.

iml-gui/crate/src/page/volumes.rs
```diff
@@ -34,6 +34,7 @@ type Row = (Arc<VolumeRecord>, Vec<Arc<VolumeNodeRecord>>, Vec<Arc<Host>>);
 
 #[derive(Default)]
 pub struct Model {
+    pub(crate) host: Option<Arc<Host>>,
```
ip1981: This could be a list of hosts, but I can't imagine a use case yet.

jgrund: I don't see this changing from a single host.

iml-gui/crate/src/page/mod.rs
jgrund: Sort of; What I mean is it's largely been a 1:1 mapping between the `Route` enum and a `Page` enum. 

We are reusing the `Page::Volume` variant for both the `Route::Volumes` and `Route::ServerVolumes` page where I think a `Page::ServerVolume` variant would make more sense as it keeps the mapping.

I understand that the page module / update / view is being reused between the volumes route and the server volumes route, and I'm not against having a `Title` trait either so the page models own that data, but I think it makes for sense for there to be separate page variants.

ip1981:
> Everything looks good overall, just create a separate variant for ServerVolume page.

Done!

iml-gui/crate/src/page/mod.rs
```diff
@@ -69,11 +69,13 @@ pub(crate) enum Page {
     Users,
     User(user::Model),
     Volumes(volumes::Model),
+    ServerVolumes(volumes::Model),
     Volume(volume::Model),
 }
 
 impl Page {
     pub(crate) fn title(self: &Self) -> String {
+        // FIXME: delegate to a model where possible, e. g. Self::User(m) => m.title()?
```
jgrund: Nit: this should be an issue instead of fixme.

iml-gui/crate/src/components/tree.rs
```diff
@@ -578,7 +579,7 @@ fn tree_host_item_view(cache: &ArcCache, model: &Model, host: &Host) -> Option<N
         ],
         alert_indicator(&cache.active_alert, &host, true, Placement::Bottom),
         if tree_node.open {
-            tree_volume_collection_view(cache, model, &address)
+            tree_volume_collection_view(cache, model, &address, &host)
```
jgrund:
```suggestion
            tree_volume_collection_view(cache, model, &address, host)
```

ip1981: Oh, I missed this one. Thanks.

PR #1727, opened 2020-03-25
Fix warp drive hanging

### LINKED ISSUE #1726: Warp Drive - Cache queries appear to be hanging
**Describe the bug**
When starting a docker setup the warp drive service appears to hang after building the initial db statements:
https://github.com/whamcloud/integrated-manager-for-lustre/blob/d18d7d6d59f19f849d3ea3334886630cc13bec9b/iml-warp-drive/src/cache.rs#L340-L386

**To Reproduce**
1. Start the iml-docker service
2. Notice the warp-drive service hangs. This becomes apparent when you attempt to load the gui and the /messaging endpoint returns a 502.

**Expected behavior**
Warp drive should start without issue

**Notes:**
There is a possibility that this could be caused by the number of futures that are being joined together and run in parallel. As a first step, it makes sense to run these serially instead of in parallel and see if the service continues to hang.

johnsonw:
Fixes #1726.

Update the cache to use serial queries instead of being run in parallel.
This will hopefully prevent the service from hanging.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1727)

### DIFF
--- a/iml-warp-drive/src/cache.rs
+++ b/iml-warp-drive/src/cache.rs
@@ -339,33 +339,24 @@ pub async fn populate_from_db(
 
     tracing::debug!("Built initial db statements");
 
-    let fut1 = future::try_join5(
-        into_row(client.query_raw(&stmts[0], iter::empty()).await?),
-        into_row(client.query_raw(&stmts[1], iter::empty()).await?),
-        into_row(client.query_raw(&stmts[2], iter::empty()).await?),
-        into_row(client.query_raw(&stmts[3], iter::empty()).await?),
-        into_row(client.query_raw(&stmts[4], iter::empty()).await?),
-    );
+    let managed_target_mount = into_row(client.query_raw(&stmts[0], iter::empty()).await?).await?;
+    let stratagem_configuration =
+        into_row(client.query_raw(&stmts[1], iter::empty()).await?).await?;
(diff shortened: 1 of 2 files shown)


iml-warp-drive/src/cache.rs
```diff
@@ -339,33 +339,24 @@ pub async fn populate_from_db(
...
-        into_row(client.query_raw(&stmts[2], iter::empty()).await?),
-        into_row(client.query_raw(&stmts[3], iter::empty()).await?),
-        into_row(client.query_raw(&stmts[4], iter::empty()).await?),
-    );
+    let managed_target_mount = into_row(client.query_raw(&stmts[0], iter::empty()).await?).await?;
+    let stratagem_configuration =
+        into_row(client.query_raw(&stmts[1], iter::empty()).await?).await?;
+    let lnet_configuration = into_row(client.query_raw(&stmts[2], iter::empty()).await?).await?;
+    let volume = into_row(client.query_raw(&stmts[3], iter::empty()).await?).await?;
+    let volume_node = into_row(client.query_raw(&stmts[4], iter::empty()).await?).await?;
```
jgrund: I was thinking we could keep the `fut{1,2,3}` as they were and just run them serially to keep some parallelism instead of running everything serially. But since this only runs on startup it should be ok.

Issue #1734, opened 2020-03-26
GUI: Breadcrumb user should reflect first + last name if present


jgrund:
When the first and last name are entered in the user field, The logged in user will update in the top right.

The current user on the breadcrumb trail should also update.

![Screen Shot 2020-03-26 at 10 45 36 AM](https://user-images.githubusercontent.com/458717/77662145-e0a27a00-6f51-11ea-8638-6ad7441cda62.png)

ip1981:
Updating breadcrumbs might be nontrivial to implement because we only store strings. Plus first+last name can be too long for breadcrumbs. User id (username) seems the most appropriate for being immutable, no?

ip1981:
I think it makes sense to update the current breadcrumb with the current page's title, iff the latter changes (and make it change with the user's names?)

ip1981:
By the way, we should compare breadcrumbs by URL only (that's easy), automatically we will be able to update them without any special code.

PR #1739, opened 2020-03-27
Update page title & breadcrumb with user

ip1981:
In general: update the current breadcrumb whenever the current page's title
changes.

In particular: when the user is updated, update the user page's title

Closes: #1734.

user

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1739)

### DIFF
--- a/iml-gui/crate/src/lib.rs
+++ b/iml-gui/crate/src/lib.rs
@@ -28,7 +28,7 @@ mod watch_state;
 mod test_utils;
 
 use components::{
-    breadcrumbs::BreadCrumbs, command_modal, font_awesome, font_awesome_outline, loading, restrict, stratagem, tree,
+    breadcrumbs, command_modal, font_awesome, font_awesome_outline, loading, restrict, stratagem, tree,
     update_activity_health, ActivityHealth,
 };
 pub(crate) use extensions::*;
@@ -124,6 +124,26 @@ impl Loading {
     }
 }
 
+struct BreadCrumb {
+    href: String,
+    title: String,
+}
+
+impl breadcrumbs::BreadCrumb for BreadCrumb {
+    fn href(&self) -> &str {
+        &self.href
+    }
+    fn title(&self) -> &str {
+        &self.title
+    }
+}
+
+impl PartialEq for BreadCrumb {
+    fn eq(&self, other: &Self) -> bool {
+        self.href == other.href
+    }
+}
+
 // ------ ------
 //     Model
 // ------ ------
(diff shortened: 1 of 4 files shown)


iml-gui/crate/src/page/user.rs
```diff
@@ -81,6 +83,14 @@ pub struct Model {
 }
 
 impl Model {
+    pub fn title(&self) -> String {
+        match (self.user.first_name.as_str(), self.user.last_name.as_str()) {
+            ("", "") => self.user.username.clone(),
+            ("", l) => l.into(),
+            (f, "") => f.into(),
+            (f, l) => format!("{} {}", f, l),
```
nlinker: Nice!

ip1981: I am just standing on the shoulders of giants ;)

iml-gui/crate/src/lib.rs
```diff
@@ -319,8 +343,12 @@ pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg, GMsg>)
             orders.send_msg(Msg::LoadPage);
         }
         Msg::UpdatePageTitle => {
-            let title = format!("{} - {}", model.page.title(), TITLE_SUFFIX);
-            document().set_title(&title);
+            let title = model.page.title();
+            document().set_title(&format!("{} - {}", &title, TITLE_SUFFIX));
+            model.breadcrumbs.push(BreadCrumb {
```
johnsonw: I wonder if we should send a message to breadcrumbs here wit hthe BreadCrumb object and let the update function inside of breadcrumbs handle the insertion?

ip1981: It looks complicated.

johnsonw: I might be reading it incorrectly. I was thinking since you have the breadcrumbs component, wouldn't it make sense to maintain the state inside of that component? And if the state is maintained inside of the component, shouldn't we send messages to the component whenever the state should be updated?

ip1981: I think messages should be used (inevitably) when we are dealing with async operations. And this is not the case. Breadcrumbs are just a set of helper functions.

johnsonw: Perhaps it is more about aesthetics, as what you have clearly works. It just seems cleaner to have the component update its own model instead of a parent updating a child model. In other words:
```
Msg::Action => child.model.count += 1;
```
vs
```
Msg::Action => {
   orders.proxy(Msg::Child).send_msg(child::Msg::Increment);
}
Msg::Child(msg) => {
   child::update(msg, &mut model.child, &mut orders.proxy(Msg::Child));
}
```
In my opinion, the advantage of the second method is that the parent is not responsible for the update logic. Instead, the message is proxied to the appropriate component and handled there.

jgrund: I think this is a nit, but I agree with @johnsonw here. As the app grows larger, it's easier to reason about if a section's model is only mutated within it's own update fn (especially if it has one).

PR #1753, opened 2020-04-01
Add repouri provisioner

jgrund:
Add a new provisioner that will allow provisioning via a passed in
REPO_URI environment variable.

`vagrant provision --provision-with=install-iml-repouri`

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1753)

### DIFF
--- a/vagrant/scripts/install_iml_repouri.sh
+++ b/vagrant/scripts/install_iml_repouri.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+if [[ -z $REPO_URI ]];
+then
+    echo "REPO_URI must be set to use this provisioner"
+    exit 1
+fi
+
+cat <<EOF > /etc/yum.repos.d/manager-for-lustre.repo
+[manager-for-lustre]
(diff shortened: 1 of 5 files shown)


vagrant/scripts/install_iml_repouri.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+if [[ -z $1 ]];
+then
+    echo "REPO_URI must be set to use this provisioner"
```
ip1981, line 5: This can be confusing, because `REPO_URI` is not used anywhere is this script.

ip1981: Technically, environment variables can be passed to scripts => https://www.vagrantup.com/docs/provisioning/shell.html#env

jgrund: `REPO_URI` is what's expected to be set by the user invoking this provisioner.

Issue #1735, opened 2020-03-26
Activity time *may be* in the future


ip1981:
**Describe the bug**

Time on the activity page *may be* in the future. In my case - "in 2 hours" (I'm in UTC+2).
Server's boot time is also in the future. **Reloading the page fixes this.**

**To Reproduce**
1. Goto a server page
2. Reboot the server
3. Open the Activity page
4. See an alert with started and ended time as "in N hours"

**Expected behavior**
Show "started now"

**Screenshots**

![back-to-the-future](https://user-images.githubusercontent.com/131844/77676486-1027a580-6f97-11ea-91c3-bcb0bf5aeb17.png)

![future](https://user-images.githubusercontent.com/131844/77675852-2aad4f00-6f96-11ea-9256-6997f7aed773.png)

Ref. https://github.com/whamcloud/integrated-manager-for-lustre/issues/1633

PR #1755, opened 2020-04-02
Fix session not polling after logout / login

### LINKED ISSUE #1737: Session not polling after logout / login
It appears we stop polling the session when logging out and logging in. I think this may be related to #1735 (since we get offset from session poll headers) and would explain why a refresh fixes it.

ip1981:
Always poll session.

Closes #1737, #1735.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1755)


ip1981:
Closing in favors of https://github.com/whamcloud/integrated-manager-for-lustre/pull/2026.

PR #1764, opened 2020-04-06
Use 2.12.4 for client

jgrund:
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1764)

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -757,7 +757,7 @@ Vagrant.configure('2') do |config|
                      type: 'shell',
                      run: 'never',
                      inline: <<-SHELL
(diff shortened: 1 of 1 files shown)


utopiabound:
Should we centrally define the Lustre version in the vagrant file and then pass it in a param to the various places that need it?

`scripts/install_zfs_no_iml.sh` is still on 2.12.3

I could do that patch.

jgrund:
> Should we centrally define the Lustre version in the vagrant file and then pass it in a param to the various places that need it?

Sure. The actual server profile repos are still using 2.12.3 as well. In the meantime landing this patch won't hurt anything.

PR #1769, opened 2020-04-07
Fix docker GUI build

jgrund:
The iml-gui docker image has started hitting
https://github.com/yarnpkg/yarn/issues/3189.

Make sure we update after adding the yarn source to avoid this.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1769)


jgrund:
> Do we not need to wait on dependencies?

No, this image is only used to copy it's binary contents into the iml-nginx image

PR #1779, opened 2020-04-12
Display tooltips with full date for relative dates

### LINKED ISSUE #1663: Improve relative time appearance
Following up #1633 and #1662 .

There are things to be improved:

1. Work with string representation of dates. If we can't calculate relative time for any reason, we should show the original string.
2. If we calculated relative time, we show an element with that relative time plus a tooltip showing the original time.
3. Do something when a date is in the future relatively to the server date.

ip1981:
Closes #1663.

date1

date2

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1779)


johnsonw:
I noticed there is some similarity between the `date_view` function in the activity and server page. That function feels like a view in general. Is there any reason we can't put that inside the date_view component and then call it with:
```
date_view::date_view(&dm, Some(date_string));
```
?

ip1981:
> I noticed there is some similarity between the `date_view` function in the activity and server page. That function feels like a view in general. Is there any reason we can't put that inside the date_view component and then call it with:
> 
> ```
> date_view::date_view(&dm, Some(date_string));

I think we should keep date string parsing out of the date_view component:

1. Eventually we might want to sort tables by date, and need parsed dates even without the date_view. So we will not need to parse dates twice.
2. Unlikely, but date string format may change, it would be easier to deal with changes locally and not in the component used in many places. Plus we already have got two date string formats.

iml-gui/crate/src/components/date_view.rs
```diff
@@ -0,0 +1,36 @@
...
+use chrono::{offset::FixedOffset, DateTime};
+use seed::{prelude::*, *};
+
+#[derive(Default)]
+pub struct Model {
+    pub(crate) basedate: Option<DateTime<FixedOffset>>,
+}
+
+impl Model {
+    pub(crate) fn date_view<T>(&self, date: &DateTime<FixedOffset>) -> Node<T> {
```
jgrund, line 18: This should be a free fn instead of existing on the model.

iml-gui/crate/src/components/date_view.rs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) 2020 DDN. All rights reserved.
```
jgrund, line 1: This module should just be called `date`.

iml-gui/crate/src/components/date_view.rs
ip1981: Yeah, but this function does not show the model, it shows something which does not belong in the model.

ip1981: On the other side it is just syntactic sugar :)

iml-gui/crate/src/components/date.rs
```diff
@@ -0,0 +1,37 @@
...
+use crate::{
+    components::{attrs, tooltip, Placement},
+    generated::css_classes::C,
+};
+use chrono::{offset::FixedOffset, DateTime};
+use seed::{prelude::*, *};
+
+#[derive(Default)]
+pub struct Model {
+    pub(crate) basedate: Option<DateTime<FixedOffset>>,
```
ip1981, line 14: Eventually, this model can hold user's preferences on how dates should be displayed.

PR #1781, opened 2020-04-13
Use postgres 9.6 instead of 9.2

jgrund:
Update the RPM and docker installs to use postgres 9.6 instead of 9.2.

9.6 has improved JSON support and `INSERT ... ON CONFLICT DO NOTHING/UPDATE` support,
both of which are useful for upcoming features.

See https://www.postgresql.org/about/featurematrix/ for a more complete
comparision.

Centos 8 provides 9.6 in the appstream repo which should make migration
simpler as well.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1781)

### DIFF
--- a/chroma_core/lib/service_config.py
+++ b/chroma_core/lib/service_config.py
@@ -418,19 +418,19 @@ class ServiceConfig(CommandLine):
             raise RuntimeError(error)
 
     def _init_pgsql(self, database):
-        rc, out, err = self.shell(["service", "postgresql", "initdb"])
+        rc, out, err = self.shell(["/usr/pgsql-9.6/bin/postgresql96-setup", "initdb"])
         if rc != 0:
             if "is not empty" not in out:
-                log.error("Failed to initialize postgresql service")
+                log.error("Failed to initialize postgresql 9.6 database")
                 log.error("stdout:\n%s" % out)
                 log.error("stderr:\n%s" % err)
-                raise CommandError("service postgresql initdb", rc, out, err)
+                raise CommandError("/usr/pgsql-9.6/bin/postgresql96-setup initdb", rc, out, err)
         # Always fixup postgres auth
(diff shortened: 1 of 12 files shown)


jgrund:
I'm a bit wary of updating straight to 12 for a few reasons.

- We are supporting an older version of Django and we'd need to test be sure everything still works correctly when jumping a few major versions.
- Centos8 supports 9.6 directly. Once we move to 8, I'd prefer to drop this external dep in favor of an OS provided one.

Is there something specific in 12 that is not provided in 9.x that we need to have?

This patch passes the RPM based system testing (https://build.whamcloud.com/job/iml-rpm-system-test/600/), and provides docker as well everything should be covered.

ip1981:
What about data migration?

jgrund:
> What about data migration?

It's doable via a manual process. I'm not worried about automating the process for it yet.

mkpankov:
> We are supporting an older version of Django and we'd need to test be sure everything still works correctly when jumping a few major versions.

We're still jumping a few major versions with `9.2` to `9.6`. Postgres changed their policy after `10` so `10` to `11` is the same as `9.2` to `9.3`.

On OS-provided RPM, I agree.

mkpankov:
Can you please check that `psql`, `createdb` and so on are available w/o path modification? This feels wrong. In my case I still had the tools from older `9.2` package and they kept working, but proper tools were in `/usr/pgsql-12/bin` instead and this breaks on clean system.

jgrund:
> Can you please check that psql, createdb and so on are available w/o path modification? This feels wrong. In my case I still had the tools from older 9.2 package and they kept working, but proper tools were in /usr/pgsql-12/bin instead and this breaks on clean system.

They are. My patch wouldn't have passed test if this was not the case.

PR #1787, opened 2020-04-15
Extract systemd into its own crate

johnsonw:
Extract systemd into its own crate

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1787)

### DIFF
--- a/iml-systemd/src/systemd_error.rs
+++ b/iml-systemd/src/systemd_error.rs
@@ -1,0 +1,89 @@
+use iml_cmd::CmdError;
+use std::{fmt, process::Output};
+
+#[derive(Debug)]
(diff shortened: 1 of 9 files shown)


iml-systemd/src/systemd_error.rs
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+impl From<CmdError> for SystemdError {
+    fn from(err: CmdError) -> Self {
+        SystemdError::CmdError(err)
+    }
+}
+
+impl serde::Serialize for SystemdError {
```
jgrund, line 82: What happens when this is derived instead of implemented directly?

PR #1796, opened 2020-04-16
1353 steps modal

nlinker:
out

Implements modal dialog window with the detailed information on the steps.

Fixes https://github.com/whamcloud/integrated-manager-for-lustre/issues/1353

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1796)


jgrund:
Some feedback found while testing our manually

1. If a modal is opened and a step is opened and then closed, the step cannot be expanded again (see GIF below)
1. When jobs / steps are first loading there should be a spinner to indicate to the user that they are being fetched

command-modal

jgrund:
I think we should do away with the steps area and just nest it directly under a job like this:

Screen Shot 2020-04-17 at 4 21 03 PM

It can be expanded / contracted by clicking on the job (should probably add a chevron as well like the top command).

iml-gui/crate/Cargo.toml
```diff
@@ -14,6 +14,8 @@ crate-type = ["cdylib"]
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3"
+rand_core = "0.5.1"
```
ip1981, line 17: Sort :)

iml-gui/crate/src/dependency_tree.rs
```diff
@@ -0,0 +1,444 @@
...
+
+pub fn build_direct_dag<K, T>(ts: &[Arc<T>]) -> DependencyDAG<K, T>
+where
+    K: Hash + Eq + Copy + Debug,
+    T: Deps<K> + Clone + Debug,
+{
+    let mut roots: Vec<K> = Vec::new();
+    let mut dag: Vec<(K, Vec<K>)> = Vec::with_capacity(ts.len());
+    if ts.len() == 1 {
+        // special case, when there is no any arc `x -> y`,
```
ip1981, line 120: Is it possible that `ts.len() > 1`, but none of the `ts` has dependencies?

ip1981: Maybe put all in roots initially, and then strike out deps?

iml-gui/crate/src/dependency_tree.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+    enrich_dag(ts, &roots, &dag)
+}
+
+pub fn build_inverse_dag<K, T>(ts: &[Arc<T>]) -> DependencyDAG<K, T>
+where
+    K: Hash + Eq + Copy + Debug,
+    T: Deps<K> + Clone + Debug,
+{
+    let mut roots: HashSet<K> = ts.iter().map(|t| t.id()).collect();
```
ip1981, line 150: Why does `build_inverse_dag` look simpler than `build_direct_dag`?

iml-gui/crate/src/dependency_tree.rs
```diff
@@ -0,0 +1,444 @@
...
+        ];
+        let x_arcs: Vec<Arc<X>> = x_list.into_iter().map(Arc::new).collect();
+
+        let dag = build_direct_dag(&x_arcs);
+        let result = build_dag_str(&dag, &x_to_string).join("\n");
+        assert_eq!(result, TREE_DIRECT);
+
+        let dag = build_inverse_dag(&x_arcs);
+        let result = build_dag_str(&dag, &x_to_string).join("\n");
+        assert_eq!(result, TREE_INVERSE);
```
ip1981, line 284: The test are looking much more bulky and complicated than the code they test :) Maybe you could just test (compare) fields of `DependencyDAG`?

iml-gui/crate/src/dependency_tree.rs
```diff
@@ -0,0 +1,444 @@
...
+    }
+
+    #[test]
+    fn cyclic_dependency() {
```
ip1981, line 288: If DAG stands for "Directed acyclic graph", why do you test cyclic deps? :)

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -880,19 +821,19 @@ impl Model {
...
-                let a0 = self.commands.iter().find(|x| x.id() == *i);
-                let b0 = self.jobs.iter().find(|x| x.id() == *j);
-                let cs = self.steps.iter().filter(|x| ks.contains(&x.id())).collect::<Vec<_>>();
-                let c0 = if cs.is_empty() { None } else { Some(cs) };
-                match (a0, b0, &c0) {
+                let a0 = self.commands.get(i);
+                let b0 = self.jobs.get(j);
+                let cs = ks.iter().all(|k| self.steps.contains_key(k));
+                let c0 = if cs { None } else { Some(()) };
+                match (a0, b0, c0) {
```
ip1981, line 819: You could just match `cs` against true and false.

iml-gui/crate/src/dependency_tree.rs
nlinker: It is true, although the absence of cycles is not enforced by types or something, and we try to build trees for the external data.

iml-gui/crate/src/dependency_tree.rs
nlinker: The function `build_dag_str` is valuable by itself, despite it is contained in the tests only :-)
Jokes aside, the function `command_modal::job_dag_view` was built on `build_dag_str` as a sample.
The part of the problem, that I have not found an easy way to test `command_modal::job_dag_view`.

iml-gui/crate/src/dependency_tree.rs
nlinker: Yeah, I'll test the case.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -64,87 +106,114 @@ pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg, GMsg>)
...
-        .await
-}
-
-const fn is_finished(cmd: &Command) -> bool {
-    cmd.complete
+fn schedule_fetch_tree(model: &mut Model, orders: &mut impl Orders<Msg, GMsg>) {
+    match &model.select {
+        Select::None => {
+            // the user has all the commands dropdowns closed
+            let cmd_ids = extract_sorted_keys(&model.commands);
```
nlinker, line 185: *Side note*: we may put an additional ordered array `commands_order: Vec<u32>` to `Model` and maintain it to be consistent with `commands`. In this case this will complicate the implementation even more, consume more memory but will be a bit faster, since we don't need to rebuld `cmd_ids` every time.
The same is for `jobs` and `steps`.

iml-gui/crate/src/dependency_tree.rs
nlinker: Indeed, there was a problem here. Good eye!

iml-gui/crate/src/components/command_modal.rs
nlinker: I tried to preserve similarities between the cases. Anyways, it is a bit different now.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +537,601 @@ fn close_button() -> Node<Msg> {
...
+            model.assign_steps(&s);
+            let expected_cmd = model.commands_view.clone();
+            let expected_jobs = model.jobs_graph.clone();
+            let expected_steps = model.steps_view.clone();
+
+            model.clear();
+            model.select = select.clone();
+            shuffle(&mut permutation, &mut rng);
+
+            // we simulate, that FetchCommands, FetchJobs and FetchSteps come in arbitrary order
```
ip1981, line 1009: This test must be unstable, if the problem really existed. I can't see how assign_foo may depend on the order they are invoked.

There are only 3! = 6 permutations, and one of them is already covered above, so you can just write them all down here without RNG. But still I don't see the problem this test is trying catch.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +537,601 @@ fn close_button() -> Node<Msg> {
...
+        assert_eq!(&pair, &(Select::Command(77), false));
+    }
+
+    /// There could the the problem, when `schedule_fetch_tree` made the api requests
+    /// in one order, but during delays or something the responses may come in different
+    /// order. In all such cases, however, model should remain consistent,
+    /// and the test checks this.
+    #[test]
+    fn test_async_handlers_consistency() {
+        let mut rng = Xoroshiro64Star::seed_from_u64(485369);
```
ip1981, line 976: These are not really random (even pseudo) numbers, are they?

iml-gui/crate/src/dependency_tree.rs
nlinker: Could you define _simpler_? They are almost the same in terms of code size and complexity, no?

iml-gui/crate/src/components/command_modal.rs
ip1981: If you print out the permutation on each pass, you will see this every run (i. e. it is not random)

```
[2, 3, 1]
[3, 1, 2]
[2, 3, 1]
[3, 2, 1]
[3, 2, 1]
[3, 2, 1]
[1, 3, 2]
[3, 1, 2]
[3, 2, 1]
```

And you can see that [2,1,3] is missed :D

nlinker: This means, just that the number of `select`-s is too low, given enough generated test cases the proof of the correctness could be made strong enough, i.e. for every `ε` the probability to bypass the false case could be made smaller than `ε`. Agree, this needs a small improvement.

iml-gui/crate/src/components/command_modal.rs
nlinker: It is pseudo random number generator with pretty long period (2^64), sufficient for the testing purposes.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -975,50 +974,39 @@ mod tests {
...
-            shuffle(&mut permutation, &mut rng);
-
-            // we simulate, that FetchCommands, FetchJobs and FetchSteps come in arbitrary order
-            for p in &permutation {
-                match p {
-                    1 => model.assign_commands(&c),
-                    2 => model.assign_jobs(&j),
-                    3 => model.assign_steps(&s),
-                    _ => unreachable!(),
+            let permutations = vec![[1, 2, 3], [1, 3, 2], [2, 1, 3], [2, 3, 1], [3, 1, 2], [3, 2, 1]];
```
ip1981, line 992: You still can omit the first case (1, 2, 3), because it is the same. Or, strictly speaking, one of the six cases :)

iml-gui/crate/src/components/command_modal.rs
ip1981: I was trying to tell that you are using a fixed seed and are getting the same sequence of numbers each time. There is no more randomness than in https://xkcd.com/221/ :)

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +548,638 @@ fn close_button() -> Node<Msg> {
...
+    fn assign_steps(&mut self, steps: &[Step]) {
+        self.steps = convert_to_rich_hashmap(steps.to_vec(), extract_children_from_step);
+        let tuple = self.check_consistency(&self.select);
+        self.refresh_view(tuple);
+    }
+
+    /// We perform the consistency check of the current collections
+    /// `self.commands`, `self.jobs` and `self.steps`.
+    /// The selection, if it is non-empty, places additional constraints.
+    fn check_consistency(&self, select: &Select) -> (bool, bool, bool) {
```
ip1981, line 805:
```suggestion
    fn check_consistency(&self) -> (bool, bool, bool) {
```

(and use `&self.select`)

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +548,638 @@ fn close_button() -> Node<Msg> {
...
+                        1 => model.assign_commands(&c),
+                        2 => model.assign_jobs(&j),
+                        3 => model.assign_steps(&s),
+                        _ => unreachable!(),
+                    }
+                }
+                // compare debug strings since there is no Eq instance for iml_wire_types
+                assert_eq!(format!("{:?}", model.commands_view), format!("{:?}", expected_cmd));
+                assert_eq!(format!("{:?}", model.jobs_graph), format!("{:?}", expected_jobs));
+                assert_eq!(format!("{:?}", model.steps_view), format!("{:?}", expected_steps));
```
ip1981, line 1019: Now I see that you compare `commands_view`, not `commands` :)

But I still think this test has no value, because you call `refresh_view()` and `check_consistency()` on each `assign_foo()` invocation, and the last one will have all the data needed to refresh the view.

It is still unclear to me, so here are the questions:

1. How do you define the consistent state?
2. If you define a consistent state, why does it depend on how we get to the state?
3. Can you provide with an example when this test fails?

What I tried:
1. Remove `match select` completely from `check_consistency()`  - the test is passing
2. Use different seeds for the RNG - the test is passing
3. Replace some `true` to `false` in  `check_consistency()` - the test is still passing (!)
4. Combinations of 2 and 3.

So what exactly is this test for? :)

nlinker: Hm, I changed the piece to
```
                if cmd0.is_some() {
                    ls[0] = false;
                }
```
and with the seed `23` these 100 samples was enough `¯\_(ツ)_/¯` to falsify the test:
image

nlinker: Ok, let me describe what I am doing and what problem I'm trying to solve.

The user (implicitly, via GUI) sets the selection to some value, and this makes a number of requests to send to the API. Assume, the requests caused to grab some commands, jobs and steps.

Now what if the steps come first? Should we discard them or write them into the model?
1. if we discard them, then we need to implement some logic to resend the request again to finally get the batch.
2. if we write it to the model, the model now potentially have old commands, no jobs, and the fresh steps.

Option 1 is dangerous, because the logic, that makes the request again can cause `fire loop`, i.e. in one handler we perform the check, and and fire the request, and in another handler we perform slightly different check and fire the different request, and this two handlers makes firing to each other indefinitely. It needs some work to prove that the process converges, which I don't want to do.

Option 2 is better, however to ensure the model is always consistent, I do the following:
- when a message is received, it unconditionally written to a `raw` part of the state
- there is a logic, that rebuilds the `view` part of the state, and ensures that the `view` is always consistent
- the rest of the logic depends only on the `view` part.

In this case, raw part of the model is:
```rust
    pub commands: HashMap<u32, Arc<RichCommand>>,
    pub jobs: HashMap<u32, Arc<RichJob>>,
    pub steps: HashMap<u32, Arc<RichStep>>,
```
and the view part of the model is
```
    pub commands_view: Vec<Arc<RichCommand>>,
    pub jobs_graph: JobsGraph,
    pub steps_view: Vec<Arc<RichStep>>,
```

As a side note, this problem is similar to the implementation of distributed queues, when there is an ordering between the messages and at some point we receive a message out of order, we maintain a local buffer to wait for some time for the needed message while the clients just see the messages always in order and can rely on this.

ip1981: > if cmd0.is_some() {
>                   ls[0] = false;
>                }

But this is the opposite to the above code (initial ls[0] value). Not surprising it fails :)

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +548,638 @@ fn close_button() -> Node<Msg> {
...
 }
+
+async fn fetch_the_batch<T, F, U>(ids: Vec<u32>, data_to_msg: F) -> Result<U, U>
+where
+    T: DeserializeOwned + EndpointName + 'static,
+    F: FnOnce(ResponseDataResult<ApiList<T>>) -> U,
+    U: 'static,
+{
+    // e.g. GET /api/something/?id__in=1&id__in=2&id__in=11&limit=0
+    let err_msg = format!("Bad query for {}: {:?}", T::endpoint_name(), ids);
```
ip1981, line 559: It is executed unconditionally, even when the request is succeeding (in most cases, I hope).

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +548,638 @@ fn close_button() -> Node<Msg> {
...
+    let mut ids: Vec<_> = ids.into_iter().map(|x| ("id__in", x)).collect();
+    ids.push(("limit", 0));
+    Request::api_query(T::endpoint_name(), &ids)
+        .expect(&err_msg)
```
ip1981, line 563: it will panic, won't it?

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -64,87 +115,125 @@ pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg, GMsg>)
...
-            model.opens.insert(x);
-        }
-        Msg::Close(x) => {
-            model.opens.remove(&x);
+        Msg::FetchedJobs(jobs_data_result) => match *jobs_data_result {
+            Ok(api_list) => {
+                model.assign_jobs(&api_list.objects);
+            }
+            Err(e) => {
+                error!("Failed to perform FetchJobs {:#?}", e);
```
ip1981, line 166:
```suggestion
                error!(format!("Failed to perform FetchJobs {:#?}", e));
```
Otherwise it will print `"Failed to perform FetchJobs {:#?}", ...`

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +548,638 @@ fn close_button() -> Node<Msg> {
...
+
+fn extract_uri_id<T: EndpointName>(input: &str) -> Option<u32> {
+    lazy_static::lazy_static! {
+        static ref RE: Regex = Regex::new(r"/api/(\w+)/(\d+)/").unwrap();
+    }
+    RE.captures(input).and_then(|cap: Captures| {
+        let s = cap.get(1).unwrap().as_str();
+        let t = cap.get(2).unwrap().as_str();
+        if s == T::endpoint_name() {
+            t.parse::<u32>().ok()
```
ip1981, line 576: nit:
```suggestion
        if s == T::endpoint_name() {
            let t = cap.get(2).unwrap().as_str();        
            t.parse::<u32>().ok()
```

ip1981:
Do I get it right, that the user cannot "select" multiple commands? Is it possible to have multiple commands? Even if not, I think I could bring a general model :)

1. We build a tree as we get the data (commands, jobs, step). No matter in which order. We just build a tree that can be displayed to the user. And yes, we need to just store all the commands, jobs, steps as we receive them from the backend. We display as much valid tree as we can.
2. The user clicks on items he wants to  see, we store user's choices
3. When displaying the tree, we walk over it and display taking user's choices into account.

For example, initially user has not made any choice and we only show the commands is any.

For the point 2 we in fact need a relational table: `(cmd_id: u32, job_id: Option<u32> , step_id: Option<u32>)`. Implementation may from trivial to some indexed set, with some optimizations in between these edge cases.

So when we are about to show a command, we look up the table of user's choices: if there are entries for that command, we show the command's details (jobs). When we are about to show a job,  we look up the table of user's choices: if there are entries for that job, we show the jobs's details (steps). And so on.

So user's choices act as a filter to the tree.

Here is some trivial implementation => https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=996a4737827b83e011977e4111e9c9c3 :)

Only that user's choice table would need some tests ;-)

iml-gui/crate/src/components/command_modal.rs
nlinker: It seems yes, but it shouldn't be fatal. I'll check.

iml-gui/crate/src/components/command_modal.rs
nlinker: I promise, I fairly diced the roll 6 times!

Well, I am sure the tests should be deterministic. If I seeded it from entropy (if I somehow overlooked a bug), then it would fail sometimes in the CI, which might mess some other's work. And this work might be more important than the bug, that we might never normally see.

iml-gui/crate/src/components/command_modal.rs
nlinker: It does panic, but the error happens only when serde cannot build parameter string `"id__in=1&id__in=2..."` which is  impossible normally.

iml-gui/crate/src/components/command_modal.rs
nlinker: Meh, this visually less nice and the condition always true for all current usages...

iml-gui/crate/src/components/command_modal.rs
nlinker: Done (there was a test, that made multiple `select` attempts for one model instance, now it is gone)

nlinker:
> Do I get it right, that the user cannot "select" multiple commands? Is it possible to have multiple commands? Even if not, I think I could bring a general model :)

Yes, selecting different commands impossible, this is the restriction brought by `Select` type. Let me take a look at your code tomorrow.

nlinker:
> 2. When jobs / steps are first loading there should be a spinner to indicate to the user that they are being fetched

Omg, at the time I checked that, I had set up a proxy to simulate delays and had verified that it works smooth without any spinners before I realized today, that I customized the proxy in the wrong way and the requests are not delayed. :oops:
Now I have fixed the proxy and see the problem. I'll add the spinners tomorrow, I expect it to be pretty easy.

jgrund:
Testing this out, looks like the following payload produces an issue:

```
{\"meta\": {\"limit\": 1000, \"next\": null, \"offset\": 0, \"previous\": null, \"total_count\": 8}, \"objects\": [{\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"ClearOldStratagemDataJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.180608\", \"description\": \"Removing old Stratagem data\", \"errored\": false, \"id\": 97, \"modified_at\": \"2020-04-28T20:21:56.180583\", \"read_locks\": [], \"resource_uri\": \"/api/job/97/\", \"state\": \"complete\", \"step_results\": {\"/api/step/135/\": null}, \"steps\": [\"/api/step/135/\"], \"wait_for\": [], \"write_locks\": []}, {\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"RunStratagemJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.190074\", \"description\": \"Stratagem: Scanning MDT fs0a9c-MDT0000\", \"errored\": false, \"id\": 98, \"modified_at\": \"2020-04-28T20:21:56.190047\", \"read_locks\": [{\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}], \"resource_uri\": \"/api/job/98/\", \"state\": \"complete\", \"step_results\": {\"/api/step/136/\": [\"/tmp/b56fb819-1d9c-412a-b411-ceaa5c848bec/\", {\"group_counters\": [{\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE < 1M\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1T\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1G\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"1M <= SIZE < 1G\", \"size\": 0}], \"name\": \"size_distribution\"}, {\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"classify\": {\"attr_type\": \"uid\", \"counters\": [], \"flist_type\": \"none\"}, \"count\": 0, \"expression\": \"!= type S_IFDIR\", \"flist_type\": \"none\", \"name\": \"top_inode_users\", \"size\": 0}], \"name\": \"user_distribution\"}]}, []], \"/api/step/143/\": null}, \"steps\": [\"/api/step/136/\", \"/api/step/143/\"], \"wait_for\": [\"/api/job/97/\"], \"write_locks\": []}, {\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"RunStratagemJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.218393\", \"description\": \"Stratagem: Scanning MDT fs0a9c-MDT0001\", \"errored\": false, \"id\": 99, \"modified_at\": \"2020-04-28T20:21:56.218374\", \"read_locks\": [{\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}], \"resource_uri\": \"/api/job/99/\", \"state\": \"complete\", \"step_results\": {\"/api/step/137/\": [\"/tmp/56fdca2c-93e5-4397-80da-904ece94b011/\", {\"group_counters\": [{\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE < 1M\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1T\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1G\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"1M <= SIZE < 1G\", \"size\": 0}], \"name\": \"size_distribution\"}, {\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"classify\": {\"attr_type\": \"uid\", \"counters\": [], \"flist_type\": \"none\"}, \"count\": 0, \"expression\": \"!= type S_IFDIR\", \"flist_type\": \"none\", \"name\": \"top_inode_users\", \"size\": 0}], \"name\": \"user_distribution\"}]}, []], \"/api/step/141/\": null}, \"steps\": [\"/api/step/137/\", \"/api/step/141/\"], \"wait_for\": [\"/api/job/97/\"], \"write_locks\": []}, {\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"RunStratagemJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.233545\", \"description\": \"Stratagem: Scanning MDT fs0a9c-MDT0002\", \"errored\": false, \"id\": 100, \"modified_at\": \"2020-04-28T20:21:56.233526\", \"read_locks\": [{\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}], \"resource_uri\": \"/api/job/100/\", \"state\": \"complete\", \"step_results\": {\"/api/step/139/\": [\"/tmp/b363f759-2e97-4d4e-9496-d652072f5024/\", {\"group_counters\": [{\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE < 1M\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1T\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1G\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"1M <= SIZE < 1G\", \"size\": 0}], \"name\": \"size_distribution\"}, {\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"classify\": {\"attr_type\": \"uid\", \"counters\": [], \"flist_type\": \"none\"}, \"count\": 0, \"expression\": \"!= type S_IFDIR\", \"flist_type\": \"none\", \"name\": \"top_inode_users\", \"size\": 0}], \"name\": \"user_distribution\"}]}, []], \"/api/step/142/\": null}, \"steps\": [\"/api/step/139/\", \"/api/step/142/\"], \"wait_for\": [\"/api/job/97/\"], \"write_locks\": []}, {\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"RunStratagemJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.248139\", \"description\": \"Stratagem: Scanning MDT fs0a9c-MDT0003\", \"errored\": false, \"id\": 101, \"modified_at\": \"2020-04-28T20:21:56.248121\", \"read_locks\": [{\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}], \"resource_uri\": \"/api/job/101/\", \"state\": \"complete\", \"step_results\": {\"/api/step/138/\": [\"/tmp/79c6d202-ef07-40a9-bb5b-955fe969a84c/\", {\"group_counters\": [{\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE < 1M\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1T\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"SIZE >= 1G\", \"size\": 0}, {\"blocks\": 0, \"count\": 0, \"flist_type\": \"none\", \"name\": \"1M <= SIZE < 1G\", \"size\": 0}], \"name\": \"size_distribution\"}, {\"counters\": [{\"blocks\": 0, \"count\": 4, \"flist_type\": \"none\", \"name\": \"Other\", \"size\": 0}, {\"blocks\": 0, \"classify\": {\"attr_type\": \"uid\", \"counters\": [], \"flist_type\": \"none\"}, \"count\": 0, \"expression\": \"!= type S_IFDIR\", \"flist_type\": \"none\", \"name\": \"top_inode_users\", \"size\": 0}], \"name\": \"user_distribution\"}]}, []], \"/api/step/140/\": null}, \"steps\": [\"/api/step/138/\", \"/api/step/140/\"], \"wait_for\": [\"/api/job/97/\"], \"write_locks\": []}, {\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"AggregateStratagemResultsJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.262532\", \"description\": \"Aggregating stratagem scan results in influxdb.\", \"errored\": false, \"id\": 102, \"modified_at\": \"2020-04-28T20:21:56.262513\", \"read_locks\": [], \"resource_uri\": \"/api/job/102/\", \"state\": \"complete\", \"step_results\": {\"/api/step/144/\": null}, \"steps\": [\"/api/step/144/\"], \"wait_for\": [\"/api/job/98/\", \"/api/job/99/\", \"/api/job/100/\", \"/api/job/101/\"], \"write_locks\": []}, {\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"MountLustreClientJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.270014\", \"description\": \"Mount co-ef-pm-143:/mnt/fs0a9c (mounted)\", \"errored\": false, \"id\": 103, \"modified_at\": \"2020-04-28T20:21:56.269996\", \"read_locks\": [{\"locked_item_content_type_id\": 50, \"locked_item_id\": 6, \"locked_item_uri\": \"/api/lnet_configuration/6/\", \"resource_uri\": \"\"}, {\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}, {\"locked_item_content_type_id\": 16, \"locked_item_id\": 6, \"locked_item_uri\": \"/api/host/6/\", \"resource_uri\": \"\"}, {\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}, {\"locked_item_content_type_id\": 50, \"locked_item_id\": 6, \"locked_item_uri\": \"/api/lnet_configuration/6/\", \"resource_uri\": \"\"}, {\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}, {\"locked_item_content_type_id\": 16, \"locked_item_id\": 6, \"locked_item_uri\": \"/api/host/6/\", \"resource_uri\": \"\"}, {\"locked_item_content_type_id\": 23, \"locked_item_id\": 1, \"locked_item_uri\": \"/api/filesystem/1/\", \"resource_uri\": \"\"}], \"resource_uri\": \"/api/job/103/\", \"state\": \"complete\", \"step_results\": {\"/api/step/145/\": null}, \"steps\": [\"/api/step/145/\"], \"wait_for\": [\"/api/job/102/\"], \"write_locks\": [{\"locked_item_content_type_id\": 35, \"locked_item_id\": 2, \"locked_item_uri\": \"/api/client_mount/2/\", \"resource_uri\": \"\"}]}, {\"available_transitions\": [], \"cancelled\": false, \"class_name\": \"SendStratagemResultsToClientJob\", \"commands\": [\"/api/command/33/\"], \"created_at\": \"2020-04-28T20:21:56.367999\", \"description\": \"Sending stratagem results to client\", \"errored\": false, \"id\": 104, \"modified_at\": \"2020-04-28T20:21:56.367980\", \"read_locks\": [], \"resource_uri\": \"/api/job/104/\", \"state\": \"complete\", \"step_results\": {\"/api/step/146/\": null}, \"steps\": [\"/api/step/146/\"], \"wait_for\": [\"/api/job/103/\"], \"write_locks\": []}]}
```

Screen Shot 2020-04-28 at 5 02 23 PM

iml-gui/crate/src/components/font_awesome.rs
```diff
@@ -27,3 +27,34 @@ fn font_awesome_base<T>(sprite_sheet: &str, more_attrs: Attrs, icon_name: &str)
         ]
     ]
 }
+
+// cool symbols
+pub fn font_awesome_minus_circle<T>(awesome_style: Attrs) -> Node<T> {
```
jgrund, line 32: Why do we inline these instead of calling the correct thing from the svg sheet?

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -3,27 +3,68 @@
...
 
-#[derive(Default, Debug)]
-pub struct Model {
-    cancel: Option<oneshot::Sender<()>>,
-    loading: bool,
-    opens: HashSet<u32>,
-    pub commands_all: Vec<Arc<Command>>,
-    pub commands: Vec<Arc<Command>>,
-    pub modal: modal::Model,
+type Job0 = Job<Option<()>>;
```
jgrund, line 28:
```suggestion
type Job0 = Job<Option<serde_json::Value>>;
```

This can't be hardcoded to unit, args coming back can be anything, and can vary between / within a collection.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -188,25 +274,26 @@ pub(crate) fn view(model: &Model) -> Node<Msg> {
...
-    } else if x.cancelled {
-        C.border_gray_500
-    } else if x.errored {
-        C.border_red_500
     } else if x.complete {
         C.border_green_500
+    } else if x.errored {
+        C.border_red_500
+    } else if x.cancelled {
+        C.border_gray_500
```
jgrund, line 286: This ordering needs to be reverted. All commands will be marked complete when they finished, it's the absence of other states that makes them successful.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -226,45 +313,219 @@ fn command_item_view(x: &Command, open: bool) -> Node<Msg> {
...
 }
 
-fn status_icon<T>(cmd: &Command) -> Node<T> {
-    let cls = class![C.w_4, C.h_4, C.inline, C.mr_4];
-
-    if cmd.cancelled {
-        font_awesome(cls, "ban").merge_attrs(class![C.text_gray_500])
+fn cmd_status_icon<T>(cmd: &RichCommand) -> Node<T> {
+    let awesome_class = class![C.w_4, C.h_4, C.inline, C.mr_4];
+    if cmd.complete {
```
jgrund, line 490: The ordering here needs to be reverted, complete *must* come last in the order.

jgrund:
> Do I get it right, that the user cannot "select" multiple commands? Is it possible to have multiple commands? Even if not, I think I could bring a general model :)

The user can possibly "select" multiple commands in the future. I.E. consider the usecase that there are multiple commands in flight and we show a notification regarding them. The user clicks the notification and the command modal pops up to show both commands within it.

iml-gui/crate/Cargo.toml
```diff
@@ -28,6 +29,7 @@ js-sys = "0.3"
 lazy_static = "1.4.0"
 natord = "1.0"
 number-formatter = { path = "../../number-formatter", version = "0.1" }
+rand_core = "0.5.1"
```
jgrund, line 32: Why is this not a dev-dependency?

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +547,643 @@ fn close_button() -> Node<Msg> {
...
+        all_steps: Vec<Step>,
+    }
+
+    impl Db {
+        fn select_cmds(&self, is: &[u32]) -> Vec<Command> {
+            self.all_cmds
+                .iter()
+                .filter(|x| is.contains(&x.id))
+                .map(|x| x.clone())
+                .collect()
```
jgrund, line 917:
```suggestion
        fn select_cmds(&self, is: &[u32]) -> Vec<&Command> {
            self.all_cmds
                .iter()
                .filter(|x| is.contains(&x.id))
                .collect()
```

We should not need to clone here.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +547,643 @@ fn close_button() -> Node<Msg> {
...
+        }
+        fn select_jobs(&self, is: &[u32]) -> Vec<Job0> {
+            self.all_jobs
+                .iter()
+                .filter(|x| is.contains(&x.id))
+                .map(|x| x.clone())
+                .collect()
```
jgrund, line 924:
```suggestion
        fn select_jobs(&self, is: &[u32]) -> Vec<&Job0> {
            self.all_jobs
                .iter()
                .filter(|x| is.contains(&x.id))
                .collect()
```

We should not need to clone here.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +547,643 @@ fn close_button() -> Node<Msg> {
...
+        }
+        fn select_steps(&self, is: &[u32]) -> Vec<Step> {
+            self.all_steps
+                .iter()
+                .filter(|x| is.contains(&x.id))
+                .map(|x| x.clone())
+                .collect()
```
jgrund, line 931:
```suggestion
        fn select_steps(&self, is: &[u32]) -> Vec<&Step> {
            self.all_steps
                .iter()
                .filter(|x| is.contains(&x.id))
                .collect()
```

We should not need to clone here.

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +547,643 @@ fn close_button() -> Node<Msg> {
...
+        let jobs = db.select_jobs(&c_ids);
+        let j_ids = jobs
+            .iter()
+            .map(|x| extract_ids::<Step>(&x.steps))
+            .flatten()
+            .collect::<Vec<u32>>();
+        let steps = db.select_steps(&j_ids);
+        let cmds = db.all_cmds.clone(); // use all roots
+        (cmds, jobs, steps)
+    }
```
jgrund, line 1134:
```suggestion
  fn prepare_subset(db: &Db, id: u32) -> (Vec<&Command>, Vec<&Job0>, Vec<&Step>) {
        let cmds = db.select_cmds(&vec![id]);
        let c_ids = cmds
            .iter()
            .map(|x| extract_ids::<Job0>(&x.jobs))
            .flatten()
            .collect::<Vec<u32>>();
        let jobs = db.select_jobs(&c_ids);
        let j_ids = jobs
            .iter()
            .map(|x| extract_ids::<Step>(&x.steps))
            .flatten()
            .collect::<Vec<u32>>();
        let steps = db.select_steps(&j_ids);
        let cmds = db.all_cmds.iter().collect(); // use all roots
        (cmds, jobs, steps)
    }
```

No need to clone here.

nlinker: Sorry, Joe, not that easy. :mask:

The thing is that `prepare_subset` function resembles going for the data from the remote API. The handlers `Msg::FetchedCommands`, `Msg::FetchedJobs` and `Msg::FetchedSteps` accept `ApiList<T>` and they need to build vectors of `RichCommand`s, `RichJob`s and `RichStep`s respectively (therefore they need to own the contents!). As a consequence, `assign_commands` cannot accept __something of__ `&Command`. The same is for jobs and steps.

I removed the unnecessary clones in the `assign_commands`, `assign_jobs` and `assign_steps` by changing their signatures:
```rust
+    fn assign_commands(&mut self, cmds: Vec<Command>) {
+        self.commands = convert_to_rich_hashmap(cmds, extract_children_from_cmd);
-     fn assign_commands(&mut self, cmds: &[Command]) {
-        self.commands = convert_to_rich_hashmap(cmds.to_vec(), extract_children_from_cmd);
```
but other than that I cannot see an easy way of getting rid of `clone`s.

I could potentially get rid of `Rich<K, V>` wrapper, also reducing the number of clones needed, however then I needed to create an `Iterator<Item=u32>` over the vector of strings like `vec!["/api/command/111/", "/api/command/112/", "/api/command/113/"]`, and this parsing was happen each time the method `.deps()` is accessed.

iml-gui/crate/Cargo.toml
nlinker: Forgot to remove after `shuffle` removal.

iml-gui/crate/src/components/command_modal.rs
nlinker: Thanks!

iml-gui/crate/src/components/command_modal.rs
nlinker: I see, thanks.

iml-gui/crate/src/components/command_modal.rs
nlinker: Let's continue to discuss it there https://github.com/whamcloud/integrated-manager-for-lustre/pull/1796/files#r417082999

iml-gui/crate/src/components/command_modal.rs
nlinker: Let's continue to discuss it there: https://github.com/whamcloud/integrated-manager-for-lustre/pull/1796/files#r417082999

iml-gui/crate/src/components/command_modal.rs
nlinker: Let's continue to discuss it there: https://github.com/whamcloud/integrated-manager-for-lustre/pull/1796/files#r417082999

iml-gui/crate/src/components/font_awesome.rs
nlinker: Because these icons are from the _light_ set, which is not present in the fonts we use. 

I tried `plus-square` and `minus-square` from the _regular_ set:
image

`minus-circle` and `plus-circle` from the _solid_ set:
image

These (hardcoded) icons:
image

For me, the latter ones are the clear winners. 😊

iml-gui/crate/src/components/command_modal.rs
nlinker: I see, thanks

iml-gui/crate/src/components/command_modal.rs
nlinker: Ok, thanks.

iml-gui/crate/src/components/command_modal.rs
jgrund: Hrrm, I tried this locally yesterday and everything compiled except the tests, which I did not bother to update.

jgrund:
Looks like docker test compilation is failing with this patch: https://build.whamcloud.com/job/manager-for-lustre/300/

iml-gui/crate/src/components/font_awesome.rs
jgrund: How about we use the first ones, and make them a gray color.

I don't think we should inline the svg icons into our app.

nlinker: Done

nlinker:
> The iconography for the steps should mirror the jobs and command, i.e.
> 
> * check for success
> * ban sign for failure
> * grey ban sign for cancel
> * spinner for in progress

Hm, it seems that `cancel` is not possible to get for steps:
> __state__: One of incomplete, failed, success

So, you propose to forget about (+) and (-) and use the check/ban/spinner icons for them, right?

jgrund:
The layout of these steps seem strange to me:

Screen Shot 2020-05-01 at 2 53 13 PM


I would have expected more depth as there are some non-shallow `wait_for` items:

Screen Shot 2020-05-01 at 2 59 42 PM

[localhost.har.zip](https://github.com/whamcloud/integrated-manager-for-lustre/files/4565911/localhost.har.zip)
[localhost.har.zip](https://github.com/whamcloud/integrated-manager-for-lustre/files/4565913/localhost.har.zip)

nlinker:
> The iconography for the steps should mirror the jobs and command, i.e.
> 
> * check for success
> * ban sign for failure
> * grey ban sign for cancel
> * spinner for in progress
> 
> In addition, it looks like there are lot's of things being written to the log:
> 
> Screen Shot 2020-05-01 at 2 47 16 PM

Ugh. Probably I need to make some machinery to avoid such debugging scaffolding from committing.

jgrund:
>Hm, it seems that cancel is not possible to get for steps:
>state: One of incomplete, failed, success
> So, you propose to forget about (+) and (-) and use the check/ban/spinner icons for them, right?

I don't think that description is accurate... I'm pretty sure there is a cancelled state as well.

https://github.com/whamcloud/GUI/blob/0cbf30d7ab36746619a3b53617c1a14a8e8d8622/source/iml/command/step-modal.js#L40-L44

I'd try it out, but it looks like ability to cancel jobs is not part of this patch.

https://github.com/whamcloud/GUI/blob/0cbf30d7ab36746619a3b53617c1a14a8e8d8622/source/iml/command/command-modal.js#L101-L125

jgrund:
We should also show the log / backtrace if there is one on error. 

Screen Shot 2020-05-01 at 4 59 57 PM

Screen Shot 2020-05-01 at 5 00 01 PM

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -598,143 +592,14 @@ fn is_all_finished(cmds: &HashMap<u32, Arc<RichCommand>>) -> bool {
...
-            Select::CommandJob(_, _) => cur_select,
-            Select::CommandJobSteps(cs, js, ss) => {
-                let ids = remove_from_sorted(&ss, step_id);
-                if ids.is_empty() {
-                    Select::CommandJob(cs, js)
-                } else {
-                    Select::CommandJobSteps(cs, js, ids)
-                }
-            }
-        },
```
nlinker: -109

iml-gui/crate/src/dependency_tree.rs
nlinker: The function `traverse_graph` introduced and now can be used for the tests and in the `command_model` code as well.

nlinker:
> The layout of these steps seem strange to me:
> I would have expected more depth as there are some non-shallow `wait_for` items:

Resolved:
image

iml-gui/crate/src/components/command_modal.rs
```diff
@@ -284,3 +539,662 @@ fn close_button() -> Node<Msg> {
...
+
+fn extract_sorted_keys<T>(hm: &HashMap<u32, T>) -> Vec<u32> {
+    let mut ids = hm.keys().copied().collect::<Vec<_>>();
+    ids.sort();
+    ids
+}
+
+fn convert_to_sorted_vec<T: Clone>(hm: &HashMap<u32, T>) -> Vec<T> {
+    extract_sorted_keys(hm).into_iter().map(|k| hm[&k].clone()).collect()
+}
```
jgrund, line 646:
```suggestion
fn convert_to_sorted_vec<T: Clone>(hm: &HashMap<u32, Arc<T>>) -> Vec<Arc<T>> {
    extract_sorted_keys(hm)
        .into_iter()
        .map(|k| Arc::clone(&hm[&k]))
        .collect()
}
```

Nit, this would make it clearer that the clone is a ref count bump instead of a deep clone (This still compiles btw).

PR #1811, opened 2020-04-20
Update deploy restart policy

johnsonw:
The restart policy needs to be changed from `on-failure` to `any`.
`Setup` is an exception and should continue to have a restart policy of
`on-failure`.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1811)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.7"
 x-deploy: &default-deploy
   restart_policy:
-    condition: on-failure
+    condition: any
     delay: 5s
     window: 5s
 services:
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.7"
 x-deploy: &default-deploy
   restart_policy:
-    condition: on-failure
+    condition: any
```
ip1981, line 4: Even when manually stopped?

johnsonw: > Always restart the container if it stops. If it is manually stopped, it is restarted only when Docker daemon restarts or the container itself is manually restarted. (See the second bullet listed in restart policy details)
-- https://docs.docker.com/config/containers/start-containers-automatically/

PR #1815, opened 2020-04-22
Check for installed repo RPM before installing

mkpankov:
Otherwise it fails when provision was aborted or,
with PostgreSQL repo, always when deploying to
previously deployed machine.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1815)

### DIFF
--- a/vagrant/scripts/install_iml_local.sh
+++ b/vagrant/scripts/install_iml_local.sh
@@ -4,8 +4,12 @@ set -ex
 
 yum copr enable -y managerforlustre/manager-for-lustre-devel
 # Get latest rpmdevtools
-yum install -y https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
-yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
+if ! yum list installed rpmdevtools-8.10-7.el8.noarch; then
+    yum install -y https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
+fi
+if ! yum list installed pgdg-redhat-repo.noarch; then
+    yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
+fi
 yum install -y rpmdevtools git ed epel-release python-setuptools gcc openssl-devel postgresql96-devel
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 source $HOME/.cargo/env
--- a/vagrant/scripts/install_iml_local.sh
+++ b/vagrant/scripts/install_iml_local.sh
@@ -3,7 +3,7 @@
 set -ex
 
 yum copr enable -y managerforlustre/manager-for-lustre-devel
-# Install repository with latest rpmdevtools
+# Install latest rpmdevtools for rpmdev-bumpspec
 if ! rpm -q rpmdevtools-8.10-7.el8.noarch; then
     yum install -y https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
 fi
@@ -12,7 +12,7 @@ if ! rpm -q pgdg-redhat-repo.noarch; then
     yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
 fi
 # Install the packages themselves
-yum install -y rpmdevtools git ed epel-release python-setuptools gcc openssl-devel postgresql96-devel
+yum install -y git ed epel-release python-setuptools gcc openssl-devel postgresql96-devel
 curl https://sh.rustup.rs -sSf | sh -s -- -y
 source $HOME/.cargo/env
 rustup update
@@ -34,6 +34,7 @@ rpmdev-bumpspec -n $V.$TS rust-iml.spec
 
 make all
 
+# Uninstall rpmdevtools and dependencies because it pull in python 3 and lots of other stuff
 yum autoremove -y rpmdevtools
 
 rm -rf /tmp/{manager,agent}-rpms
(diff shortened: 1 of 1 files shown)


vagrant/scripts/install_iml_local.sh
```diff
@@ -4,8 +4,12 @@ set -ex
...
 # Get latest rpmdevtools
-yum install -y https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
-yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
+if ! yum list installed rpmdevtools-8.10-7.el8.noarch; then
+    yum install -y https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
+fi
+if ! yum list installed pgdg-redhat-repo.noarch; then
+    yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
+fi
 yum install -y rpmdevtools git ed epel-release python-setuptools gcc openssl-devel postgresql96-devel
```
ip1981, line 13: Might want to remove `rpmdevtools` here.

ip1981: And probably `yum autoremove -y rpmdevtools` below, Because why would we want to reinstall it every time?

mkpankov: > Might want to remove rpmdevtools here.

No. The first `yum install` installs the repository, the second installs the package from that repository.

> And probably yum autoremove -y rpmdevtools below, Because why would we want to reinstall it every time?

No idea why it's done this way, the build is pretty brittle so I'm not touching it.

johnsonw, line 13: Need to remove rpmdevtools on line 13

vagrant/scripts/install_iml_local.sh
```diff
@@ -4,8 +4,12 @@ set -ex
 
 yum copr enable -y managerforlustre/manager-for-lustre-devel
 # Get latest rpmdevtools
-yum install -y https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
-yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-7-x86_64/pgdg-redhat-repo-latest.noarch.rpm
+if ! yum list installed rpmdevtools-8.10-7.el8.noarch; then
```
utopiabound, line 7: I don't think the check should be version locked, should it?

johnsonw: I think a grep could be used here to avoid the version right?

mkpankov: We want the specific version, that's the reason for that entire manipulation. It never changes.

vagrant/scripts/install_iml_local.sh
ip1981: I think the build has to be idempotent and re-entrant. As such it will stop being brittle :)

ip1981: > The first yum install installs the repository

Are you sure? It looks like a normal RPM package, and it updates already installed rpmdevtools. The second yum install is indeed adding a repository.

vagrant/scripts/install_iml_local.sh
johnsonw: I'm still a bit confused. This is installing the rpm but then on line 13 it's installing rpmdevtools again.

vagrant/scripts/install_iml_local.sh
mkpankov: > I think the build has to be idempotent and re-entrant.
Yes, but it isn't.
> As such it will stop being brittle :)
No, this is done for some reason unknown to me, so unless we clear that reason I won't touch it. I've been spending hours on debugging obscure build failures last time I tried to dig there.

vagrant/scripts/install_iml_local.sh
utopiabound: you can just do `yum list installed rpmdevtools`
There's no `yum remove` so having a different version isn't checked for.

mkpankov: > I'm still a bit confused. This is installing the rpm but then on line 13 it's installing rpmdevtools again.

First RPM is just RPM with repository data. It's symmetric to what happens with PostgreSQL installation - first repo RPM, then RPM of package itself.

vagrant/scripts/install_iml_local.sh
```diff
@@ -4,8 +4,12 @@ set -ex
...
+    yum install -y https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
+fi
+if ! yum list installed pgdg-redhat-repo.noarch; then
```
utopiabound, line 10: This should also just be `rpm -q pgdg-redhat-repo`

vagrant/scripts/install_iml_local.sh
johnsonw: Ah that's why it's confusing me. Any reason we can't use yum-config-manager to add the repo? http://man7.org/linux/man-pages/man1/yum-config-manager.1.html

johnsonw: To be clear, i'm not trying to hold this up, when I first looked at it it looked like it was installing the same rpm twice (since I didn't know the first rpm was to install the repository). I think something like yum-config-manager would make it a little clearer if it is possible.

mkpankov: @johnsonw 

`yum-config-manager` is nice, I tried what you suggested and it even checks for duplicate repositories. Problem is, it errors when there's a duplicate, and `rpm -q` doesn't show the repo RPM as installed, so the only way around that would be to silence errors from `yum-config-manager`, which isn't great as it can fail for another reason rather than duplicate repo.
```
    adm: adding repo from: https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm
    adm: Cannot add repo from https://copr-be.cloud.fedoraproject.org/results/managerforlustre/buildtools/epel-8-x86_64/01152137-rpmdevtools/rpmdevtools-8.10-7.el8.noarch.rpm as is a duplicate of an existing repo
```

jgrund:
@ip1981 Mentioned we should be building in a chroot ex: https://blog.packagecloud.io/eng/2015/05/11/building-rpm-packages-with-mock/ . 

This would remedy the build-time checking and install / uninstall of build packages

mkpankov:
Rebased and took care of comments, please re-review.

PR #1821, opened 2020-04-23
Add CLI commands for loading / removing profiles

### LINKED ISSUE #1680: Reloading docker does not add new profiles in the /etc/iml-docker/setup directory
**Describe the bug**
When stopping iml-docker and adding a new profile to /etc/iml-docker/setup, I noticed that the profile was not present after starting iml-docker.

**Expected**
I should be able to stop iml-docker, add a new profile to /etc/iml-docker/setup, start iml-docker, and see the new profile when attempting to add a new server.

jgrund:
Add a new CLI command that will enable server profiles to be
loaded / removed post initial setup.

general format is:

```sh
cat profile.json > iml server profile load
```

and 

```sh
iml server profile remove <name>
```

Fixes #1680.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1821)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,7 +130,7 @@ checksum = "da71fef07bc806586090247e971229289f64c210a278ee5ae419314eb386b31d"
 dependencies = [
  "proc-macro2 1.0.10",
  "quote 1.0.3",
- "syn 1.0.17",
+ "syn 1.0.18",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


ip1981:
I would like `iml server profile add <file>` more.

jgrund:
> I would like iml server profile add <file> more. With - for stdin :)

Me too, but `<file>` won't work in docker deployments.

ip1981:
> Me too, but `<file>` won't work in docker deployments.

Yeah, but `cat .file | <some docker thing> iml server profile add - ` will, this is the thing where explicit is better than implicit, especially in the `--help` output.

jgrund:
Yes, `cat .file | iml server profile add` will work on the host in a docker deployment.

However, `iml server profile add foo.profile` will not work on the host, and will be confusing why it didn't work.

jgrund:
iml-server-profile

jgrund:
> It would be nice to have iml server profile load PROFILE.profile but otherwise looks good

If we can figure out how to do so in a docker env, I'm not opposed to adding it; I don't have any good ideas for that right now.

PR #1823, opened 2020-04-24
Implement SFA monitor

### LINKED ISSUE #1806: Implement SFA API monitor
Implement a new service that will monitor select items from the SFA API

## Requirements

- Lives on the manager. Since this service will only need to monitor the API, it does not need to exist on each agent.
- Configurable via endpoints. The service will use a list of endpoints to serve as a request pool. Requests are limited to 1 per endpoint at a time.
- Writes state to db. The service will write state into the database.
- Able to monitor multiple appliances. Has the ability to monitor and differentiate multiple appliances.
- Written in Rust.

jgrund:
Fixes #1806.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1823)


iml-manager-env/src/lib.rs
```diff
@@ -218,6 +219,40 @@ pub fn get_use_stratagem() -> bool {
...
+        Err(e) => panic!("{}", e),
+        Ok(num) => Some(num),
+    }?;
+
+    let num = num
+        .parse::<u32>()
+        .unwrap_or_else(|_| panic!("Could not parse SFA_ENDPOINTS={} to a u32", num));
+
+    let xs = (1..=num).fold(vec![], |mut acc, x| {
+        let xs: Vec<_> = get_var(&format!("SFA_ENDPOINTS{}", x))
```
ip1981, line 236: I wonder if we could parse any env vars starting with `SFA_ENDPOINT`, for example: `SFA_ENDPOINT1`, `SFA_ENDPOINT_foo`, etc.

iml-orm/src/sfa/mod.rs
```diff
@@ -0,0 +1,491 @@
...
+    fn build(row: Self::Row) -> Self {
+        i16::build(row).into()
+    }
+}
+#[derive(
+    serde::Serialize, serde::Deserialize, Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd,
+)]
+#[cfg_attr(feature = "postgres-interop", derive(AsExpression))]
+#[cfg_attr(feature = "postgres-interop", sql_type = "SmallInt")]
+pub enum JobState {
```
ip1981, line 216: Here and in `JobType` would it make sense to omit the prefixes (`State` and `Type`)?

jgrund: I want these to be easily referenceable in the SFA API docs. There they refer to each enum for this property with the JOB_TYPE_* prefix.

iml-manager-env/src/lib.rs
jgrund: I'm not sure what you mean, this will only parse endpoints in the range from 1 to num.

ip1981: We have to set at least two variables `SFA_ENDPOINTS` and `SFA_ENDPOINTSXXX`, right?

I am talking about a less error-prone design: keep only `SFA_ENDPOINTXXX`, just find them all by prefix. As a side effect we could have any format for `SFA_ENDPOINTXXX`: `SFA_ENDPOINTS_foo`, `SFA_ENDPOINTS_f1_qwerty`, etc.

jgrund: Ah. 

I suppose we could iterate all the environment variables and pick out the ones that match

`SFA_ENDPOINTS<NUM>`, however I don't know if that's less error prone. 

We're trying to simulate an array with environment variables here which has a length and values for each index.

ip1981: It would still be an array, we could even sort it :) It would make trivial to remove some URLs, etc.

iml-orm/src/sfa/mod.rs
jgrund: On second thought, I'll remove it. The enum identifier will be there in most cases.

iml-manager-env/src/lib.rs
jgrund: Sounds reasonable. I'll take a look at what's needed to change.

iml-manager-env/src/lib.rs
```diff
@@ -218,6 +220,32 @@ pub fn get_use_stratagem() -> bool {
     string_to_bool(get_var("USE_STRATAGEM"))
 }
 
+pub fn get_sfa_endpoints() -> Option<Vec<Vec<Url>>> {
+    let xs: BTreeMap<_, _> = env::vars()
+        .filter(|(k, _)| k.starts_with("SFA_ENDPOINTS_"))
```
ip1981, line 225: Isn't plural form confusing?

iml-sfa.service
```diff
@@ -0,0 +1,17 @@
...
+Description=IML SFA Service
+PartOf=iml-manager.target
+After=postgresql-9.6.service
+After=iml-settings-populator.service
+Requires=iml-settings-populator.service
+
+
+[Service]
+Type=simple
+Environment=RUST_LOG=info
```
johnsonw, line 11: I think we recently moved the RUST_LOG environment variable to the iml-settings.conf didn't we?

docker/docker-compose.yml
```diff
@@ -127,6 +127,22 @@ services:
...
+    deploy:
+      restart_policy:
+        condition: on-failure
+        delay: 5s
+        window: 5s
+    volumes:
+      - "manager-config:/var/lib/chroma"
+    environment:
+      - RUST_LOG=info
+      - SFA_ENDPOINTS
```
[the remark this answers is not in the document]
jgrund: This is gone in the latest build.

iml-manager-env/src/lib.rs
```diff
@@ -218,6 +219,40 @@ pub fn get_use_stratagem() -> bool {
...
+    let num = env::var("SFA_ENDPOINTS");
+
+    let num = match num {
+        Err(env::VarError::NotPresent) => None,
+        Err(e) => panic!("{}", e),
+        Ok(num) => Some(num),
+    }?;
+
+    let num = num
+        .parse::<u32>()
```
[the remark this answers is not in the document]
jgrund: This is gone in the latest build.

iml-sfa.service
johnsonw: actually, I was thinking of the overrides file. nvm.

jgrund: Not on the manager.

PR #1824, opened 2020-04-24
Task Queue: rename Action queue to task queue

utopiabound:
Update to new design: https://wiki.whamcloud.com/display/HYD/Task+Queue

Generalize name away from mailboxes.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1824)


chroma_core/migrations/0016_task.py
```diff
@@ -47,22 +39,28 @@ class Migration(migrations.Migration):
...
                     "filesystem",
                     models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to="chroma_core.ManagedFilesystem"),
                 ),
+                (
+                    "running_on",
+                    models.ForeignKey(
+                        blank=True,
+                        null=True,
+                        on_delete=django.db.models.deletion.SET_NULL,
+                        to="chroma_core.LustreClientMount",
```
jgrund, line 56: This should probably be a `ManagedHost`. If we delegate all actions to just be action plugins, then it's the actions plugins responsibility to efficiently schedule work.

chroma_core/models/task.py
```diff
@@ -43,30 +43,24 @@ class Meta:
     fids_failed = models.BigIntegerField(default=0)
     data_transfered = models.BigIntegerField(default=0)
 
+    single_runner = models.BooleanField(default=False, null=False)
     keep_failed = models.BooleanField(default=True, null=False)
-    # Actually links to ActionType
-    actions = ArrayField(models.PositiveIntegerField())
+    # this is an array of text fields because an array of Char(16) causes problems for
+    # diesel on the rust side
```
jgrund, line 49: What sort of problems does this cause?

chroma_api/task.py
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2020 DDN. All rights reserved.
```
jgrund, line 1: We should not build any more api endpoints in Python.

Reasons:

- We are using a lib that is loosely maintained and is dependant on runtime RPMs to function.
- These RPMs are subject to change.
- The Python API has significantly worse performance characteristics.
- We have an existing Rust service that can host API endpoints, and communicate with the job_scheduler / db.

At this point, we should consider the Python API legacy, and move to replace it instead of building on top of it.

chroma_core/models/task.py
utopiabound: Diesel complains about array's of chars when creating schema.rs

jgrund:
Python unit tests are failing

chroma-manager.conf.template
```diff
@@ -192,6 +192,17 @@ server {
         gzip_types application/json;
     }
 
+    location /api/task {
+        proxy_set_header Host $http_host;
```
jgrund, line 196:
```suggestion
    auth_request /auth;
```

Do we want auth on this endpoint? Seems like not everyone should be able to read / submit tasks.

chroma_core/models/task.py
```diff
@@ -0,0 +1,136 @@
...
+    class Meta:
+        app_label = "chroma_core"
+        unique_together = ("name",)
+
+
+class CreateTaskJob(AdvertisedJob):
+    task = models.ForeignKey("Task", on_delete=CASCADE)
+    requires_confirmation = False
+    classes = ["Task"]
+    verbe = "Create"
```
jgrund, line 63:
```suggestion
    verb = "Create"
```

chroma_core/models/task.py
```diff
@@ -0,0 +1,136 @@
...
+class CreateTaskStep(Step):
+    def run(self, args):
+        self.invoke_rust_agent_expect_result(args["host"], "postoffice_add", args["task"])
+
+
+class RemoveTaskJob(AdvertisedJob):
+    task = models.ForeignKey("Task", on_delete=CASCADE)
+    requires_confirmation = False
+    classes = ["Task"]
+    verbe = "Remove"
```
jgrund, line 97:
```suggestion
    verb = "Remove"
```

chroma_core/models/task.py
```diff
@@ -0,0 +1,136 @@
...
+    fids_failed = models.BigIntegerField(default=0)
+    data_transfered = models.BigIntegerField(default=0)
+
+    single_runner = models.BooleanField(default=False, null=False)
+    keep_failed = models.BooleanField(default=True, null=False)
+    # this is an array of text fields because an array of Char(16) causes problems for
+    # diesel on the rust side
+    actions = ArrayField(models.TextField())
+
+    running_on = models.ForeignKey("ManagedHost", blank=True, null=True, on_delete=SET_NULL)
```
jgrund, line 50: Does this mean a task can only be running on a single host?

rust-iml.spec
```diff
@@ -93,6 +93,7 @@ Obsoletes: iml-device-scanner-proxy
 
 %post agent
 systemctl preset rust-iml-agent.path
+%tmpfiles_create %{_tmpfilesdir}/iml-agent.conf
```
jgrund, line 96: I haven't looked at what current in CentOS 7.7, but looks like this is deprecated upstream at least:

https://github.com/systemd/systemd/blob/master/src/core/macros.systemd.in#L92-L96

iml-orm/src/fidtaskqueue.rs
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+pub use crate::models::ChromaCoreFidtaskqueue;
+use crate::{lustrefid::LustreFid, schema::chroma_core_fidtaskqueue, task::ChromaCoreTask, DbPool};
+//use diesel::{dsl, prelude::*};
```
jgrund, line 9: dead code.

iml-orm/src/fidtaskqueue.rs
```diff
@@ -0,0 +1,40 @@
...
+use serde_json;
+use tokio_diesel::AsyncRunQueryDsl as _;
+
+impl ChromaCoreFidtaskqueue {}
```
jgrund, line 13: dead code.

iml-api/src/task.rs
```diff
@@ -0,0 +1,68 @@
...
+use iml_orm::{task::ChromaCoreTask, tokio_diesel::AsyncRunQueryDsl as _};
+use iml_rabbit::Connection;
+use iml_wire_types::{ApiList, CmdWrapper};
+use warp::Filter;
+
+async fn create_task(
+    client: Connection,
+    task: serde_json::Value,
+) -> Result<impl warp::Reply, warp::Rejection> {
+    let pool = iml_orm::pool().map_err(ImlApiError::ImlR2D2Error)?;
```
jgrund, line 16: Doing this per call defeats the purpose of using a pool. 

We should instead use a warp filter to wrap this and pass it into each call (similar to the client filter).

iml-api/src/task.rs
```diff
@@ -0,0 +1,68 @@
...
+    let command = get_command(&pool, xs[1]).await?;
+
+    Ok(warp::reply::json(&CmdWrapper { command }))
+}
+
+async fn remove_task(
+    client: Connection,
+    ids: Vec<i32>,
+) -> Result<impl warp::Reply, warp::Rejection> {
+    let pool = iml_orm::pool().map_err(ImlApiError::ImlR2D2Error)?;
```
jgrund, line 32: Doing this per call defeats the purpose of using a pool. 

We should instead use a warp filter to wrap this and pass it into each call (similar to the client filter).

iml-api/src/task.rs
```diff
@@ -0,0 +1,68 @@
...
+        .map_err(ImlApiError::ImlJobSchedulerRpcError)
+        .await?;
+
+    let command = get_command(&pool, xs[1]).await?;
+
+    Ok(warp::reply::json(&CmdWrapper { command }))
+}
+
+async fn get_tasks() -> Result<impl warp::Reply, warp::Rejection> {
+    let pool = iml_orm::pool().map_err(ImlApiError::ImlR2D2Error)?;
```
jgrund, line 45: Doing this per call defeats the purpose of using a pool. 

We should instead use a warp filter to wrap this and pass it into each call (similar to the client filter).

iml-api/src/command.rs
```diff
@@ -0,0 +1,48 @@
...
+use iml_wire_types::{Command, EndpointName as _, TestHostJob};
+use itertools::Itertools;
+
+pub(crate) async fn get_command(pool: &DbPool, id: i32) -> Result<Command, ImlApiError> {
+    let mut cmd: Vec<ChromaCoreCommand> = ChromaCoreCommand::by_id(id)
+        .get_results_async(&pool)
+        .map_err(ImlApiError::ImlDieselAsyncError)
+        .await?;
+
+    let cmd = cmd.pop().unwrap();
```
jgrund, line 23:
```suggestion
         let cmd: Option<ChromaCoreCommand> = ChromaCoreCommand::by_id(id)
            .first_async(&pool)
            .await
            .optional()
            .ok_or_else(|| ImlApiError::NoneError);
```

This requires bringing `tokio_diesel::OptionalExtension as _` into scope and creating a new error variant to handle conversion of `None` into an `ImlApiError`.

iml-api/src/command.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let jobs: Vec<ChromaCoreJob> = get_jobs_by_cmd(id, &pool)
+        .map_err(ImlApiError::ImlDieselAsyncError)
+        .await?;
```
jgrund, line 27:
```suggestion
let jobs: Vec<ChromaCoreJob> = get_jobs_by_cmd(id, &pool).await?;
```

iml-api/src/command.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let steps: Vec<ChromaCoreStepresult> = ChromaCoreStepresult::by_jobs(jobs.iter().map(|j| j.id))
+        .get_results_async(&pool)
+        .map_err(ImlApiError::ImlDieselAsyncError)
+        .await?;
```
jgrund, line 32:
```suggestion
    let steps: Vec<ChromaCoreStepresult> = ChromaCoreStepresult::by_jobs(jobs.iter().map(|j| j.id))
        .get_results_async(&pool)
        .await?;
```

rust-iml.spec
utopiabound: for our version, `%tmpfiles_create` is the only option available.

iml-api/src/command.rs
```diff
@@ -0,0 +1,43 @@
...
+        created_at: format!("{}", cmd.created_at),
+        errored: cmd.errored,
+        id: cmd.id as u32,
+        message: cmd.message,
+        resource_uri: format!("/api/{}/{}/", Command::endpoint_name(), cmd.id),
+        jobs: jobs
+            .iter()
+            .map(|j| format!("/api/{}/{}/", TestHostJob::endpoint_name(), j.id))
+            .collect(),
+        logs: steps.iter().map(|s| s.log.clone()).join("\n"),
```
jgrund, line 41:
```suggestion
        logs: steps.into_iter().map(|s| s.log).join("\n"),
```

This can potentially be expensive since the log strings can be quite large (HA failover for instance). `into_iter` takes ownership so we can avoid the clone.

iml-orm/src/fidtaskqueue.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright (c) 2020 DDN. All rights reserved.
```
johnsonw, line 1: copied twice?

iml-orm/src/fidtaskqueue.rs
```diff
@@ -0,0 +1,37 @@
...
+// Use of this source code is governed by a MIT-style
```
jgrund, line 2:
```suggestion
```

PR #1825, opened 2020-04-24
Use optional environmentfile for config

jgrund:
We have a few config vars that need to be overridden in a docker
environment.

We currently do this by updating the config from print-settings.py
by trying to read an override file

Instead, we can read in an environment file that will be passed into
docker stack and populated at startup.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1825)


johnsonw:
Looks good. Are we already generating the environment file that's being specified?

jgrund:
> Looks good. Are we already generating the environment file that's being specified?

This patch reuses the `/etc/iml-docker/setup/config` which is already in environmentfile format.

PR #1828, opened 2020-04-25
Fix fight between the tests over the SOCK_DIR var

### LINKED ISSUE #1826: Intermittent test failure: action_plugins::lpurge::lpurge_conf_tests::works
Rust unit testing has an intermittent failure:

https://github.com/whamcloud/integrated-manager-for-lustre/runs/617943042

Screen Shot 2020-04-25 at 9 37 46 AM

ip1981:
Closes #1826.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1828)

### DIFF
--- a/iml-agent/src/daemon_plugins/postoffice.rs
+++ b/iml-agent/src/daemon_plugins/postoffice.rs
@@ -44,7 +44,7 @@ pub fn create() -> impl DaemonPlugin {
 
 /// Return socket address for a given mailbox
 pub fn socket_name(mailbox: &str) -> String {
-    let sock_dir = env::get_var("SOCK_DIR");
+    let sock_dir = env::get_var_else("SOCK_DIR", "/run/iml");
     format!("{}/postman-{}.sock", sock_dir, mailbox)
 }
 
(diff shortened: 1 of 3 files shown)


iml-agent/src/daemon_plugins/postoffice.rs
```diff
@@ -44,7 +44,7 @@ pub fn create() -> impl DaemonPlugin {
 
 /// Return socket address for a given mailbox
 pub fn socket_name(mailbox: &str) -> String {
-    let sock_dir = env::get_var("SOCK_DIR");
+    let sock_dir = env::get_var_else("SOCK_DIR", "/run/iml");
```
jgrund, line 47: Hardcoding a default doesn't seem like the right fix. We should separate the code under test from the configuration.

ip1981: I see two separate issues here:

1. The service shouldn't crash without an environment variable defined (i. e. there must be a default value, if possible).
2. Your point :) The tests should depend on the environment at all.

Let me take another look.

PR #1831, opened 2020-04-28
Add device information to integration tests

johnsonw:
Add device information loaded into IML. This will allow the tests to
verify that the devices have been loaded before attempting to add
servers.

### DIFF
--- a/iml-system-test-utils/src/vagrant.rs
+++ b/iml-system-test-utils/src/vagrant.rs
@@ -8,6 +8,7 @@ use crate::{
 };
 use futures::future::try_join_all;
 use iml_cmd::{CheckedCommandExt, CmdError};
+use iml_wire_types::Volume;
 use std::{
     collections::{BTreeSet, HashMap},
     env, str,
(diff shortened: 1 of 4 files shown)


iml-system-test-utils/src/vagrant.rs
```diff
@@ -320,13 +321,57 @@ pub async fn setup_iml_install(
...
+        delay_for(Duration::from_secs(5)).await;
+
+        let iml_devices: BTreeSet<String> = get_iml_devices(config).await?;
+        all_volumes_accounted_for = wwids.is_subset(&iml_devices);
+        count += 1;
+
+        println!("Comparing iml devices to bindings files.");
+        println!("iml_devices: {:?}", iml_devices);
+        println!("binding wwids: {:?}", wwids);
+    }
```
ip1981, line 361: I am just pondering: :)
```rust
loop {
  count += 1;

  let iml_devices: BTreeSet<String> = get_iml_devices(config).await?;
  let all_volumes_accounted_for = wwids.is_subset(&iml_devices);
  
  println!("Comparing iml devices to bindings files ({})", &count);
  println!("iml_devices: {:?}", iml_devices);
  println!("binding wwids: {:?}", wwids);

  if all_volumes_accounted_for {
    return Ok();
  }
  if count >= max_tries {
    return Err(..);
  }
  delay_for(Duration::from_secs(5)).await;
}
```

johnsonw: Yeah this could probably be updated. Looks a bit cleaner.

PR #1846, opened 2020-04-30
Use smaller locks around daemon plugins

### LINKED ISSUE #1458: DaemonPlugin destroy should be async
Since we should be using parking-lot with async code (it blocks the whole task), we need to update `DaemonPlugin.destroy` to be async as well.

### LINKED ISSUE #1479: Remove all usages of parking lot from iml-agent

jgrund:
The IML agent architecture currently uses a large blocking RWLock around
all the active sessions.

Since sessions are largely independent of each other, we can instead use
many smaller async RWLocks around each session.

This approach uses interior mutability a bit better and makes easier to remove parking-lot (which blocks an entire task when it takes a lock), and is now removed in this patch.

Fixes #1479.
Fixes #1458.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1846)


jgrund:
___
*[iml-agent/src/http_comms/session.rs, line 85 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1846#-M6FOJ5x6rlkT_VcB71h:-M6FQlCl9VGEBD3Sxyaa:by9pzqx) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/0fb6ea2d0c765c4bf02303005fd18052a9b09546/iml-agent/src/http_comms/session.rs#L85)):*

I'm not sure I follow - do you refer to `mem::replace` as interior mutability? I expected to see one of `*Cell` types and I see none.
</blockquote></details>

https://ricardomartins.cc/2016/06/25/interior-mutability-thread-safety

PR #1847, opened 2020-04-30
Make settings populator call atomic

johnsonw:
The iml-settings file should be generated in a way that is atomic.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1847)

### DIFF
--- a/iml-settings-populator.service
+++ b/iml-settings-populator.service
@@ -5,5 +5,7 @@ After=postgresql-9.6.service
 
 [Service]
 WorkingDirectory=/usr/share/chroma-manager
-ExecStart=/bin/bash -c "exec /bin/python ./manage.py print-settings > /var/lib/chroma/iml-settings.conf"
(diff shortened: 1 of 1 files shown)


iml-settings-populator.service
```diff
@@ -5,5 +5,7 @@ After=postgresql-9.6.service
 
 [Service]
 WorkingDirectory=/usr/share/chroma-manager
-ExecStart=/bin/bash -c "exec /bin/python ./manage.py print-settings > /var/lib/chroma/iml-settings.conf"
+ExecStart=/bin/bash -c "exec /bin/python ./manage.py print-settings > /tmp/temp-settings.conf"
+ExecStart=mv /tmp/temp-settings.conf /var/lib/chroma/iml-settings.conf
+ExecStart=rm /tmp/temp-settings.conf
```
utopiabound, line 10: This last line shouldn't be needed.  And without an `-f` it will emit an error if the file is missing.

johnsonw: oh you're right, it's being moved, no copied. I'll remove.

PR #1849, opened 2020-05-01
Add force-remove hosts to IML CLI

jgrund:
Add the ability to force-remove hosts via the IML CLI

[asciicast](https://asciinema.org/a/Ezc0CWqYpSY6XqoNwrUUaqs5V)

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1849)

### DIFF
--- a/iml-manager-cli/src/server.rs
+++ b/iml-manager-cli/src/server.rs
@@ -3,7 +3,9 @@
 // license that can be found in the LICENSE file.
 
 use crate::{
-    api_utils::{get, get_all, get_hosts, post, put, wait_for_cmds, wait_for_cmds_success},
+    api_utils::{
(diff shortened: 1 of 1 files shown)


ip1981:
I would expect an option (e. g. `--force`) and, in any case, small update to the removal code, but it looks like a lot of code duplication between the remove and force-remove commands.

jgrund:
> I would expect an option (e. g. --force) and, in any case, small update to the removal code, but it looks like a lot of code duplication between the remove and force-remove commands.

force remove isn't a variant of remove; it's an entirely different action. 

There will be more actions in the future as well and they won't make sense as flags on an existing command.

In terms of DRYness, I'm not cleaning this up yet with the thought that we'll need a generic way to run any action available on a resource (as we can in the GUI). Without that part implemented, reducing duplication here is a premature optimization.

PR #1868, opened 2020-05-05
Use mock to perform local builds

jgrund:
Instead of building in the "global" namespace on the adm node,
we should build in a chroot to ensure that builds are reproducible
in a clean room env, do not require root, and do not pollute the adm
node.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1868)


jgrund:
Landed as d9e6df37ed73c4f3095148408f4723f3aaf4a67d

PR #1869, opened 2020-05-06
Fill device state

jgrund:
The device plugin emits a value upon connecting, but does not fill the
internal state with this value.

Because of this, data will always be sent for the first two poll
intervals, even if nothing has changed.

Always fill the state, even when resolving the initial future.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1869)


mkpankov:
I'm still seeing 2 initial updates, did you observe that as well?

mkpankov:
What I was doing specifically: run `import-pools`, restart the `iml-manager`. Then there are two updates on `iml-device` side. This happens every restart of the manager.

jgrund:
> I'm still seeing 2 initial updates, did you observe that as well?

Good catch, I should start the state as Send instead of pending.

iml-agent/src/daemon_plugins/device.rs
```diff
@@ -31,17 +30,23 @@ fn device_stream() -> impl Stream<Item = Result<String, ImlAgentError>> {
...
+enum State {
+    Pending,
+    Sent,
+}
+
 pub fn create() -> impl DaemonPlugin {
     Devices {
         trigger: None,
-        state: Arc::new(Mutex::new((None, None))),
+        state: Arc::new(Mutex::new((None, State::Pending))),
```
jgrund, line 42:
```suggestion
        state: Arc::new(Mutex::new((None, State:: Sent))),
```

jgrund:
> While it fixed the issue, it looks counter-intuitive to start in Sent state. Maybe call it Ready or something?

Yeah, it is a little counter-intuitive.

State is currently:

```rust
#[derive(Eq, PartialEq)]
enum State {
    Pending,
    Sent,
}
```

`Sent` basically means the value has been sent. This will be the case if the start_session method succeeds. 

`Pending` means there is a new value to be sent by `update_session`. 

If the `start_session` method fails, then we never call `update_session` any way, so it doesn't make sense to have another state here.

PR #1883, opened 2020-05-08
Add optional overrides file to services

johnsonw:
Adding an optional overrides file to the services will allow us to
override service environment variables easily.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1883)

### DIFF
--- a/iml-action-runner.service
+++ b/iml-action-runner.service
@@ -13,6 +13,7 @@ After=iml-action-runner.socket
 Type=simple
 Environment=RUST_LOG=info
 EnvironmentFile=/var/lib/chroma/iml-settings.conf
+EnvironmentFile=-/var/lib/chroma/overrides.conf
 ExecStart=/bin/iml-action-runner
 Restart=always
 RestartSec=2
--- a/iml-agent-comms.service
+++ b/iml-agent-comms.service
@@ -10,6 +10,7 @@ Requires=iml-settings-populator.service
 Type=simple
 Environment=RUST_LOG=info
 EnvironmentFile=/var/lib/chroma/iml-settings.conf
+EnvironmentFile=-/var/lib/chroma/overrides.conf
 ExecStart=/bin/iml-agent-comms
 Restart=always
 RestartSec=2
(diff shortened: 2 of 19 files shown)


jgrund:
- `rust-iml-agent.service` needs an override as well?
- Should this patch also update settings.py to override the log_level?

johnsonw:
Sure, I can update settings.py as well. Are there any other agent services I am missing besides rust-iml-agent?

jgrund:
> Are there any other agent services I am missing besides rust-iml-agent?

the `chroma-agent.service` but that is in a separate repo.

johnsonw:
I will open a separate patch to address chroma-agent.repo.

johnsonw:
I did not add the overrides to the settings-populator or docker service files as it doesn't look like it would make a difference in either of those services. Does that sound right?

iml-action-runner.service
```diff
@@ -13,6 +13,7 @@ After=iml-action-runner.socket
 Type=simple
 Environment=RUST_LOG=info
```
ip1981, line 14: This takes priority over the files.

iml-agent-comms.service
```diff
@@ -10,6 +10,7 @@ Requires=iml-settings-populator.service
 Type=simple
 Environment=RUST_LOG=info
```
ip1981, line 11: I suggest removing it.

iml-action-runner.service
jgrund: It doesn't:

https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=

> Settings from these files override settings made with Environment=.

johnsonw: From https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=:
> Settings from these files override settings made with Environment=. If the same variable is > set twice from these files, the files will be read in the order they are specified and the later > setting will override the earlier setting.

ip1981: Hm, I wasn't able to override that in a file...

PR #1889, opened 2020-05-11
Add `Executable` trait to iml-orm

jgrund:
Add a new trait `Executable` that can be used as a return for Insert (Upsert)
statements.

This trait allows the execution of the insert to be decoupled from the
actual statement, which means the statements can be reused in both
transactional / non-transactional contexts.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1889)

### DIFF
--- a/iml-manager-cli/src/profile.rs
+++ b/iml-manager-cli/src/profile.rs
@@ -80,14 +80,24 @@ pub async fn cmd(cmd: Option<Cmd>) -> Result<(), ImlManagerCliError> {
                 .into());
             }
 
-            profile::upsert_user_profile(profile, &pool).await?;
+            let (e1, e2, e3) = profile::upsert_user_profile(profile);
+
+            e1.execute_async(&pool).await?;
+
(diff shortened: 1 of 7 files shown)


iml-manager-cli/src/profile.rs
```diff
@@ -80,14 +80,24 @@ pub async fn cmd(cmd: Option<Cmd>) -> Result<(), ImlManagerCliError> {
...
 
             display_success("Profile loaded");
         }
         Some(Cmd::Remove { name }) => {
             let pool = iml_orm::pool()?;
 
-            profile::remove_profile_by_name(name, &pool).await?;
+            let (e1, e2, e3) = profile::remove_profile_by_name(name);
+
+            e1.execute_async(&pool).await?;
```
johnsonw, line 96: These can't be removed in parallel?

jgrund: No, there is a FK constraint between them.

PR #1901, opened 2020-05-15
Add Integration test debugging

johnsonw:
Add integration test debugging

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1901)

### DIFF
--- a/iml-system-test-utils/src/vagrant.rs
+++ b/iml-system-test-utils/src/vagrant.rs
@@ -3,15 +3,14 @@
 // license that can be found in the LICENSE file.
 
 use crate::{
-    iml, ssh, try_command_n_times, ServerList as _, SetupConfig, SetupConfigType,
-    STRATAGEM_CLIENT_PROFILE, STRATAGEM_SERVER_PROFILE,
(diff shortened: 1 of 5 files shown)


iml-system-test-utils/src/vagrant.rs
```diff
@@ -410,21 +439,74 @@ pub async fn add_docker_servers<S: std::hash::BuildHasher>(
...
+                STRATAGEM_CLIENT_PROFILE,
+            )
+            .as_str(),
+        )
+        .await?;
+
+        run_vm_command(
+            cluster_config.manager,
+            format!(
+                "sudo chroma-config profile register {}/stratagem-server.profile \
```
ip1981, line 482: We can use `iml server profile load` now. It would read from stdin and I was thinking that in other places we could transmit from the host to VM via pipes, like `echo ... | ssh 'cat > /destination'`.

johnsonw: I'll update to use `iml server profile load`

PR #1904, opened 2020-05-15
Extract the pool filter to iml-orm

jgrund:
Extract the warp pool filter to the iml-orm crate and put behind a
feature flag.

This way, we can use the filter in other crates that need a db pool
connection per request.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1904)

### DIFF
--- a/iml-orm/src/lib.rs
+++ b/iml-orm/src/lib.rs
@@ -42,6 +42,14 @@ pub mod step;
 #[cfg(feature = "postgres-interop")]
 pub mod task;
 
+use futures::channel::oneshot;
+use thiserror::Error;
+
+#[cfg(feature = "warp-filters")]
+use futures::{channel::mpsc, future, Future, StreamExt, TryFutureExt};
+#[cfg(feature = "warp-filters")]
+use warp::Filter;
+
 #[cfg(feature = "postgres-interop")]
 use diesel::{
     prelude::RunQueryDsl,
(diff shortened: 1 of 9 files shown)


iml-orm/src/lib.rs
```diff
@@ -75,10 +83,71 @@ pub use tokio_diesel;
...
 pub use r2d2;
 
+#[derive(Debug, Error)]
+pub enum ImlOrmError {
+    #[error(transparent)]
+    R2D2Error(#[from] r2d2::Error),
+    #[error(transparent)]
+    OneshotCanceled(#[from] oneshot::Canceled),
+    #[error(transparent)]
+    AsyncError(#[from] tokio_diesel::AsyncError),
```
nlinker, line 93: Does `transparent` mean that there are generated `From` instances included?

iml-orm/src/lib.rs
```diff
@@ -75,10 +83,71 @@ pub use tokio_diesel;
...
+
+#[cfg(feature = "warp-filters")]
+pub fn get_cloned_pool(
+    pool: DbPool,
+) -> (mpsc::UnboundedSender<PoolSender>, impl Future<Output = ()>) {
+    let (tx, rx) = mpsc::unbounded();
+
+    let fut = rx.for_each(move |sender: PoolSender| {
+        let _ = sender
+            .send(pool.clone())
```
nlinker, line 120: Does the pool get cloned on each request?
Isn't this too expensive?

iml-orm/src/lib.rs
jgrund: The from attribute is what generates it.

iml-orm/src/lib.rs
jgrund: The type for pool is: `pub struct Pool<M>(Arc<SharedPool<M>>)`

`.clone()` is incrementing the refcount of the `Arc`, so this is very cheap.

PR #1909, opened 2020-05-18
Update how vm's are destroyed

### LINKED ISSUE #1907: Update how vm's are destroyed in integration tests
During the cleanup step the integration tests destroy the virtual machines by simply running:
```
vagrant destroy -f
```

The nodes are destroyed as follows:
```
[root@onyx-25 vagrant]# vagrant destroy -f
==> c8: VM not created. Moving on...
==> c7: VM not created. Moving on...
==> c6: VM not created. Moving on...
==> c5: VM not created. Moving on...
==> c4: VM not created. Moving on...
==> c3: VM not created. Moving on...
==> c2: VM not created. Moving on...
==> c1: VM not created. Moving on...
==> oss2: VM not created. Moving on...
==> oss1: VM not created. Moving on...
==> mds2: VM not created. Moving on...
==> mds1: VM not created. Moving on...
==> adm: Forcing shutdown of VM...
==> adm: Destroying VM and associated drives...
==> iscsi: Forcing shutdown of VM...
==> iscsi: Destroying VM and associated drives...
```

Instead of doing a `vagrant destroy -f` I would like to instead halt all hosts and then destroy each host one by one (not using the force flag), starting with the client node(s), server nodes, adm, and finally iscsi.

johnsonw:
Fixes #1907.

Halt all nodes and destroy one by one to prevent resource from hanging
on iscsi.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1909)

### DIFF
--- a/iml-system-test-utils/src/vagrant.rs
+++ b/iml-system-test-utils/src/vagrant.rs
@@ -59,10 +59,10 @@ pub async fn destroy<'a>(config: &ClusterConfig) -> Result<(), CmdError> {
     nodes.reverse();
 
     for node in &nodes {
-        let mut halt_cmd = halt().await?;
-        halt_cmd.arg(node);
+        let mut suspend_cmd = suspend().await?;
+        suspend_cmd.arg(node);
 
-        try_command_n_times(3, 1, &mut halt_cmd).await?;
+        try_command_n_times(3, 1, &mut suspend_cmd).await?;
     }
 
     for node in &nodes {
@@ -78,6 +78,13 @@ pub async fn destroy<'a>(config: &ClusterConfig) -> Result<(), CmdError> {
 pub async fn halt() -> Result<Command, CmdError> {
     let mut x = vagrant().await?;
     x.arg("halt");
+
+    Ok(x)
+}
+
+pub async fn suspend() -> Result<Command, CmdError> {
+    let mut x = vagrant().await?;
+    x.arg("suspend");
 
     Ok(x)
 }
--- a/iml-system-test-utils/src/vagrant.rs
+++ b/iml-system-test-utils/src/vagrant.rs
@@ -55,8 +55,7 @@ pub async fn up<'a>() -> Result<Command, CmdError> {
 }
 
 pub async fn destroy<'a>(config: &ClusterConfig) -> Result<(), CmdError> {
-    let mut nodes = config.all();
-    nodes.reverse();
+    let nodes = config.destroy_list();
 
     for node in &nodes {
         let mut suspend_cmd = suspend().await?;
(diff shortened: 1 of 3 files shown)


iml-system-test-utils/src/vagrant.rs
```diff
@@ -58,12 +58,25 @@ pub async fn up<'a>() -> Result<Command, CmdError> {
     Ok(x)
 }
 
-pub async fn destroy<'a>() -> Result<(), CmdError> {
-    let mut x = vagrant().await?;
+pub async fn destroy<'a>(config: &ClusterConfig) -> Result<(), CmdError> {
+    let mut nodes = config.all();
+    nodes.reverse();
+
+    for node in &nodes {
```
jgrund, line 65: What's the benefit to halting in addition to destroying?

iml-system-test-utils/src/vagrant.rs
```diff
@@ -58,12 +58,25 @@ pub async fn up<'a>() -> Result<Command, CmdError> {
     Ok(x)
 }
 
-pub async fn destroy<'a>() -> Result<(), CmdError> {
-    let mut x = vagrant().await?;
+pub async fn destroy<'a>(config: &ClusterConfig) -> Result<(), CmdError> {
+    let mut nodes = config.all();
+    nodes.reverse();
```
jgrund, line 63: This seems brittle, we are making the assumption that reversing the config will always present nodes in an order suitable for shutdown.

Perhaps it would be better to make this a method on the `ClusterConfig`. That way, the intent is clear and it has a better chance being noticed in future updates.

johnsonw: sure 👍

iml-system-test-utils/src/vagrant.rs
johnsonw: Will suspend instead of halting.

iml-system-test-utils/src/vagrant.rs
```diff
@@ -54,12 +54,24 @@ pub async fn up<'a>() -> Result<Command, CmdError> {
...
 }
 
-pub async fn destroy<'a>() -> Result<(), CmdError> {
-    let mut x = vagrant().await?;
+pub async fn destroy<'a>(config: &ClusterConfig) -> Result<(), CmdError> {
+    let nodes = config.destroy_list();
+
+    for node in &nodes {
+        let mut suspend_cmd = suspend().await?;
+        suspend_cmd.arg(node);
```
ip1981, line 62: So it looks like a network outage, right?
Something bad happen when a node is shutting down?

johnsonw: Basically we are running `vagrant destroy -f` which will attempt to destroy all nodes, starting with c8 -> c1, oss2, oss1, mds2,mds1,adm,iscsi. This process will attempt to shut down each node and then destroy it along with resources one by one. This new approach will loop through each node and suspend each node. By doing it this way, we ensure everything is suspended before we attempt to destroy. This will hopefully prevent resources from getting hung up.

PR #1910, opened 2020-05-19
Use Signed 32 bit integer type for ids

jgrund:
Postgres does not support unsigned numbers (besides the oid type).
As such, all ids in IML are actually signed 32 bit ints.

Update the representation of these types in wire-types and other spots to reflect this.

This will pave the way for using diesel struct types instead of
repeating types in both iml-orm and wire-types.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1910)

### DIFF
--- a/iml-gui/crate/src/components/command_modal.rs
+++ b/iml-gui/crate/src/components/command_modal.rs
@@ -26,23 +26,23 @@ const POLL_INTERVAL: Duration = Duration::from_millis(1000);
 
 type Job0 = Job<Option<serde_json::Value>>;
 
-type RichCommand = Rich<u32, Arc<Command>>;
-type RichJob = Rich<u32, Arc<Job0>>;
-type RichStep = Rich<u32, Arc<Step>>;
+type RichCommand = Rich<i32, Arc<Command>>;
+type RichJob = Rich<i32, Arc<Job0>>;
+type RichStep = Rich<i32, Arc<Step>>;
 
-type JobsGraph = DependencyDAG<u32, RichJob>;
-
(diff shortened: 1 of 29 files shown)


iml-gui/crate/src/components/command_modal.rs
```diff
@@ -551,21 +551,21 @@ where
...
+            // we always can url encode a vector of i32-s
             unreachable!("Cannot encode request for {} with params {:?}", T::endpoint_name(), ids)
         }
     }
 }
 
-fn extract_uri_id<T: EndpointName>(input: &str) -> Option<u32> {
+fn extract_uri_id<T: EndpointName>(input: &str) -> Option<i32> {
     lazy_static::lazy_static! {
         static ref RE: Regex = Regex::new(r"/api/(\w+)/(\d+)/").unwrap();
```
nlinker, line 562:
```suggestion
        static ref RE: Regex = Regex::new(r"/api/(\w+)/(-?\d+)/").unwrap();
```
The rationale is that since now the ids can be negative in principle, the urls can become like `/api/command/-12/`.
This however might be enforced somewhere else, so the flexibility might not be required.

jgrund: The invariant has been loosened, but in principle we won't accept negative numbers. I don't think we should change this to account for negatives.

jgrund: in other words, the ids should always be > 0.

nlinker: Ok, then.

jgrund:
Landed as 3a00eb46f46577f29129c4949b205a0600615914

PR #1911, opened 2020-05-19
Monitor client mounts in iml-device

### LINKED ISSUE #1800: Return both mounts and devices from device scanner
device scanner has two separate calls to retrieve data.

1. `Stream` - produces a tree of devices:

   https://github.com/whamcloud/device-scanner/blob/af84681ab8949ef0eee64cf2fd3b83d46f4fea5c/device-scanner-daemon/src/daemon.rs#L81-L89

1. `GetMounts` - produces a list of current mounts:

   https://github.com/whamcloud/device-scanner/blob/af84681ab8949ef0eee64cf2fd3b83d46f4fea5c/device-scanner-daemon/src/daemon.rs#L90-L100

Going forward, we need to be able to monitor client mounts via the device plugin. To do this, the `Stream` call should return a tuple of both the device tree and local mounts.

### LINKED ISSUE #1801: Update device plugin to monitor client mounts
Follow-on from #1800. 

The device plugin should monitor and update Lustre client mounts found on any running iml-agent instances.

https://github.com/whamcloud/integrated-manager-for-lustre/blob/f629610657db940f6001a36fef723850b37a287d/chroma_core/models/client_mount.py#L19

ip1981:
Example:

```
chroma=> select * from chroma_core_lustreclientmount;
 id |       state_modified_at       |  state  | immutable_state | not_deleted | mountpoint | content_type_id | filesystem | host_id
----+-------------------------------+---------+-----------------+-------------+------------+-----------------+---------------+---------
 16 | 2020-05-18 16:01:10.768484+00 | mounted | f               | t           | /mnt/fs    |              35 |             fs |       5
 17 | 2020-05-18 16:01:10.768484+00 | mounted | f               | t           | /mnt/fs2   |              35 |             fs2 |       5
```

```
may 19 10:18:29 adm.local iml-device[13087]: May 19 10:18:29.751 DEBUG iml_device: Client mounts at c1.local: [("fs2", Some("/mnt/fs2"))]
may 19 10:18:29 adm.local iml-device[13087]: May 19 10:18:29.756 DEBUG iml_device: Inserted ChromaCoreLustreclientmount { id: 17, state_modified_at: 2020-05-18T16:01:10.768484Z, state: "mounted", immutable_state: false, not_deleted: Some(true), mountpoint: Some("/mnt/fs2"), content_type_id: Some(35), filesystem_id: 2, host_id: 5 }
may 19 10:18:29 adm.local iml-device[13087]: May 19 10:18:29.759 DEBUG iml_device: Updated [ChromaCoreLustreclientmount { id: 16, state_modified_at: 2020-05-19T10:18:29.751826Z, state: "unmounted", immutable_state: false, not_deleted: Some(true), mountpoint: None, content_type_id: Some(35), filesystem_id: 1, host_id: 5 }]
```

```
chroma=> select * from chroma_core_lustreclientmount;
 id |       state_modified_at       |   state   | immutable_state | not_deleted | mountpoint | content_type_id | filesystem | host_id
----+-------------------------------+-----------+-----------------+-------------+------------+-----------------+---------------+---------
 17 | 2020-05-18 16:01:10.768484+00 | mounted   | f               | t           | /mnt/fs2   |              35 |             fs2 |       5
 16 | 2020-05-19 10:18:29.751826+00 | unmounted | f               | t           |            |              35 |             fs |       5
```

Closes #1800, #1801.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1911)

### DIFF
--- a/iml-services/iml-device/src/main.rs
+++ b/iml-services/iml-device/src/main.rs
@@ -2,9 +2,10 @@
 // Use of this source code is governed by a MIT-style
 // license that can be found in the LICENSE file.
 
-use device_types::devices::Device;
-use diesel::{self, pg::upsert::excluded, prelude::*};
+use device_types::{devices::Device, mount::Mount};
+use diesel::{self, dsl, pg::upsert::excluded, prelude::*, sql_types};
 use futures::{lock::Mutex, TryFutureExt, TryStreamExt};
+use im::HashSet;
 use iml_device::{
     linux_plugin_transforms::{
         build_device_lookup, devtree2linuxoutput, get_shared_pools, populate_zpool, update_vgs,
@@ -13,9 +14,12 @@ use iml_device::{
     ImlDeviceError,
 };
 use iml_orm::{
-    models::{ChromaCoreDevice, NewChromaCoreDevice},
-    schema::chroma_core_device::{devices, fqdn, table},
-    tokio_diesel::*,
+    models::{ChromaCoreDevice, ChromaCoreLustreclientmount, NewChromaCoreDevice},
+    schema::{
+        chroma_core_device as dvc, chroma_core_lustreclientmount as clmnt,
+        chroma_core_managedhost as host, django_content_type as djct,
+    },
+    tokio_diesel::{AsyncRunQueryDsl as _, OptionalExtension as _},
     DbPool,
 };
 use iml_service_queue::service_queue::consume_data;
@@ -29,7 +33,7 @@ use warp::Filter;
 type Cache = Arc<Mutex<HashMap<Fqdn, Device>>>;
 
 async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
-    let data: HashMap<Fqdn, Device> = table
+    let data: HashMap<Fqdn, Device> = dvc::table
         .load_async(&pool)
         .await?
         .into_iter()
@@ -43,6 +47,116 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
         .collect::<Result<_, _>>()?;
 
     Ok(Arc::new(Mutex::new(data)))
+}
+
+async fn update_devices(
+    pool: &DbPool,
+    host: &Fqdn,
+    devices: &Device,
+) -> Result<(), ImlDeviceError> {
+    let device_to_insert = NewChromaCoreDevice {
+        fqdn: host.to_string(),
+        devices: serde_json::to_value(devices)
+            .expect("Could not convert incoming Devices to JSON."),
+    };
+
+    let new_device = diesel::insert_into(dvc::table)
+        .values(device_to_insert)
+        .on_conflict(dvc::fqdn)
+        .do_update()
+        .set(dvc::devices.eq(excluded(dvc::devices)))
+        .get_result_async::<ChromaCoreDevice>(&pool)
+        .await?;
+    tracing::info!("Inserted devices from host '{}'", host);
+    tracing::debug!("Inserted {:?}", new_device);
+    Ok(())
+}
+
+async fn update_client_mounts(
+    pool: &DbPool,
+    ct_id: Option<i32>,
+    host: &Fqdn,
+    mounts: &HashSet<Mount>,
+) -> Result<(), ImlDeviceError> {
+    let host_id = match host::table
+        .select(host::id)
+        .filter(host::fqdn.eq(host.to_string()))
+        .get_result_async::<i32>(&pool)
+        .await
+        .optional()?
+    {
+        Some(id) => id,
+        None => {
+            tracing::warn!("Host '{}' is unknown", host);
+            return Ok(());
+        }
+    };
+
+    let lustre_mounts = mounts
+        .into_iter()
+        .filter(|m| m.fs_type.0 == "lustre")
+        .filter_map(|m| {
+            m.source
+                .0
+                .to_str()
+                .and_then(|p| p.splitn(2, ":/").nth(1))
+                .map(|fs| (fs.to_string(), m.target.0.to_str().map(String::from)))
+        })
+        .collect::<Vec<_>>();
+
+    tracing::debug!(
+        "Client mounts at {}({}): {:?}",
+        host,
+        host_id,
+        &lustre_mounts
+    );
+
+    let state_modified_at = chrono::offset::Utc::now().into_sql::<sql_types::Timestamptz>();
+
+    let xs: Vec<_> = lustre_mounts
+        .into_iter()
+        .map(|(fs, tg)| {
+            (
+                clmnt::host_id.eq(host_id),
+                clmnt::filesystem.eq(fs),
+                clmnt::mountpoint.eq(tg),
+                clmnt::state.eq("mounted"),
+                clmnt::state_modified_at.eq(state_modified_at),
+                clmnt::immutable_state.eq(false),
+                clmnt::not_deleted.eq(true),
+                clmnt::content_type_id.eq(ct_id),
+            )
+        })
+        .collect();
+
+    let existing_mounts = diesel::insert_into(clmnt::table)
+        .values(xs)
+        .on_conflict((clmnt::host_id, clmnt::filesystem, clmnt::not_deleted))
+        .do_update()
+        .set((
+            clmnt::mountpoint.eq(excluded(clmnt::mountpoint)),
+            clmnt::state.eq(excluded(clmnt::state)),
+            clmnt::state_modified_at.eq(state_modified_at),
+        ))
+        .get_results_async::<ChromaCoreLustreclientmount>(&pool)
+        .await?;
+
+    let fs_names: Vec<String> = existing_mounts.into_iter().map(|c| c.filesystem).collect();
+
+    let updated = diesel::update(clmnt::table)
+        .filter(clmnt::filesystem.ne(dsl::all(fs_names)))
+        .filter(clmnt::host_id.eq(host_id))
+        .set((
+            clmnt::mountpoint.eq(Option::<String>::None),
+            clmnt::state.eq("unmounted".as_sql::<sql_types::Text>()),
+            clmnt::state_modified_at.eq(state_modified_at),
+        ))
+        .get_results_async::<ChromaCoreLustreclientmount>(&pool)
+        .await?;
+
+    tracing::debug!("Updated client mounts: {:?}", updated);
+
+    Ok(())
 }
 
 #[tokio::main]
@@ -112,29 +226,22 @@ async fn main() -> Result<(), ImlDeviceError> {
 
     let ch = iml_rabbit::create_channel(&conn).await?;
 
-    let mut s = consume_data::<Device>(&ch, "rust_agent_device_rx");
+    let mut s = consume_data::<(Device, HashSet<Mount>)>(&ch, "rust_agent_device_rx");
 
-    while let Some((f, d)) = s.try_next().await? {
+    // Django's artifact:
+    let lustreclientmount_ct_id = djct::table
+        .select(djct::id)
+        .filter(djct::model.eq("lustreclientmount"))
+        .first_async::<i32>(&pool)
+        .await
+        .optional()?;
+
+    while let Some((host, (devices, mounts))) = s.try_next().await? {
         let mut cache = cache2.lock().await;
+        cache.insert(host.clone(), devices.clone());
 
-        cache.insert(f.clone(), d.clone());
-
-        let device_to_insert = NewChromaCoreDevice {
-            fqdn: f.to_string(),
-            devices: serde_json::to_value(d).expect("Could not convert incoming Devices to JSON."),
-        };
-
-        let new_device = diesel::insert_into(table)
-            .values(device_to_insert)
-            .on_conflict(fqdn)
-            .do_update()
-            .set(devices.eq(excluded(devices)))
-            .get_result_async::<ChromaCoreDevice>(&pool)
-            .await
-            .expect("Error saving new device");
-
-        tracing::info!("Inserted device from host {}", new_device.fqdn);
-        tracing::trace!("Inserted device {:?}", new_device);
+        update_devices(&pool, &host, &devices).await?;
+        update_client_mounts(&pool, lustreclientmount_ct_id, &host, &mounts).await?;
     }
 
     Ok(())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1349,9 +1349,11 @@ dependencies = [
 name = "iml-device"
 version = "0.3.0"
 dependencies = [
+ "chrono",
  "device-types",
  "diesel",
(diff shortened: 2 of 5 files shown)


iml-services/iml-device/src/main.rs
```diff
@@ -13,8 +14,12 @@ use iml_device::{
     ImlDeviceError,
 };
 use iml_orm::{
-    models::{ChromaCoreDevice, NewChromaCoreDevice},
-    schema::chroma_core_device::{devices, fqdn, table},
+    models::{ChromaCoreDevice, ChromaCoreLustreclientmount, NewChromaCoreDevice},
+    schema::{
+        chroma_core_device as dvc, chroma_core_lustreclientmount as clmnt,
+        chroma_core_managedfilesystem as mgtfs, chroma_core_managedhost as mgthost,
```
jgrund, line 20:
```suggestion
        chroma_core_managedfilesystem as fs, chroma_core_managedhost as host,
```

mgt is overloaded term in lustre world.

iml-services/iml-device/src/main.rs
```diff
@@ -86,31 +91,131 @@ async fn main() -> Result<(), ImlDeviceError> {
...
+
+        let insert_mounts = lustre_mounts
+            .iter()
+            .map(|(fs, tg)| {
+                let mp = tg
+                    .clone()
+                    .into_sql::<sql_types::Nullable<sql_types::Text>>();
+                mgtfs::table
+                    .inner_join(mgthost::table.on(cross))
+                    .inner_join(djct::table.on(cross))
```
jgrund, line 148:
```suggestion
                    .inner_join(djct)
```

There is a `joinable!` implemented for this in the schema.

iml-services/iml-device/src/main.rs
```diff
@@ -86,31 +91,131 @@ async fn main() -> Result<(), ImlDeviceError> {
...
+        let immutable_state = false.into_sql::<sql_types::Bool>();
+
+        let insert_mounts = lustre_mounts
+            .iter()
+            .map(|(fs, tg)| {
+                let mp = tg
+                    .clone()
+                    .into_sql::<sql_types::Nullable<sql_types::Text>>();
+                mgtfs::table
+                    .inner_join(mgthost::table.on(cross))
```
jgrund, line 147: This cross join is strange given there is no fk constraint between filesystems and hosts.

ip1981:
> you should select over the lustre client mount table

How can I query an empty table?

jgrund:
> How can I query an empty table?

Possibly empty, and you'd be doing an upsert.

iml-services/iml-device/src/main.rs
ip1981: That's the point of cross join. Actually since FS and host names are unique, I only get one record from each table.

jgrund: `chroma_core_lustreclientmount` has all the fk constraints in place already, which means there is no need for a cross join.

```
chroma=> \d chroma_core_lustreclientmount;
                                       Table "public.chroma_core_lustreclientmount"
      Column       |           Type           |                                 Modifiers
-------------------+--------------------------+----------------------------------------------------------------------------
 id                | integer                  | not null default nextval('chroma_core_lustreclientmount_id_seq'::regclass)
 state_modified_at | timestamp with time zone | not null
 state             | character varying(32)    | not null
 immutable_state   | boolean                  | not null
 not_deleted       | boolean                  |
 mountpoint        | character varying(1024)  |
 content_type_id   | integer                  |
 filesystem_id     | integer                  | not null
 host_id           | integer                  | not null
Indexes:
    "chroma_core_lustreclientmount_pkey" PRIMARY KEY, btree (id)
    "chroma_core_lustreclient_host_id_filesystem_id_no_b6406a2e_uniq" UNIQUE CONSTRAINT, btree (host_id, filesystem_id, not_deleted)
    "chroma_core_lustreclientmount_content_type_id_2ed6c185" btree (content_type_id)
    "chroma_core_lustreclientmount_filesystem_id_ce4463ef" btree (filesystem_id)
    "chroma_core_lustreclientmount_host_id_e5d16643" btree (host_id)
Foreign-key constraints:
    "chroma_core_lustrecl_content_type_id_2ed6c185_fk_django_co" FOREIGN KEY (content_type_id) REFERENCES django_content_type(id) DEFERRABLE INITIALLY DEFERRED
    "chroma_core_lustrecl_filesystem_id_ce4463ef_fk_chroma_co" FOREIGN KEY (filesystem_id) REFERENCES chroma_core_managedfilesystem(id) DEFERRABLE INITIALLY DEFERRED
    "chroma_core_lustrecl_host_id_e5d16643_fk_chroma_co" FOREIGN KEY (host_id) REFERENCES chroma_core_managedhost(id) DEFERRABLE INITIALLY DEFERRED
```

ip1981:
I need filesystem id and host id, which I can't know without without querying other tables first, can I?

ip1981:
It is still possible in this case to do upsert in a batch, but in raw SQL :)

jgrund:
> It is still possible in this case to do upsert in a batch, but in raw SQL :)

I think it's possible within the query builder as well

jgrund:
Here's a high-level way to do the batch upsert on mounted devices (I checked that this compiles):

```rust
// This can be done once at startup, doesn't need to be inside the loop.
let content_type_id = djct::table
            .select(djct::id)
            .filter(djct::model.eq("lustreclientmount"))
            .first_async::<i32>(&pool)
            .await
            .optional()?;

        let host_id: Option<i32> = mgthost::table
            .select(mgthost::id)
            .filter(mgthost::fqdn.eq(host.to_string()))
            .first_async::<i32>(&pool)
            .await
            .optional()?;

        let host_id = match host_id {
            Some(x) => x,
            None => continue,
        };

        let xs: Vec<_> = lustre_mounts
            .into_iter()
            .map(|(fs, tg)| {
                (
                    clmnt::host_id.eq(host_id),
                    clmnt::filesystem_id.eq(1), // hardcoded because this needs to be filesystem_name instead
                    clmnt::mountpoint.eq(tg),
                    clmnt::state.eq("mounted"),
                    clmnt::state_modified_at.eq(chrono::offset::Utc::now()),
                    clmnt::immutable_state.eq(false),
                    clmnt::not_deleted.eq(true),
                    clmnt::content_type_id.eq(content_type_id),
                )
            })
            .collect();
```

I didn't implement upsert for unmounted clients, but it should be mostly similar to this.

iml-services/iml-device/src/main.rs
ip1981: Maybe it's better to show what I mean, technically, I select from a cross join:

```sql
INSERT INTO chroma_core_lustreclientmount
            (state_modified_at,
             state,
             immutable_state,
             not_deleted,
             mountpoint,
             content_type_id,
             filesystem_id,
             host_id)
(SELECT '2020-05-13 20:15:25.050399+00',
        'mounted',
        'f',
        't',
        '/tmp/fs2',
        35,
        fs.id,
        h.id
 FROM   chroma_core_managedfilesystem AS fs,
        chroma_core_managedhost AS h
 WHERE  fs.NAME = 'fs2'
        AND h.fqdn = 'c1.local')  
```

ip1981:
I needs some refactoring.

ip1981:
Refactored and made it ready for filesystem foreign key removal.

iml-services/iml-device/src/main.rs
```diff
@@ -106,29 +245,22 @@ async fn main() -> Result<(), ImlDeviceError> {
...
 
-    while let Some((f, d)) = s.try_next().await? {
-        let mut cache = cache2.lock().await;
+    // Django's artefact:
+    let lustreclientmount_ct_id = djct::table
+        .select(djct::id)
+        .filter(djct::model.eq("lustreclientmount"))
+        .first_async::<i32>(&pool)
+        .await
+        .ok();
```
jgrund, line 256:
```suggestion
    let lustreclientmount_ct_id = djct::table
        .select(djct::id)
        .filter(djct::model.eq("lustreclientmount"))
        .first_async::<i32>(&pool)
        .await
        .optional()?;
```

This differentiates between a db error and the row not existing, instead of treating everything as an error.

iml-services/iml-device/src/main.rs
```diff
@@ -13,8 +14,12 @@ use iml_device::{
...
 use iml_orm::{
-    models::{ChromaCoreDevice, NewChromaCoreDevice},
-    schema::chroma_core_device::{devices, fqdn, table},
+    models::{ChromaCoreDevice, ChromaCoreLustreclientmount, NewChromaCoreDevice},
+    schema::{
+        chroma_core_device as dvc, chroma_core_lustreclientmount as clmnt,
+        chroma_core_managedfilesystem as fs, chroma_core_managedhost as host,
+        django_content_type as djct,
+    },
     tokio_diesel::*,
```
jgrund, line 23:
```suggestion
    tokio_diesel::{AsyncRunQueryDsl as _, OptionalExtension as _},
```

Needed for `optional()` change below.

iml-services/iml-device/src/main.rs
```diff
@@ -45,6 +50,140 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
...
+        .filter(host::fqdn.eq(host.to_string()))
+        .get_result_async::<i32>(&pool)
+        .await
+    {
+        Ok(id) => {
+            tracing::debug!("Host id: {}", id);
+            id
+        }
+        Err(e) => {
+            tracing::error!("Host '{}' is unknown: {}", host, e);
```
jgrund, line 95: An error here doesn't just mean a host is unknown. 

It means a host is unknown or there was a db error. This is why `optional()` should be used.

iml-services/iml-device/src/main.rs
```diff
@@ -45,6 +50,140 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
...
+            return;
+        }
+    };
```
jgrund, line 98:
```suggestion
    let host_id = host::table
        .select(host::id)
        .filter(host::fqdn.eq(host.to_string()))
        .first_async::<i32>(&pool)
        .await
        .optional()?;
        
    let host_id = match host_id {
        Some(x) => x,
        None => {
            tracing::error!("Host '{}' is unknown", host);

            return Ok(());
        }
    };
    
    tracing::debug!("Host id: {}", host_id);
```

iml-services/iml-device/src/main.rs
```diff
@@ -45,6 +50,140 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
...
+
+    let lustre_mounts = mounts
+        .into_iter()
+        .filter(|m| m.fs_type.0.as_str() == "lustre")
```
jgrund, line 102:
```suggestion
        .filter(|m| m.fs_type.0 == "lustre")
```

iml-services/iml-device/src/main.rs
```diff
@@ -45,6 +50,140 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
...
+    tracing::debug!("Client mounts at {}: {:?}", host, &lustre_mounts);
+
+    let state_modified_at = chrono::offset::Utc::now().into_sql::<sql_types::Timestamptz>();
+
+    let insert_mounts = lustre_mounts
+        .iter()
+        .map(|(fs, tg)| {
+            let mp = tg
+                .clone()
+                .into_sql::<sql_types::Nullable<sql_types::Text>>();
```
jgrund, line 121:
```suggestion
            let mp = tg.into_sql::<sql_types::Nullable<sql_types::Text>>();
```

iml-services/iml-device/src/main.rs
```diff
@@ -45,6 +50,140 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
...
+        }
+    }
+}
+
+async fn update_client_mounts(
+    pool: &DbPool,
+    ct_id: Option<i32>,
+    host: &Fqdn,
+    mounts: &HashSet<Mount>,
+) {
```
jgrund, line 83:
```suggestion
) -> Result<(), ImlDeviceError> {
```

iml-services/iml-device/src/main.rs
```diff
@@ -45,6 +50,140 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
...
+                .on_conflict((clmnt::host_id, clmnt::filesystem_id, clmnt::not_deleted))
+                .do_update()
+                .set((
+                    clmnt::mountpoint.eq(excluded(clmnt::mountpoint)),
+                    clmnt::state.eq(excluded(clmnt::state)),
+                    clmnt::state_modified_at.eq(state_modified_at),
+                ))
+                .get_result_async::<ChromaCoreLustreclientmount>(&pool)
+        })
+        .collect::<Vec<_>>();
```
jgrund, line 154: This still does one query + 1 subselect for each incoming mount. The code I posted earlier does a single batch upsert, which is preferable to this approach.

ip1981: Yeah, but we still need to set the foreign key.

jgrund: I think we're going to have to get rid of the foreign key constraint first.

This patch won't be reliable without that change, because any mounts that come in before fs discovery will be discarded.

iml-services/iml-device/src/main.rs
```diff
@@ -45,6 +50,140 @@ async fn create_cache(pool: &DbPool) -> Result<Cache, ImlDeviceError> {
     Ok(Arc::new(Mutex::new(data)))
 }
 
+async fn update_devices(pool: &DbPool, host: &Fqdn, devices: &Device) {
```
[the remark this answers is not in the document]
ip1981: I wrote these functions just to keep the main shorter. They are dead-end and not supposed to return anything (just log issues, if any, and handle errors inside). They can be thought of as inline macros.

If they were to return their results, then I think the results should be handled and logged outside the functions, which would undermine the initial goal :)

iml-services/iml-device/src/main.rs
ip1981: might be an artifact :)

ip1981:
On the clippy warning "redundant closure found" => https://github.com/rust-lang/rust-clippy/issues/3791

jgrund:
> On the clippy warning "redundant closure found"

It makes sense since `vec![]` expands to `Vec::new()` in rust 1.44 (https://github.com/rust-lang/rust/pull/70632/). 

Workaround is to use `Vec::new` constructor points-free e.g:

```
.or_insert_with(Vec::new)
```

jgrund:
This won't pass because the python agent needs updates as well (I'll take care of this).

jgrund:
https://github.com/whamcloud/iml-agent/pull/150

PR #1925, opened 2020-05-26
Add enclosure model field

jgrund:
So we can tell which specific component to render in the GUI.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1925)


chroma_core/migrations/0017_sfaenclosure_sfastoragesystem.py
```diff
@@ -47,6 +47,7 @@ class Migration(migrations.Migration):
                 ("element_name", models.TextField()),
                 ("health_state", models.PositiveSmallIntegerField()),
                 ("health_state_reason", models.TextField()),
+                ("model", models.TextField()),
```
ip1981, line 50: It is ok to modify the migrations before it hits production, right?

jgrund: yes.

PR #1926, opened 2020-05-26
SFA Integration fixups

jgrund:
- Save enclosures before disks to ensure FK constraint is valid.
- Use serde_repr to {de}serialize c-like enums.
- Publically expose EnclosureType through iml-wire-types.
- Remove `Additions` and `Updates` from `get_changes`, it's not possible to determine what's changed without a key fn. Instead, return an `Upsert` struct that can be used to do an upsert in the db.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1926)

### DIFF
--- a/iml-sfa/src/main.rs
+++ b/iml-sfa/src/main.rs
@@ -68,19 +68,19 @@ async fn main() -> Result<(), ImlSfaError> {
 
         let (new_enclosures, x, new_drives) = future::try_join3(fut1, fut2, fut3).await?;
 
-        tracing::debug!("SfaStorageSystem {:?}", x);
-        tracing::debug!("SfaEnclosures {:?}", new_enclosures);
-        tracing::debug!("SfaDiskDrives {:?}", new_drives);
+        tracing::trace!("SfaStorageSystem {:?}", x);
+        tracing::trace!("SfaEnclosures {:?}", new_enclosures);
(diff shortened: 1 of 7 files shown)


iml-sfa/src/main.rs
```diff
@@ -68,19 +68,19 @@ async fn main() -> Result<(), ImlSfaError> {
 
         let (new_enclosures, x, new_drives) = future::try_join3(fut1, fut2, fut3).await?;
 
-        tracing::debug!("SfaStorageSystem {:?}", x);
-        tracing::debug!("SfaEnclosures {:?}", new_enclosures);
-        tracing::debug!("SfaDiskDrives {:?}", new_drives);
+        tracing::trace!("SfaStorageSystem {:?}", x);
+        tracing::trace!("SfaEnclosures {:?}", new_enclosures);
+        tracing::trace!("SfaDiskDrives {:?}", new_drives);
```
ip1981, line 73: I personally find "trace" useless. If it is worth to exists, it better be something like `trace!("I'm here!")` - i. e. just a mark. When it prints some data, it must be a `debug`.

jgrund: I changed this from debug to trace because it is *very* verbose. In general I won't want to see this being displayed (even when debugging), but every so often I will.

PR #1929, opened 2020-05-27
Add Action Runner Client crate

utopiabound:
This adds helper function invoke_rust_agent()
This does not support cancellation.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1929)


grafana/dashboards/foo.json
```diff
@@ -0,0 +1,2481 @@
+{
```
jgrund, line 1: cruft

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+        .try_next()
+        .await
+        .unwrap_or(None)
+        .is_some()
+}
+
+pub async fn invoke_rust_agent(
+    host: impl Into<Fqdn>,
+    command: impl Into<ActionName>,
+    args: serde_json::Value,
```
jgrund, line 25: This should take something that can be serialized to JSON. It shouldn't require the value to be JSON.

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+) -> AgentResult {
```
jgrund, line 26: Since this is all in Rust, we can do better than the Error side being just a string. Add a new enum for error handling and derive cases with thiserror.

jgrund, line 26: We need to return a future oneshot in addition to the result here. The oneshot will then race against the rpc to see who finishes first.

If the oneshot is called, it means we should cancel the RPC in flight.

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    let request_id = Uuid::new_v4().to_hyphenated().to_string();
+    let conn = if running_in_docker().await {
```
jgrund, line 28: This seems better suited for being synchronous and using lazy_static to ensure it's only called once at startup.

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+            }
+        }
+    };
+}
+
+#[derive(Error, Debug)]
+pub enum ImlActionClientError {
+    #[error(transparent)]
+    ReqwestError(#[from] reqwest::Error),
+    CancelledRequest,
```
jgrund, line 42:
```suggestion
    #[error("Request Cancelled")]
    CancelledRequest,
```

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl std::fmt::Display for ImlActionClientError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        match *self {
+            ImlActionClientError::ReqwestError(ref err) => write!(f, "{}", err),
+            ImlActionClientError::CancelledRequest => write!(f, "Request Cancelled"),
+        }
+    }
+}
```
jgrund, line 52: You should not need to implement this by hand.

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    command: impl Into<ActionName>,
+    args: impl serde::Serialize,
+) -> (oneshot::Sender<()>, impl Future<Output = Result<serde_json::Value, ImlActionClientError>>) {
+    let request_id = Uuid::new_v4().to_hyphenated().to_string();
+    let conn = if *RUNNING_IN_DOCKER {
+        get_action_runner_connect()
+    } else {
+        get_action_runner_socket()
+    };
+    let conn2 = conn.clone();
```
jgrund, line 57:
```suggestion
```

You can avoid this clone by refactoring the code a bit (see below).

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    let post = async move {
+        client
+            .post(&conn)
+            .json(&action)
+            .send()
+            .await?
+            .json()
+            .err_into()
+            .await
+    }.boxed();
```
jgrund, line 81:
```suggestion
        let post = client
        .post(&conn)
        .json(&action)
        .send()
        .err_into()
        .and_then(|resp| resp.json().err_into())
        .boxed();
```

This let's us reuse `client` and `conn` below.

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+                    }
+                ));
+                let client = reqwest::Client::new();
+                let _rc = client
+                    .post(&conn2)
+                    .json(&cancel)
+                    .send()
+                    .await;
+            }
+            either
```
jgrund, line 99: This will call cancel if the right side of the select completes (which is the RPC call).

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+        }).map(|either| match either {
+            Either::Left((_, b)) => { drop(b); Err(ImlActionClientError::CancelledRequest) }
+            Either::Right((a, _)) => a,
+        });
```
jgrund, line 103:
```suggestion
    let fut = async move {
        let either = future::select(c, post).await;

        match either {
            Either::Left((x, f)) => {
                if let Err(_) = x {
                    return f.await;
                };

                let cancel = ActionType::Remote((
                    host.into(),
                    Action::ActionCancel {
                        id: ActionId(request_id),
                    },
                ));

                let _rc = client.post(&conn).json(&cancel).send().await;

                Err(ImlActionClientError::CancelledRequest)
            }
            Either::Right((x, _)) => x,
        }
    };
```

I think this a little cleaner and avoids some closures. Of note is we return the right side of the either future if the left  resolves with a cancellation error. This means we can now drop the returned tx handle if we don't care about cancelling (which should generally not be the case).

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    host: impl Into<Fqdn> + Clone,
+    command: impl Into<ActionName>,
+    args: impl serde::Serialize,
+) -> (oneshot::Sender<()>, impl Future<Output = Result<serde_json::Value, ImlActionClientError>>) {
+    let request_id = Uuid::new_v4().to_hyphenated().to_string();
+    let conn = if *RUNNING_IN_DOCKER {
+        get_action_runner_connect()
+    } else {
+        get_action_runner_socket()
+    };
```
jgrund, line 56: I wonder if this whole block should be moved into iml-manager-env

iml-action-client/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+                    .lines()
+                    .filter(|l| {
+                        l.as_ref().unwrap_or(&"".to_string()).split(':').nth(1) == Some("docker")
+                    })
+                    .next()
+                    .is_some()
+            }
+        }
+    };
+}
```
jgrund, line 36: It seems like this would fit nicely in `iml-manager-env`.

iml-manager-env/src/lib.rs
```diff
@@ -8,8 +8,28 @@ use std::{
...
 
+lazy_static! {
+    static ref RUNNING_IN_DOCKER: bool = {
+        match std::fs::File::open("/proc/self/cgroup") {
+            Err(_) => false,
+            Ok(file) => {
+                let reader = std::io::BufReader::new(file);
+                reader
+                    .lines()
+                    .filter(|l| {
```
jgrund, line 23: I'll fix this before merging

iml-manager-env/src/lib.rs
```diff
@@ -10,6 +12,24 @@ use std::{
...
+                        l.as_ref().unwrap_or(&"".to_string()).split(':').nth(1) == Some("docker")
+                    })
+                    .next()
```
jgrund, line 26:
```suggestion
                    .find(|l| {
                        l.as_ref().unwrap_or(&"".to_string()).split(':').nth(1) == Some("docker")
                    })
```

PR #1931, opened 2020-05-28
Remove fk relation between lustreclientmount and fs

### LINKED ISSUE #1920: Remove fk relation between lustreclientmount and fs
We may learn about a lustreclientmount prior to learning about a fs. We should remove the fk association so we can keep mounts before we have a fs.

Precursor to #1801

Ref. 
* https://github.com/whamcloud/integrated-manager-for-lustre/blob/e415c2cd813de1e49254968450cd374f6c2dc2dc/chroma_core/models/client_mount.py#L155
* https://github.com/whamcloud/integrated-manager-for-lustre/blob/e415c2cd813de1e49254968450cd374f6c2dc2dc/chroma_core/models/client_mount.py#L198
* https://github.com/whamcloud/integrated-manager-for-lustre/blob/e415c2cd813de1e49254968450cd374f6c2dc2dc/chroma_core/models/client_mount.py#L295
* https://github.com/whamcloud/integrated-manager-for-lustre/blob/e415c2cd813de1e49254968450cd374f6c2dc2dc/chroma_core/models/client_mount.py#L355

ip1981:
We may learn about a lustreclientmount prior to learning about a fs. We
should remove the fk association so we can keep mounts before we have a fs.

Closes #1920.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1931)

### DIFF
--- a/chroma_api/host.py
+++ b/chroma_api/host.py
@@ -96,7 +96,7 @@ class HostTestValidation(HostValidation):
                     if not len(private_key.strip()):
                         errors["private_key"].append(self.mandatory_message)
         except KeyError:
-            #  What?  Now auth_type? assume existing key default case.
+            #  What?  Now auth_type? Assume existing key default case.
             pass
 
         return errors
@@ -135,10 +135,8 @@ class ServerProfileResource(ChromaModelResource):
 
 
 class ClientMountResource(ChromaModelResource):
-    # This resource is only used for integration testing.
-
     host = fields.ToOneField("chroma_api.host.HostResource", "host")
-    filesystem = fields.ToOneField("chroma_api.filesystem.FilesystemResource", "filesystem")
+    filesystem = fields.CharField()
     mountpoint = fields.CharField()
 
     class Meta:
@@ -157,7 +155,7 @@ class ClientMountResource(ChromaModelResource):
     def obj_create(self, bundle, **kwargs):
         request = bundle.request
         host = self.fields["host"].hydrate(bundle).obj
-        filesystem = self.fields["filesystem"].hydrate(bundle).obj
+        filesystem = ManagedFilesystem.objects.get(name=bundle.data["filesystem"])
         mountpoint = bundle.data["mountpoint"]
 
         client_mount = JobSchedulerClient.create_client_mount(host, filesystem, mountpoint)
@@ -209,7 +207,7 @@ class HostResource(StatefulModelResource, BulkResourceOperation):
         search = lambda cm: cm.host == bundle.obj
         mounts = ObjectCache.get(LustreClientMount, search)
         return [
-            {"filesystem_name": mount.filesystem.name, "mountpoint": mount.mountpoint, "state": mount.state}
+            {"filesystem_name": mount.filesystem, "mountpoint": mount.mountpoint, "state": mount.state}
             for mount in mounts
         ]
 
--- a/chroma_core/migrations/0020_clientmounts_remove_fk.py
+++ b/chroma_core/migrations/0020_clientmounts_remove_fk.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def resolve_fk(apps, schema_editor):
+    ClientMount = apps.get_model("chroma_core", "LustreClientMount")
+    ManagedFilesystem = apps.get_model("chroma_core", "ManagedFilesystem")
+    for m in ClientMount.objects.all():
+        m.filesystem = ManagedFilesystem.objects.get(id=m.filesystem).name
+        m.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("chroma_core", "0019_auto_20200529_2026"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="lustreclientmount",
+            name="filesystem",
+            field=models.CharField(help_text=b"Mounted filesystem", max_length=8),
+        ),
+        migrations.RunPython(resolve_fk),
+        migrations.RemoveField(model_name="managedhost", name="client_filesystems",),
+    ]
--- a/chroma_core/models/client_mount.py
+++ b/chroma_core/models/client_mount.py
@@ -8,6 +8,7 @@ from django.db.models import CASCADE
 from chroma_core.lib.cache import ObjectCache
 from chroma_core.models.utils import CHARFIELD_MAX_LENGTH
 from chroma_core.models.host import ManagedHost, HostOfflineAlert, HostContactAlert
+from chroma_core.models.filesystem import ManagedFilesystem
 from chroma_core.models.jobs import DeletableStatefulObject
 from chroma_core.models.jobs import StateChangeJob
 from chroma_core.models.alert import AlertState
@@ -18,7 +19,7 @@ from chroma_help.help import help_text
 
 class LustreClientMount(DeletableStatefulObject):
     host = models.ForeignKey("ManagedHost", help_text="Mount host", related_name="client_mounts", on_delete=CASCADE)
-    filesystem = models.ForeignKey("ManagedFilesystem", help_text="Mounted filesystem", on_delete=CASCADE)
+    filesystem = models.CharField(max_length=8, help_text="Mounted filesystem", null=False, blank=False,)
     mountpoint = models.CharField(
         max_length=CHARFIELD_MAX_LENGTH, help_text="Filesystem mountpoint on host", null=True, blank=True
     )
@@ -47,8 +48,10 @@ class LustreClientMount(DeletableStatefulObject):
             deps.append(DependOn(self.host.lnet_configuration, "lnet_up", fix_state="unmounted"))
 
         if state != "removed":
+            fs = ObjectCache.get_one(ManagedFilesystem, lambda mf: mf.name == self.filesystem)
+
             # Depend on the fs being available.
-            deps.append(DependOn(self.filesystem, "available", fix_state="unmounted"))
+            deps.append(DependOn(fs, "available", fix_state="unmounted"))
 
             # But if either the host or the filesystem are removed, the
             # mount should follow.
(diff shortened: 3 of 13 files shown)


chroma_core/models/client_mount.py
```diff
@@ -18,7 +18,9 @@
 
 class LustreClientMount(DeletableStatefulObject):
     host = models.ForeignKey("ManagedHost", help_text="Mount host", related_name="client_mounts", on_delete=CASCADE)
-    filesystem = models.ForeignKey("ManagedFilesystem", help_text="Mounted filesystem", on_delete=CASCADE)
+    filesystem = models.CharField(
+        max_length=CHARFIELD_MAX_LENGTH, help_text="Mounted filesystem", null=False, blank=False
+    )
```
jgrund, line 23: This should have the same restrictions as the `name` field on `ManagedFilesystem`:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/65490babfbae9d378d52be74f637923ef4850e47/chroma_core/models/filesystem.py#L44-L48

chroma_core/services/job_scheduler/job_scheduler.py
```diff
@@ -1106,13 +1106,10 @@ def key(td):
         # if we have an entry with 'root'=true then move it to the front of the list before returning the result
         return sorted(sorted_list, key=lambda entry: entry.get("root", False), reverse=True)
 
-    def create_client_mount(self, host_id, filesystem_id, mountpoint):
+    def create_client_mount(self, host_id, filesystem, mountpoint):
```
jgrund, line 1109: pass fs name instead of the whole filesystem object

jgrund:
These lines all reference the fs model directly (which is no longer possible with this patch):

https://github.com/whamcloud/integrated-manager-for-lustre/blob/6c2ec14bf1768256bbcff8c731016a43ff5e6843/chroma_core/models/client_mount.py#L69

https://github.com/whamcloud/integrated-manager-for-lustre/blob/6c2ec14bf1768256bbcff8c731016a43ff5e6843/chroma_core/models/client_mount.py#L67

https://github.com/whamcloud/integrated-manager-for-lustre/blob/6c2ec14bf1768256bbcff8c731016a43ff5e6843/chroma_core/models/client_mount.py#L53

chroma_core/services/job_scheduler/job_scheduler.py
ip1981: Actually, I believe this is the name, but I will double check.

ip1981:
Clean deploys and upgrades works fine.

chroma_api/host.py
```diff
@@ -96,7 +96,7 @@ def is_valid(self, bundle, request=None):
                     if not len(private_key.strip()):
                         errors["private_key"].append(self.mandatory_message)
         except KeyError:
-            #  What?  Now auth_type? assume existing key default case.
+            #  What?  Now auth_type? Assume existing key default case.
```
ip1981, line 99: I've enabled spell checking in Vim :)

chroma_api/host.py
```diff
@@ -138,7 +138,7 @@ class ClientMountResource(ChromaModelResource):
     # This resource is only used for integration testing.
 
     host = fields.ToOneField("chroma_api.host.HostResource", "host")
-    filesystem = fields.ToOneField("chroma_api.filesystem.FilesystemResource", "filesystem")
+    filesystem = fields.CharField()
```
ip1981, line 141: This was the cause of failing tests.

chroma_core/models/host.py
```diff
@@ -111,13 +111,6 @@ class ManagedHost(DeletableStatefulObject):
         default=False, help_text="True if there are package updates available for this server"
     )
 
-    client_filesystems = models.ManyToManyField(
-        "ManagedFilesystem",
-        related_name="workers",
-        through="LustreClientMount",
-        help_text="Filesystems for which this node is a non-server worker",
-    )
-
```
ip1981, line 120: This was a dead piece of code

chroma_api/host.py
ip1981, line 141: This is why the tests were failing. Note that this is unrelated to running in production ("This resource is only used for integration testing"). What do you think about rewriting the tests without this resource? We could just get filesystem details and number of mounts from there, I think.

chroma_api/host.py
```diff
@@ -209,7 +209,7 @@ def dehydrate_client_mounts(self, bundle):
         search = lambda cm: cm.host == bundle.obj
         mounts = ObjectCache.get(LustreClientMount, search)
         return [
-            {"filesystem_name": mount.filesystem.name, "mountpoint": mount.mountpoint, "state": mount.state}
+            {"filesystem_name": mount.filesystem, "mountpoint": mount.mountpoint, "state": mount.state}
```
ip1981, line 212: This is why the service was failing :)

chroma_api/host.py
jgrund: This resource is also available in production. If anything that comment should be removed as it's not accurate.

chroma_core/migrations/0020_clientmounts_remove_fk.py
```diff
@@ -0,0 +1,29 @@
...
+    dependencies = [
+        ("chroma_core", "0019_auto_20200529_2026"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="lustreclientmount",
+            name="filesystem",
+            field=models.CharField(help_text=b"Mounted filesystem", max_length=8),
+        ),
```
jgrund, line 26: Just to confirm, the `AlterField` / `RemoveField` sections were generated by `./manage.py makemigrations`?

chroma_core/models/client_mount.py
```diff
@@ -18,7 +19,12 @@
 
 class LustreClientMount(DeletableStatefulObject):
     host = models.ForeignKey("ManagedHost", help_text="Mount host", related_name="client_mounts", on_delete=CASCADE)
-    filesystem = models.ForeignKey("ManagedFilesystem", help_text="Mounted filesystem", on_delete=CASCADE)
+    filesystem = models.CharField(
+        max_length=ManagedFilesystem._meta.get_field("name").max_length,
```
jgrund, line 23:
```suggestion
        max_length=8,
```

It's better to just hardcode this, I don't think it will ever change.

chroma_core/models/host.py
jgrund: This is missing a corresponding migration.

chroma_core/models/stratagem.py
```diff
@@ -678,7 +679,8 @@ def get_steps(self):
         client_host = ManagedHost.objects.get(
             Q(server_profile_id="stratagem_client") | Q(server_profile_id="stratagem_existing_client")
         )
-        client_mount = LustreClientMount.objects.get(host_id=client_host.id, filesystem_id=self.filesystem.id)
+        filesystem = ManagedFilesystem.objects.get(id=self.filesystem)
```
jgrund, line 682: Isn't `self.filesystem` already a filesystem object?

chroma_core/services/job_scheduler/job_scheduler.py
jgrund: I mean we cannot pass the whole filesystem model into this RPC as it gets copied and is subject to mutation errors.

Previously it was `filesystem_id` and was looked up inside this method, now it should be `filesystem_name` and still be looked up inside this method.

chroma_core/services/job_scheduler/job_scheduler_client.py
```diff
@@ -245,7 +245,7 @@ def create_targets(cls, targets_data):
     def create_client_mount(cls, host, filesystem, mountpoint):
```
jgrund, line 245:
```suggestion
    def create_client_mount(cls, host, filesystem_name, mountpoint):
```

chroma_core/services/job_scheduler/job_scheduler_client.py
```diff
@@ -245,7 +245,7 @@ def create_targets(cls, targets_data):
...
         from chroma_core.models import LustreClientMount
 
-        client_mount_id = JobSchedulerRpc().create_client_mount(host.id, filesystem.id, mountpoint)
+        client_mount_id = JobSchedulerRpc().create_client_mount(host.id, filesystem, mountpoint)
```
jgrund, line 248:
```suggestion
        client_mount_id = JobSchedulerRpc().create_client_mount(host.id, filesystem_name, mountpoint)
```

chroma_core/models/host.py
ip1981: `./manage.py makemigrations` does not notice it, and there is not `client_filesystem` fields in the `chroma_core_managedhost` table.

chroma_core/migrations/0020_clientmounts_remove_fk.py
ip1981: Yes. Just checked it from scratch.

chroma_core/models/host.py
ip1981: I overlooked it : `migrations.RemoveField(model_name="managedhost", name="client_filesystems",),` :)
It may be argued to be in a separate migration file or even PR, but this this is a consequence of the main gain (removal of the foreign key), it might rest in here as well.

chroma_core/services/job_scheduler/job_scheduler.py
ip1981: I wonder if we should keep it as the id.

chroma_core/models/host.py
jgrund: I'm ambivalent if it needs to be in a separate file, but a field change and cooresponding migration must be in the same patch. Otherwise everyone using the patch will have out of date migrations.

chroma_core/services/job_scheduler/job_scheduler.py
jgrund: It's probably better as name so you don't need to look it up outside the RPC.

ip1981:
Does work when a FS is unknown:

```
chroma=> select * from chroma_core_lustreclientmount;
 id |       state_modified_at       |   state   | immutable_state | not_deleted | mountpoint | content_type_id | filesystem | host_id 
----+-------------------------------+-----------+-----------------+-------------+------------+-----------------+------------+---------
  1 | 2020-06-02 19:55:11.447428+00 | unmounted | f               | t           | /mnt/fs2   |              39 | fs2        |       5
(1 fila)

chroma=> \q
[root@adm ~]# iml filesystem list

[root@adm ~]# 

```

chroma_core/models/host.py
ip1981: Yes, it is in the migration file attached to this PR.

jgrund: Ah, sorry missed that.

PR #1933, opened 2020-05-28
Add id column for all SFA db objects

jgrund:
We have an (eventual) requirement to be able to monitor n SFA appliances at once. In order to do this, we need to have a composite primary key for most SFA objects of the `index` and `storage_system`. Unfortunately, Django does not allow for composite primary or foreign keys even though it's doable in Postgres.

Long term we should move our migrations out of Django and use diesel instead (which allows you to embed migrations into a binary) but for now we'll keep using the Django migrations and add an `unique_together` constraint for `index` and `storage_system` and drop the fk constraints for index on tables.

Additionally, there is a bug where a changed row will appear in both the `Upserts` and `Deletions` of get_changes. We need to have a key fn to determine if something actually belongs in the Deletions. 

Add an `Identifiable` trait that will uniquely id a record so it can be removed from `Deletions` if it's actually a change.

Finally, split out the SFA objects to their own modules within the iml-orm crate for better maintainability.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1933)


iml-orm/src/lib.rs
```diff
@@ -167,6 +167,12 @@ mod change {
         iter::FromIterator,
     };
 
+    pub trait Identifiable {
+        type Id: Eq + Ord;
```
nlinker, line 171: Maybe require `Debug` as well?

jgrund: I'm only adding the required traits for now. We don't use `Debug` or `Hash` when calling `Identifiable` below. If we end up needing them later we can change this restriction at the same time.

jgrund:
___
*[iml-sfa/src/main.rs, line 9 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1933#-M8VGwYk-QvE7yTf9bHZ:-M8VltjTEYn7PXa2Imog:b-nqdx8e) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/ddaa41465748816327dee06b026905ffe23d8b43/iml-sfa/src/main.rs#L9)):*

Should we rename all of these to not include `Sfa` prefix since all of them are in `sfa` module?
</blockquote></details>

These names mirror the XML elements they are pulled from.

iml-sfa/src/main.rs
```diff
@@ -66,83 +73,228 @@ async fn main() -> Result<(), ImlSfaError> {
 
         let fut3 = client.fetch_sfa_disk_drives(endpoints[0][0].clone());
 
-        let (new_enclosures, x, new_drives) = future::try_join3(fut1, fut2, fut3).await?;
+        let fut4 = client.fetch_sfa_jobs(endpoints[0][0].clone());
+
+        let fut5 = client.fetch_sfa_power_supply(endpoints[0][0].clone());
+
+        let (new_enclosures, x, new_drives, new_jobs, new_power_supplies) =
```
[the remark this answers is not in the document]
jgrund: That was due to the futures trying to read from postgres at the same time. In this case we are reading from an HTTP API.

PR #1934, opened 2020-05-29
Use Single-step server deployment

jgrund:
We currently use three phases to deploy a server

1. Preflight checks
2. Agent deployment
3. Setting profile

The only reason to have three steps is to validate profiles after
deploying the agent.

Instead, we can move the profile validation into the preflight check.

This will allow us to combine steps two and three.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1934)


jgrund:
Agent deployment gets rolled into overall profile deployment:

Screen Shot 2020-05-29 at 6 03 06 PM

PR #1959, opened 2020-06-11
Update iml-request-retry to pass iteration index

johnsonw:
There may be a case in which the future factory may need to generate a
different future based on the incremented tries. For example, if there
are two url's that can be fetched for the same information but the first
url fails, we should retry using the second url. If both of the urls are
stored in a vec, then this becomes trivial to fetch using the
incrementing index.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1959)

### DIFF
--- a/iml-manager-client/src/lib.rs
+++ b/iml-manager-client/src/lib.rs
@@ -173,7 +173,7 @@ pub async fn get_retry<T: DeserializeOwned + Debug>(
     let query = &query;
 
     retry_future(
-        move || get(client2.clone(), path.to_string(), query),
+        move |_| get(client2.clone(), path.to_string(), query),
         &mut policy,
     )
     .await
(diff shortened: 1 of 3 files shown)


jgrund:
@johnsonw needs more `rustfmt`

jgrund:
Should the count be 1 indexed instead of 0 indexed?

johnsonw:
Before the change it was 0 based so I figured I would keep it there. It works good for zero-based structures like vecs too but if you think it should start with 1 I can change it.

nlinker:
I think the documentation on this (i.e. README.md) needs a bit of update, also it needs to emphasize how it is different from `retry_future_gen` (it accepts also `Fn(request_no, E) -> ...` function). The extension itself is good, since we often rely on the request numbers.

PR #1963, opened 2020-06-11
Update footer and change "support" to "help"

johnsonw:
- Add exascaler_version to settings and conf
- Display appropriate branding and version in footer

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1963)


johnsonw:
RPM Setup:
image

iml-manager-env/src/lib.rs
```diff
@@ -86,6 +86,11 @@ pub fn get_version() -> String {
     get_var("VERSION")
 }
 
+/// Get exascaler version
+pub fn get_exascaler_version() -> String {
+    get_var("EXASCALER_VERSION")
```
jgrund, line 91: This won't be guaranteed to be there, so this needs to return an option

johnsonw:
Do not include exa version in print settings if it is none

PR #1978, opened 2020-06-17
Use the filesystem from the object cache when unmounting / mounting filesystems.

johnsonw:
Use the filesystem from the object cache when unmounting / mounting filesystems.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1978)

### DIFF
--- a/chroma_core/models/client_mount.py
+++ b/chroma_core/models/client_mount.py
@@ -293,7 +293,17 @@ class MountLustreFilesystemsJob(AdvertisedJob):
     def get_steps(self):
         search = lambda cm: (cm.host == self.host and cm.state == "unmounted")
         unmounted = ObjectCache.get(LustreClientMount, search)
-        args = dict(host=self.host, filesystems=[(m.filesystem.mount_path(), m.mountpoint) for m in unmounted])
+        args = {
+            "host": self.host,
+            "filesystems": [
+                (
+                    ObjectCache.get_one(ManagedFilesystem, lambda mf, mtd=m: mf.name == mtd.filesystem).mount_path(),
+                    m.mountpoint,
+                )
+                for m in unmounted
+            ],
+        }
+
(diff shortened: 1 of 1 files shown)


chroma_core/models/client_mount.py
```diff
@@ -293,7 +293,17 @@ def description(self):
     def get_steps(self):
         search = lambda cm: (cm.host == self.host and cm.state == "unmounted")
         unmounted = ObjectCache.get(LustreClientMount, search)
-        args = dict(host=self.host, filesystems=[(m.filesystem.mount_path(), m.mountpoint) for m in unmounted])
+        args = {
+            "host": self.host,
+            "filesystems": [
+                (
+                    ObjectCache.get_one(ManagedFilesystem, lambda mf, mtd=m: mf.name == mtd.filesystem).mount_path(),
```
ip1981, line 300: What's the difference comparing to the above two lines (`search=` and `unmounted=`)?

johnsonw: The `unmounted` is a list of unmount `LustreClientMount` objects. We loop through each of these and match on the filesystem whose name in the umounted client mount objects matches the filesystem name. That will then give a list of filesystems that correspond to the mounts. For example, say we have:

Filesystems:
fs1, fs2, and fs3.

Mounts:
mnt1, mnt2

mnt1 is associated with fs1
mnt2 is associated with fs2

We unmount both of these instances but their object still exists. We then later want to mount them. The expression would return a list with:
[fs1, fs2]

johnsonw: The problem i'm working through at the moment is that the mount path is erased when a client is unmounted.

jgrund: > That can (should) be in a separate patch.

PR #1979, opened 2020-06-17
Fix policy index in sfa

johnsonw:
Fix policy index in sfa

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1979)


iml-sfa/src/main.rs
```diff
@@ -36,7 +36,7 @@ enum ImlSfaError {
 
 fn create_retry_endpoint_policy<E: Debug>(len: u32) -> impl RetryPolicy<E> + Send {
     move |k: u32, e| {
-        if k < len {
+        if k < len - 1 {
             RetryAction::RetryNow
         } else {
             RetryAction::ReturnError(e)
```
ip1981, line 42:
```rust
if k >= len {
   RetryAction::ReturnError(e)
} else {
  RetryAction::RetryNow
}
```

:) Just an idea

ip1981: And you probably would not need a test for that.

iml-sfa/src/main.rs
```diff
@@ -36,7 +36,7 @@ enum ImlSfaError {
 
 fn create_retry_endpoint_policy<E: Debug>(len: u32) -> impl RetryPolicy<E> + Send {
     move |k: u32, e| {
-        if k < len {
+        if k < len - 1 {
```
ip1981, line 39: What if `len==0`? What would it mean?

iml-sfa/src/main.rs
johnsonw: sure, makes sense

iml-sfa/src/main.rs
johnsonw: When len is 0 that would be the first attempt at making the call.

iml-sfa/src/main.rs
johnsonw: actually, my unit test failed with this. Looking.

johnsonw: ah right, it would have to be >= len - 1. I think i'm going to leave it as is.

iml-sfa/src/main.rs
johnsonw: I misread this earlier. Basically the policy is used to limit how many times we retry hitting an endpoint. So if we create a policy with a length of two then that means it will try to hit the endpoint twice. If we create a policy with a length of zero then it would never attempt to hit the endpoint. I'll add a unit test to cover this.

ip1981: I would expect that retry mean literally retry, and thus `len` would mean the allowed number of failures: 0 mean hit just one time, 1 - retry one one if failed, and so on.

iml-sfa/src/main.rs
```diff
@@ -357,3 +357,55 @@ impl SfaClassExt for Client {
         Ok(ys)
     }
 }
+
+#[cfg(test)]
```
jgrund, line 361: This belongs in the retry-future crate instead of here.

johnsonw: Since it's testing the policy related to the fetch in this crate I don't think it makes sense to move it. I'll just remove the tests.

PR #1981, opened 2020-06-17
Fix mount jobs such that they use the correct mount point

johnsonw:
Fix mount jobs such that they use the correct mount point

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1981)


johnsonw:
Unmounting:
image

Mounting:
image

PR #1986, opened 2020-06-18
Update influxdb healthcheck

### LINKED ISSUE #1984: Influx healthcheck causes high CPU usage
In:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/305fcf1ff4f4a021479c09b3d704d9e4dccefb33/docker/docker-compose.yml#L35-L40

We run a healthcheck for influxdb every 5s. This check appears to be expensive taking around 80% of a core while it runs.

Screen Shot 2020-06-18 at 10 22 15 AM

Instead we can do a check like:

`curl -v -sL -I influxdb:8086/ping?wait_for_leader=30s`

Which does not appear to take much resource to run.

jgrund:
Use a lower resource check.

Fixes #1984.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1986)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -33,9 +33,9 @@ services:
       - influxdb:/var/lib/influxdb/
       - "manager-config:/var/lib/chroma"
     healthcheck:
-      test: ["CMD-SHELL", "influx -host influxdb -execute 'show databases'"]
+      test: ["CMD-SHELL", "curl --fail influxdb:8086/ping?wait_for_leader=10s"]
       interval: 5s
-      timeout: 5s
+      timeout: 15s
       retries: 5
       start_period: 60s
     ports:


docker/docker-compose.yml
```diff
@@ -33,9 +33,9 @@ services:
       - influxdb:/var/lib/influxdb/
       - "manager-config:/var/lib/chroma"
     healthcheck:
-      test: ["CMD-SHELL", "influx -host influxdb -execute 'show databases'"]
+      test: ["CMD-SHELL", "curl --fail influxdb:8086/ping?wait_for_leader=10s"]
```
utopiabound, line 36: This can also just use `influx -host influxdb -execute 'exit'` instead.  It's what's used during setup to wait for influx to be online.

jgrund: Looks like that takes ~ 50% of a core:

Screen Shot 2020-06-18 at 11 08 10 AM


Keep in mind this is being run every 5s.

PR #2006, opened 2020-06-23
Fix user disconnects

jgrund:
In: https://github.com/whamcloud/integrated-manager-for-lustre/blob/83d065653a55ee4c1a6068c1d024f02a02c73172/iml-warp-drive/src/users.rs#L54-L58

We spawn a task to disconnect users. However, we move both the tx and rx
handles into the task and block on the tx being cancelled.

Due to this approach, we will not disconnect users and instead buffer
their handles.

Instead, we should use .retain when sending to users and remove the ones
that error.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2006)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -392,6 +392,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cloudabi"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


jgrund:
Pass in https://build.whamcloud.com/view/IML/job/manager-for-lustre/993/

PR #2007, opened 2020-06-23
EX-12345 FileSync: add single search, iml-socket arguments

beevans:
Add arguments for an IML socket from lipe
Add single-instance search, no config file generation

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2007)


jgrund:
The base branch needs to change to master for this patch and be repushed

PR #2008, opened 2020-06-23
Do not set environment variables in tests

### LINKED ISSUE #1826: Intermittent test failure: action_plugins::lpurge::lpurge_conf_tests::works

jgrund:
We have a number of intermittently failing tests that are due to setting environment variables within the tests.

Rust tests run in parallel and the environment is basically a singleton. Therefore, these tests leak state with each other and are non-deterministic.

In addition, we are pulling items between action plugins, where we could factor it out into the env instead.

Fixes #1826.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2008)

### DIFF
--- a/iml-agent/src/action_plugins/lamigo.rs
+++ b/iml-agent/src/action_plugins/lamigo.rs
@@ -42,10 +42,9 @@ pub struct Config {
     mailbox: String,
 }
 
-impl fmt::Display for Config {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
+impl Config {
+    fn generate_unit(&self, mailbox: String) -> String {
+        format!(
             "mdt={fs}-MDT{mdt:04x}\n\
              mountpoint={mountpoint}\n\
              user={user}\n\
@@ -61,7 +60,7 @@ impl fmt::Display for Config {
             age = self.min_age,
             fast = self.hot_pool,
             slow = self.cold_pool,
-            mailbox = postoffice::socket_name(&self.mailbox),
+            mailbox = mailbox,
         )
     }
 }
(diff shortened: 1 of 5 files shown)


iml-agent/src/action_plugins/lamigo.rs
```diff
@@ -42,10 +42,9 @@ pub struct Config {
     mailbox: String,
 }
 
-impl fmt::Display for Config {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
+impl Config {
+    fn generate_unit(&self, mailbox: String) -> String {
```
ip1981, line 46: I've been doing almost the same :)

iml-agent/src/action_plugins/lpurge.rs
```diff
@@ -125,7 +126,9 @@ mod lpurge_conf_tests {
             freelo: 60,
             mailbox: "foobar".to_string(),
         };
-        let file = conf_name(&cfg).await.expect("name could not be created");
+        let file = conf_name(&cfg, "/etc/lpurge/{fs}/{ost}-{pool}.conf")
```
ip1981, line 129: :+1:

iml-agent/src/action_plugins/lpurge.rs
```diff
@@ -26,10 +26,9 @@ pub struct Config {
     mailbox: String,
 }
 
-impl fmt::Display for Config {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
+impl Config {
+    fn generate_unit(&self, socket: String) -> String {
```
ip1981, line 30: I think this could be unified with `expand_path_fmt`: we just format strings, file name of content.

ip1981:
There are some unused imports.

iml-agent/src/action_plugins/lpurge.rs
jgrund: Can you expand on how the unification would work? One item is generating a systemd unit the other is generating a path with optional values.

Are you suggesting to use strfmt for both fns?

ip1981: Yes, strfmt for both. Even maybe in one pass, like `cfg.write(socket, file_template);`

PR #2026, opened 2020-06-30
Handle sessions in one place

### LINKED ISSUE #1805: Logged out users can still see logged-in resources
if you log out, you still have access to what only logged in users have access, e.g. you can follow links on the views to other views that are unavailable to logged out users, or you can use breadcrumbs for navigation to such views.

We should use the session change monitoring to change breadcrumb display and redirect from authorized pages.

### LINKED ISSUE #2030: Endless session removal and retrival on the login page
**Describe the bug**
When on the login page the browser continuously makes requests to `/api/session/` deleting and retrieving sessions.
This happens when anonymous read is disabled.

**To Reproduce**
1. Disable anonymous read: 
 ```
# echo ALLOW_ANONYMOUS_READ=False >> /usr/share/chroma-manager/local_settings.py
# chroma-config restart
```
2. Go to the GUI, find yourself at the login page.
3. Open the console and see this:
delete

### PR DESCRIPTION
**Pull request by ip1981**
This includes logging in and logging out.

`logged_out` is removed because "disabled" is not working with `a` anyway.
Can be added to the auth::Model if needed.

Closes  #2030, #1805,  #1737, #1735 (supersedes https://github.com/whamcloud/integrated-manager-for-lustre/pull/1755).

sessions

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2026)

### DIFF
--- a/iml-gui/crate/src/auth.rs
+++ b/iml-gui/crate/src/auth.rs
@@ -14,7 +14,6 @@ use web_sys::HtmlDocument;
 #[derive(Default)]
 pub struct Model {
     session: Option<Session>,
-    pub state: State,
     request_controller: Option<fetch::RequestController>,
     cancel: Option<oneshot::Sender<()>>,
 }
@@ -25,45 +24,25 @@ impl Model {
     }
 }
 
-#[derive(PartialEq, Eq)]
-pub enum State {
-    Fetching,
-    Stopped,
-}
-
-impl Default for State {
-    fn default() -> Self {
-        Self::Fetching
-    }
-}
-
 #[allow(clippy::large_enum_variant)]
 #[derive(Clone, Debug)]
 pub enum Msg {
     Fetch,
     Fetched(fetch::FetchObject<Session>),
-    SetSession(Session),
+    Logout,
+    LoggedIn,
     Loop,
-    Stop,
     Noop,
 }
 
 pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg, GMsg>) {
     match msg {
         Msg::Fetch => {
-            model.state = State::Fetching;
             model.cancel = None;
 
             let request = fetch_session().controller(|controller| model.request_controller = Some(controller));
 
             orders.skip().perform_cmd(request.fetch_json(Msg::Fetched));
-        }
-        Msg::SetSession(session) => {
-            if session.needs_login() {
-                orders.send_g_msg(GMsg::RouteChange(Route::Login.into()));
-            }
-
-            model.session = Some(session);
         }
         Msg::Fetched(data) => {
             match data.response() {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: iml-gui/crate/src/auth.rs [hunk 1]
```diff
@@ -95,6 +88,19 @@ pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg, GMsg>)
...
             };
         }
+        Msg::LoggedIn => {
+            orders.skip();
+            orders.send_msg(Msg::Fetch);
+            orders.send_g_msg(GMsg::RouteChange(Route::Dashboard.into()));
+        }
+
+        Msg::Logout => {
+            orders.perform_g_cmd(
```
ip1981, line 98: I think we could make a call of `event_source::init(orders);` from here to reset the records, but I don't this it is needed yet.

iml-gui/crate/src/auth.rs
```diff
@@ -95,6 +88,19 @@ pub fn update(msg: Msg, model: &mut Model, orders: &mut impl Orders<Msg, GMsg>)
...
+                fetch_session()
+                    .method(fetch::Method::Delete)
+                    .fetch(|_| GMsg::AuthProxy(Box::new(Msg::LoggedIn))),
```
ip1981, line 101: > A samurai without a sword is like a samurai with one, but only without one.

ip1981:
> Just `Login` to be symmetric with `Logout`?

They are different: Logout is a command, LoggedIn is a result.

ip1981:
There might be an minor issue when anonymous access is not allowed, but easy to fix if any.

ip1981:
> There might be an minor issue when anonymous access is not allowed, but easy to fix if any.

This happens without this patch: https://github.com/whamcloud/integrated-manager-for-lustre/issues/2030

iml-gui/crate/src/auth.rs
```diff
@@ -25,46 +24,26 @@ impl Model {
...
     match msg {
         Msg::Fetch => {
-            model.state = State::Fetching;
             model.cancel = None;
 
             let request = fetch_session().controller(|controller| model.request_controller = Some(controller));
 
             orders.skip().perform_cmd(request.fetch_json(Msg::Fetched));
         }
-        Msg::SetSession(session) => {
```
ip1981, line 61: I think this was the cause of #2030.

iml-gui/crate/src/auth.rs
jgrund: Why would this be necessary?

ip1981: For example, if we had more sophisticated access control, some _data_ (e. g. filesystems) could be made hidden to users, and we would need to reset local state. This is hypothetical, I don't think we will ever need this :)

PR #2045, opened 2020-07-02
Only unmount clients that were on the current host

### LINKED ISSUE #2044: iml-device unmounting mounted clients in the db
In:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/c2dcd1e90c66f10bf7d1e65406b796bfab0d9ac1/iml-services/iml-device/src/main.rs#L163-L168

We unmount clients that have been mounted.

This is because we unmount everything that does not match updated filesystems.

Instead we should only unmount records with the same host_id that were not previously mounted in the current tick.

jgrund:
and were not updated in the current tick

Fixes #2044.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2045)


iml-services/iml-device/src/main.rs
```diff
@@ -165,9 +165,11 @@ async fn update_client_mounts(
                 mountpoint = Null,
                 state = 'unmounted',
                 state_modified_at = now()
-            WHERE filesystem != ALL ($1)
+            WHERE host_id = $1
```
ip1981, line 168: This was lost in https://github.com/whamcloud/integrated-manager-for-lustre/pull/2025, right?

iml-services/iml-device/src/main.rs
```diff
@@ -165,9 +165,11 @@ async fn update_client_mounts(
...
+            AND id != ALL($2)
```
ip1981, line 169: As I understand, `filesystem` vs `id` shouldn't matter, the former is more descriptive, but the latter is shorter :)

iml-services/iml-device/src/main.rs
jgrund: Yes

iml-services/iml-device/src/main.rs
jgrund: That would only be the case if (filesystem, host_id) was guaranteed to be unique, but it's not.

iml-services/iml-device/tests/integration_tests.rs
```diff
@@ -0,0 +1,144 @@
...
+    immutable_state: bool,
+    not_deleted: Option<bool>,
+    content_type_id: Option<i32>,
+    filesystem: String,
+    host_id: i32,
+    mountpoints: Vec<String>,
+}
+
+#[tokio::test]
+#[ignore = "Requires an active and populated DB"]
```
ip1981, line 32: Wouldn't it better belong to the tests crate?

jgrund: Tests crate?

ip1981: Well, like `iml-system-docker-tests`

jgrund: This test is meant to sit between unit tests and system tests (integration test). I can run this test in 30 seconds instead of 2 hours once I have a populated db.

chroma_core/models/copytool.py
```diff
@@ -387,7 +387,7 @@ def run(self, kwargs):
             bin_path=copytool.bin_path,
             archive_number=copytool.archive,
             filesystem=filesystem.name,
-            mountpoint=client_mount.mountpoint,
+            mountpoint=client_mount.mountpoints[0],
```
johnsonw, line 390: Are we guaranteed that mountpoints isn't empty?

chroma_core/models/stratagem.py
```diff
@@ -687,7 +687,7 @@ def get_steps(self):
                 {
                     "client_args": (
                         client_host.fqdn,
-                        client_mount.mountpoint,
+                        client_mount.mountpoints[0],
```
johnsonw, line 690: Are we guaranteed that mountpoints isn't empty?

iml-services/iml-device/Cargo.toml
```diff
@@ -1,25 +1,29 @@
...
+iml-manager-env = {path = "../../iml-manager-env", version = "0.3"}
+iml-postgres = {path = "../../iml-postgres", version = "0.3"}
+iml-rabbit = {path = "../../iml-rabbit", version = "0.3"}
+iml-service-queue = {path = "../iml-service-queue", version = "0.3"}
+iml-tracing = {version = "0.2", path = "../../iml-tracing"}
+iml-wire-types = {path = "../../iml-wire-types", version = "0.3", features = ["postgres-interop"]}
+serde = {version = "1", features = ["derive"]}
 serde_json = "1.0"
-sqlx = { git = "https://github.com/launchbadge/sqlx", default-features = false, features = ["json", "macros", "offline", "postgres", "runtime-tokio" ] }
+sqlx = {git = "https://github.com/jgrund/sqlx", branch = "workspace-prepare", default-features = false, features = ["json", "macros", "offline", "postgres", "runtime-tokio"]}
```
johnsonw, line 19: Does sqlx need some updates?

chroma_core/models/copytool.py
jgrund: Yes, the invariant is the same as before, now it's just a list.

chroma_core/models/stratagem.py
jgrund: Yes, the invariant is the same as before, now it's just a list.

iml-services/iml-device/Cargo.toml
jgrund: Lot's of them. We should be able to switch off my branch once 0.4 is released hopefully.

PR #2049, opened 2020-07-06
Do not install postgres-devel

### LINKED ISSUE #2048: Remove postgresql96-devel install from makefile
In 

https://github.com/whamcloud/integrated-manager-for-lustre/blob/1890e4731ae07ef2cff9b437bdf3234f17a0121c/.copr/Makefile#L38-L40

We may install `postgresql96-devel` if it does not already exist.

This is an issue, as we may hit the case where repodata is out of date, causing build failures.

The right way to fix this is to:

1. Add the postgres repo and `postgresql96-devel` package here:

https://github.com/whamcloud/Dockerfiles/blob/4db15356ddcc46d30b5693f79c6a34ad05c46779/copr-rust/Dockerfile#L7

2. Remove the conditional `postgresql96-devel` install here:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/1890e4731ae07ef2cff9b437bdf3234f17a0121c/.copr/Makefile#L38-L40

jgrund:
Fixes #2048.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2049)

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -35,10 +35,6 @@ substs: $(SUBSTS)
 	chmod +x $(SUBSTS_SHELL)
 
 srpm:
-	if ! rpm -q postgresql96-devel 2> /dev/null; then \
-		yum install -y postgresql96-devel; \
-	fi
-
 	mkdir -p ${TMPDIR}/_topdir/{SOURCES,SPECS}
 	mkdir -p ${TMPDIR}/release/rust-iml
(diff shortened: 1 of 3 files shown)


jgrund:
Passed in https://build.whamcloud.com/job/manager-for-lustre/1146/ which is based on this patch.

PR #2060, opened 2020-07-09
update monitored ldiskfs to use lvm and ha

johnsonw:
update monitored ldiskfs to use lvm and ha

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2060)

### DIFF
--- a/iml-system-test-utils/src/lib.rs
+++ b/iml-system-test-utils/src/lib.rs
@@ -514,22 +514,36 @@ pub async fn configure_docker_network(config: &Config) -> Result<(), CmdError> {
 }
 
 async fn create_monitored_ldiskfs(config: &Config) -> Result<(), CmdError> {
-    let xs = config
-        .storage_servers()
-        .into_iter()
-        .map(|x| {
-            tracing::debug!("creating ldiskfs fs for {}", x);
-            async move {
-                vagrant::provision_node(x, "configure-lustre-network,create-ldiskfs-fs,create-ldiskfs-fs2,mount-ldiskfs-fs,mount-ldiskfs-fs2")
-                    .await?
-                    .checked_status()
-                    .await?;
+    let futures = config.storage_servers().into_iter().map(|x| {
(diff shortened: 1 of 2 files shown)


iml-system-test-utils/src/lib.rs
```diff
@@ -514,22 +514,36 @@ pub async fn configure_docker_network(config: &Config) -> Result<(), CmdError> {
...
+        }
+    });
 
-    try_join_all(xs).await?;
+    try_join_all(futures).await?;
+
+    let futures = config.storage_servers().into_iter().map(|x| {
+        tracing::debug!("setting up HA for {}", x);
+        async move {
+            vagrant::provision_node(x, "ha-ldiskfs-lvm-fs-setup")
```
jgrund, line 537: This is missing a step

johnsonw: Which step is this missing? The docs show:
```
 vagrant provision --provision-with=install-ldiskfs-no-iml,configure-lustre-network,create-ldiskfs-lvm-fs,ha-ldiskfs-lvm-fs-prep
 VBOX_PASSWD=<ROOT_PW_HERE>
 vagrant provision --provision-with=ha-ldiskfs-lvm-fs-setup
```
The `install-ldiskfs-no-iml` step is being done earlier and I was able to run `ha-ldiskfs-lvm-fs-setup` without the environment variable. Is there something else that's missing?

jgrund:
> and I was able to run ha-ldiskfs-lvm-fs-setup without the environment variable

This breaks the VBox fence agent, which cannot communicate without the var being set

johnsonw:
Ah I see. Ok I'll make sure to run the command with that environment variable set.

jgrund:
> Ah I see. Ok I'll make sure to run the command with that environment variable set.

Make sure it's obfuscated and doesn't appear in the source or test output.

PR #2065, opened 2020-07-13
Update test debugging

johnsonw:
Update test debugging

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2065)

### DIFF
--- a/iml-system-test-utils/src/lib.rs
+++ b/iml-system-test-utils/src/lib.rs
@@ -273,14 +273,22 @@ impl Config {
     pub fn get_setup_config(&self) -> String {
         match &self.test_type {
             TestType::Rpm => format!(
-                r#"USE_STRATAGEM = {}
-BRANDING = "{}""#,
+                r#"USE_STRATAGEM={}
+BRANDING={}
+LOG_LEVEL=10
+RUST_LOG=debug
+NTP_SERVER_HOSTNAME=adm.local
+"#,
                 if self.use_stratagem { "True" } else { "False" },
                 self.branding.to_string()
             ),
             TestType::Docker => format!(
                 r#"USE_STRATAGEM={}
-BRANDING={}"#,
+BRANDING={}
+LOG_LEVEL=10
+RUST_LOG=debug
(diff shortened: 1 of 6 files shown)


iml-system-test-utils/src/lib.rs
```diff
@@ -273,14 +273,22 @@ impl Config {
     pub fn get_setup_config(&self) -> String {
         match &self.test_type {
             TestType::Rpm => format!(
-                r#"USE_STRATAGEM = {}
-BRANDING = "{}""#,
+                r#"USE_STRATAGEM={}
+BRANDING={}
+LOG_LEVEL=10
```
nlinker, line 278: This makes me thinking what the 10 is. I was expecting `LOG_LEVEL` to be usual one of `trace`, `debug`, `warn` etc.
How is it used?

ip1981:
```
$ python
>>> import logging
>>> logging.DEBUG
10
```

iml-system-test-utils/src/lib.rs
```diff
@@ -273,14 +273,22 @@ impl Config {
...
+RUST_LOG=debug
```
ip1981, line 279: Just a note, this makes less noise:

```
RUST_LOG=debug,hyper=warn,lapin=warn,warp=info
```

iml-system-test-utils/src/lib.rs
johnsonw: `LOG_LEVEL` is the python's log level and it's an integer. Here is the python repl output:
```
>>> import logging
>>> logging.DEBUG
10
>>> logging.INFO
20
```

iml-system-test-utils/src/lib.rs
johnsonw: If we find that we need to change this i'm willing to do so in a separate patch.

jgrund: I think we will actually want the verbosity in this case.

PR #2067, opened 2020-07-14
Don't accept invalid certificates

mkpankov:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2067)


mkpankov:
This is not needed / would forbid using self-signed certs

PR #1832, opened 2020-04-28
Task mailbox

utopiabound:
Convert Fast File Scan to use Task Queues
Flesh out ingest/outgest of task queues
Issue #1875 

This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/1832)

PR #2079, opened 2020-07-15
iml-postgres: Add optional connection pool

utopiabound:
Add deadpool based thread-pool to iml-posgres

Related #1832 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2079)

### DIFF
--- a/iml-postgres/src/pool.rs
+++ b/iml-postgres/src/pool.rs
@@ -1,0 +1,21 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+pub use deadpool_postgres::{config::ConfigError, Client, Pool, PoolError};
+
+use iml_manager_env::get_db_conn_hash;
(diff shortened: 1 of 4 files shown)


iml-postgres/src/pool.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+pub use deadpool_postgres::{config::ConfigError, Client, Pool, PoolError};
+
+use iml_manager_env::get_db_conn_hash;
+use tokio_postgres::NoTls;
+
+pub async fn pool() -> Result<Pool, ConfigError> {
```
jgrund, line 10: Instead of adding this connection pool, how much work would it be to use sqlx which has built in pooling?

https://github.com/whamcloud/integrated-manager-for-lustre/blob/master/iml-postgres/src/lib.rs#L24-L51

PR #2087, opened 2020-07-17
Add "make check" to Makefile

utopiabound:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2087)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,16 @@ all: copr-rpms rpms
 local:
 	$(MAKE) RPM_DIST="0.$(shell date '+%s')" all
 
+check:
+	black --check ./
+	cargo fmt --all -- --check
+	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo check
+	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo clippy -- -W warnings
+
+fmt:
+	black ./
+	cargo fmt --all
+
 iml-gui-rpm:
 	$(MAKE) -f .copr/Makefile iml-gui-srpm outdir=.
 	rpmbuild --rebuild ${RPM_OPTS} _topdir/SRPMS/rust-iml-gui-*.src.rpm


Makefile
```diff
@@ -40,6 +40,12 @@ all: copr-rpms rpms
 local:
 	$(MAKE) RPM_DIST="0.$(shell date '+%s')" all
 
+check:
```
ip1981, line 43: Technically, blank and rustfmt are changing the source code, so  "check" is misleading.

Makefile
```diff
@@ -40,6 +40,12 @@ all: copr-rpms rpms
...
+	black chroma_core/ chroma_api/ chroma_cli/ tests/
```
ip1981, line 44: I always run blank without any dirs, it just works.

Makefile
jgrund: `fmt` would be a better target name

Makefile
```diff
@@ -40,6 +40,12 @@ all: copr-rpms rpms
...
+	cargo fmt
+	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo check
+	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo clippy -- -W warnings
```
jgrund, line 47: This will only work for the first compile. There is a bug where clippy doesn't properly report errors when doing incremental compilation.

Makefile
jgrund: Though there are things for both formatting and checking. Perhaps create two targets `fmt` and `check`, where `check` depends on `fmt`

Makefile
ip1981: It looks fixed now.

Makefile
utopiabound:
```
nate@gopher ~/src/integrated-manager-for-lustre>$ black
No Path provided. Nothing to do 😴
```

jgrund: `black --check ./`

PR #2100, opened 2020-07-23
EX-1465 - Create and populate ldev.conf file on MGT nodes

johnsonw:
Create action plugin to write ldev.conf on MGS node.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2100)


iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -62,7 +62,8 @@ pub fn create_registry() -> action_plugins::Actions {
             action_configure::update_and_write_new_config,
         )
         .add_plugin("is_ntp_configured", is_ntp_configured::is_ntp_configured)
-        .add_plugin("create_ltuer_conf", ltuer::create_ltuer_conf);
+        .add_plugin("create_ltuer_conf", ltuer::create_ltuer_conf)
+        .add_plugin("create_ldev_conf", snapshot::create_ldev_conf);
```
ip1981, line 66: Maybe "snapshot" is not the best name? :)

iml-agent/src/action_plugins/snapshot.rs
```diff
@@ -0,0 +1,199 @@
...
+                primary: "oss2".into(),
+                failover: Some("oss1".into()),
+                label: "zfsmo-OST00012".into(),
+                device: "zfs:ost18/ost18".into(),
+            },
+            LdevEntry {
+                primary: "oss2".into(),
+                failover: Some("oss1".into()),
+                label: "zfsmo-OST00013".into(),
+                device: "zfs:ost19/ost19".into(),
```
ip1981, line 185: Do you think more than two entries are necessary?

iml-agent/src/action_plugins/snapshot.rs
```diff
@@ -0,0 +1,199 @@
...
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_create_ldev_conf() -> Result<(), ImlAgentError> {
+        // oss2 oss1 zfsmo-OST0013 zfs:ost19/ost19
```
ip1981, line 47: This could be an inline test (without a snapshot file) :)

iml-agent/src/action_plugins/snapshot.rs
```diff
@@ -0,0 +1,199 @@
...
+            },
+            LdevEntry {
+                primary: "oss2".into(),
+                failover: Some("oss1".into()),
+                label: "zfsmo-OST00012".into(),
+                device: "zfs:ost18/ost18".into(),
+            },
+            LdevEntry {
+                primary: "oss2".into(),
+                failover: Some("oss1".into()),
```
ip1981, line 183: A case without failover is missed.

iml-agent/src/action_plugins/snapshot.rs
johnsonw: My personal preference was to generate the config file in the ticket. It doesn't hurt either way though.

iml-agent/src/action_plugins/snapshot.rs
johnsonw: It could, I like using snapshots though :)

iml-agent/src/action_plugins/snapshot.rs
johnsonw: indeed, I will be adding another test to cover this.

iml-agent/src/action_plugins/action_plugin.rs
johnsonw: I thought about creating a name that had to do with `ldev` but I thought `snapshot` made more sense. The snapshots folder you are referring to is created by `insta` I believe and has unit test snapshots in it. My reasoning behind naming the file `snapshot.rs` is because the actions that live in it are related to snapshots. With that said, I'm open to changing it though.

ip1981: There will be more actions related to lustre snapshots.

I think something like `ldev_conf::create()` would be fine.

johnsonw: sure, i'll update.

iml-agent/src/action_plugins/snapshot.rs
jgrund: Snapshots can be inlined as well (not that I'm advocating for one over the other).

johnsonw:
@ip1981 just a matter of preference really. I like to look at unit tests as examples / scenarios so that's why I took the example from the ticket.

jgrund:
This needs rust formatting run on it.

johnsonw:
Odd I ran the formatting on it and pushed the changes. I guess i'll try agian.

johnsonw:
Sure, sounds good to me. My initial thinking was to put multiple actions in the snapshot.rs file:
```
create()
destroy()
unmount()
mount()
create_config()
```

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,218 @@
...
+    let mut file = File::create(ldev_path).await?;
+    file.write_all(content.as_bytes()).await?;
+
+    Ok(())
+}
+
+async fn create_internal<F>(
+    entries: Vec<LdevEntry>,
+    write_to_file: impl Fn(String) -> F,
+) -> Result<(), ImlAgentError>
```
jgrund, line 21: This fn is more complex than needed.

If you called it something like `convert`, you could pass in entries and return a `String`. Then you would not need to make this a higher-order function, and could still test it in isolation.

iml-wire-types/src/lib.rs
```diff
@@ -1727,5 +1727,57 @@ impl fmt::Display for Branding {
...
+    }
+}
+
+impl fmt::Display for LdevEntry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{} {} {} {}",
+            self.primary,
+            self.failover.as_ref().unwrap_or(&"-".to_string()),
```
jgrund, line 1775:
```suggestion
            self.failover.as_deref().unwrap_or("-"),
```

deref `String` to `&str` instead of converting `&'static str` -> `&String`

jgrund:
This has a conflict with latest master

iml-wire-types/src/lib.rs
johnsonw: I'll try changing this. I had made the change based on the compilers recommendation.

johnsonw:
I've been making updates to this patch and am now using the BTreeSet. This requires ordering to be implemented, which will cause the output to change to the following:
```
oss1 oss2 zfsmo-OST0000 zfs:ost0/ost0
oss1 oss2 zfsmo-OST0001 zfs:ost1/ost1
oss2 oss1 zfsmo-OST00010 zfs:ost16/ost16
oss2 oss1 zfsmo-OST00011 zfs:ost17/ost17
oss2 oss1 zfsmo-OST00012 zfs:ost18/ost18
oss2 oss1 zfsmo-OST00013 zfs:ost19/ost19
oss1 oss2 zfsmo-OST0002 zfs:ost2/ost2
```
I believe it's applying a lexicographical ordering based on the labels. I just want to make sure we are ok with this.

iml-agent/src/action_plugins/snapshots/iml_agent__action_plugins__ldev__tests__create.snap
```diff
@@ -0,0 +1,27 @@
...
+oss1 oss2 zfsmo-OST0007 zfs:ost7/ost7
+oss1 oss2 zfsmo-OST0008 zfs:ost8/ost8
+oss1 oss2 zfsmo-OST0009 zfs:ost9/ost9
+oss2 oss1 zfsmo-OST000a zfs:ost10/ost10
+oss2 oss1 zfsmo-OST000b zfs:ost11/ost11
+oss2 oss1 zfsmo-OST000c zfs:ost12/ost12
+oss2 oss1 zfsmo-OST000d zfs:ost13/ost13
+oss2 oss1 zfsmo-OST000e zfs:ost14/ost14
+oss2 oss1 zfsmo-OST000f zfs:ost15/ost15
+oss2 oss1 zfsmo-OST00010 zfs:ost16/ost16
```
ip1981, line 24: Oops:
```
[root@mds1 ~]# lctl snapshot_create --fsname zfsmo --name manual
Invalid snapshot config file /etc/ldev.conf at the line 20 'oss2 oss1 zfsmo-OST00010 zfs:ost16/ost16'
```

I guess it should be `zfsmo-OST0010` (minus one `0`).

johnsonw: Sure, i'll update

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,325 @@
...
+fn parse_entries(ldev_config: String) -> Result<BTreeSet<LdevEntry>, ImlAgentError> {
+    let existing_entries = ldev_config
+        .lines()
+        .map(LdevEntry::from)
+        .collect::<BTreeSet<LdevEntry>>();
+
+    Ok(existing_entries)
+}
+
+fn config_needs_update_check(
```
ip1981, line 46: Maybe just compare two strings: the one got from the file and the other rendered from the entries? :)

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,325 @@
...
+
+fn convert(entries: &[LdevEntry]) -> String {
+    entries
+        .iter()
+        .map(|x| x.to_string())
+        .collect::<Vec<String>>()
+        .join("\n")
+}
+
+pub async fn create(entries: Vec<LdevEntry>) -> Result<(), ImlAgentError> {
```
jgrund, line 66: Should check that the incoming entries is not empty, that seems like an error condition.

iml-agent/src/action_plugins/ldev.rs
johnsonw: If the order is different they won't match. I think the BTreeSet is more reliable.

ip1981: I think the order of ldev.conf doesn't matter, so we could just sort.

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,325 @@
...
+
+    Ok(())
+}
+
+async fn read_ldev_config() -> Result<String, ImlAgentError> {
+    let ldev_path = env::get_ldev_conf();
+
+    let mut file = OpenOptions::new()
+        .read(true)
+        .write(true)
```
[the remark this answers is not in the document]
johnsonw: I added it based on the following documentation:
https://doc.rust-lang.org/std/fs/struct.OpenOptions.html#method.create
> Sets the option to create a new file, or open it if it already exists.
> In order for the file to be created, write or append access must be used.

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,325 @@
...
+}
+
+fn parse_entries(ldev_config: String) -> Result<BTreeSet<LdevEntry>, ImlAgentError> {
+    let existing_entries = ldev_config
+        .lines()
+        .map(LdevEntry::from)
+        .collect::<BTreeSet<LdevEntry>>();
+
+    Ok(existing_entries)
+}
```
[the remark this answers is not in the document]
johnsonw: Looks good to me. I've updated.

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,325 @@
...
+    existing_entries: &BTreeSet<LdevEntry>,
+    entries: &BTreeSet<LdevEntry>,
+) -> bool {
+    let diff: Vec<_> = existing_entries
+        .symmetric_difference(&entries)
+        .cloned()
+        .collect();
+
+    !diff.is_empty()
+}
```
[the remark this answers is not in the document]
johnsonw: Makes sense. I'll update.

iml-agent/src/action_plugins/ldev.rs
johnsonw: Good call. I added a check.

iml-agent/src/action_plugins/ldev.rs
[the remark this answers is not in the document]
jgrund: Hrrm, given this fn is called `read_ldev_config`, should it just check if the file exists and if so read it?

I think it could be surprising that this fn does a side-effect when it's called.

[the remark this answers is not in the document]
johnsonw: It feels like an extra step to me but if you want I can add the additional check. I think the flow would be like this:
1. Check if file exists
   a. File exists
      i. Open the file and get a file handle
      i. Read the file contents into a String buffer
   b. File does not exist
      i. Return an empty string

In either case, the file will be created by the end of the action. Should I go ahead and make the change?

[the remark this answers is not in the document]
jgrund: I think it's worth doing. The file will only be completed by the end of the action if all the in-between code is successful.

As it stands now we have the (small) potential to have an empty file.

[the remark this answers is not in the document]
johnsonw: Sounds good. I'll push an update shortly.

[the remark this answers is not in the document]
jgrund: These fns should help with that btw:

https://github.com/whamcloud/integrated-manager-for-lustre/blob/a86998683c14db89779510a675a517578bf2a792/iml-fs/src/lib.rs#L184-L192

https://github.com/whamcloud/integrated-manager-for-lustre/blob/a86998683c14db89779510a675a517578bf2a792/iml-fs/src/lib.rs#L69-L85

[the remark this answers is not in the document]
johnsonw: I ended up using `read_to_string` (https://docs.rs/tokio/0.2.22/tokio/fs/fn.read_to_string.html) to avoid the mutable declaration.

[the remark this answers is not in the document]
johnsonw: Let me know if this update looks good. It should follow the flow as discussed above.

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,310 @@
...
+        Ok(_) => Ok(read_to_string(&ldev_path).await?),
+        Err(_) => Ok("".into()),
+    }
+}
+
+fn parse_entries(ldev_config: String) -> BTreeSet<LdevEntry> {
+    ldev_config.lines().map(LdevEntry::from).collect()
+}
+
+fn config_needs_update_check(
```
jgrund, line 34: nit: is this fn even worth having? It's just doing a `!=` check at this point.

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,310 @@
...
+fn convert(entries: &[LdevEntry]) -> String {
+    entries
+        .iter()
+        .map(|x| x.to_string())
+        .collect::<Vec<String>>()
+        .join("\n")
+}
+
+pub async fn create(entries: Vec<LdevEntry>) -> Result<(), ImlAgentError> {
+    if !entries.is_empty() {
```
jgrund, line 50: Shouldn't this raise an error? If this returns `Ok(())` without actually writing anything then the next steps will fail.

iml-agent/src/action_plugins/ldev.rs
johnsonw: I could see both sides but if you're questioning it then it's probably an unnecessary abstraction. I'll take it out.

iml-agent/src/action_plugins/ldev.rs
johnsonw: It will now panic! if the entries are empty.

iml-agent/src/agent_error.rs
```diff
@@ -171,6 +173,7 @@ impl std::error::Error for ImlAgentError {
             ImlAgentError::UrlParseError(ref err) => Some(err),
             ImlAgentError::Utf8Error(ref err) => Some(err),
             ImlAgentError::XmlError(ref err) => Some(err),
+            ImlAgentError::LdevEntriesError(ref err) => None,
```
jgrund, line 176:
```suggestion
            ImlAgentError::LdevEntriesError(_) => None,
```

Clippy is picking up on this

iml-agent/src/action_plugins/ldev.rs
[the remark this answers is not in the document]
jgrund: LGTM

iml-agent/src/action_plugins/ldev.rs
```diff
@@ -0,0 +1,240 @@
...
+            LdevEntry {
+                primary: "oss2".into(),
+                failover: Some("oss1".into()),
+                label: "zfsmo-OST0010".into(),
+                device: "zfs:ost16/ost16".into(),
+            },
+            LdevEntry {
+                primary: "oss2".into(),
+                failover: Some("oss1".into()),
+                label: "zfsmo-OST00011".into(),
```
ip1981, line 191: This is an invalid Id (OST00011)

iml-wire-types/src/lib.rs
```diff
@@ -1705,5 +1705,75 @@ impl fmt::Display for Branding {
...
+
+impl fmt::Display for LdevEntry {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{} {} {} {}",
+            self.primary,
+            self.failover.as_deref().unwrap_or("-"),
+            self.label,
+            self.device
```
ip1981, line 1755: Just a note: with this simple implementation tests are hardly add any value.

iml-wire-types/src/lib.rs
```diff
@@ -1705,5 +1705,75 @@ impl fmt::Display for Branding {
...
 
+#[derive(Debug, Eq, Clone, serde::Serialize, serde::Deserialize)]
+pub struct LdevEntry {
+    pub primary: String,
+    pub failover: Option<String>,
+    pub label: String,
+    pub device: String,
+}
+
+impl From<&str> for LdevEntry {
```
ip1981, line 1716: Note: this would need tests :) At least more need than writing.

PR #2102, opened 2020-07-23
tests: Mount filesystem(s) during detect_fs

utopiabound:
vagrant: Split mount from create for zfs backed fs
test: Check that the correct number of filesystems
were detected in detect_fs()

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2102)

### DIFF
--- a/iml-system-test-utils/src/ssh.rs
+++ b/iml-system-test-utils/src/ssh.rs
@@ -268,6 +268,17 @@ pub async fn detect_fs(host: &str) -> Result<(), CmdError> {
         .await
 }
 
+pub async fn list_fs_json(host: &str) -> Result<Vec<serde_json::Value>, CmdError> {
+    ssh_exec_cmd(host, "iml filesystem list -p json")
+        .await?
+        .checked_output()
+        .await
+        .map(|o| {
+            let v: Vec<serde_json::Value> = serde_json::from_slice(&o.stdout).unwrap();
+            v
+        })
+}
+
 pub async fn systemd_status(host: &str, service_name: &str) -> Result<Command, CmdError> {
     let cmd = ssh_exec_cmd(host, format!("systemctl status {}", service_name).as_str()).await?;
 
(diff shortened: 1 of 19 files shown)


iml-system-test-utils/src/ssh.rs
```diff
@@ -268,6 +268,17 @@ pub async fn detect_fs(host: &str) -> Result<(), CmdError> {
         .await
 }
 
+pub async fn list_fs_json(host: &str) -> Result<Vec<serde_json::Value>, CmdError> {
+    ssh_exec_cmd(host, "iml filesystem list -p json")
+        .await?
+        .checked_output()
+        .await
+        .map(|o| {
+            let v: Vec<serde_json::Value> = serde_json::from_slice(&o.stdout).unwrap();
```
johnsonw, line 277: Just curious, could it also be:
```
serde_json::from_slice(&o.stdout)l.unwrap::<Vec<serde_json::Value>>()
```
?

utopiabound: I tried cajoling that line to do that, but I couldn't figure it out:
```
error[E0107]: wrong number of type arguments: expected 0, found 1
   --> iml-system-test-utils/src/ssh.rs:279:61
    |
279 |         .map(|o| serde_json::from_slice(&o.stdout).unwrap::<Vec<serde_json::Value>>())
    |                                                             ^^^^^^^^^^^^^^^^^^^^^^ unexpected type argument

```

jgrund: The reason that doesn't work is `unwrap` has no generic parameter to fill in.

The turbofish needs to be on `from_slice` which does have a generic parameter:

https://docs.serde.rs/serde_json/fn.from_slice.html

iml-system-test-utils/src/ssh.rs
```diff
@@ -268,6 +268,17 @@ pub async fn detect_fs(host: &str) -> Result<(), CmdError> {
...
+            v
+        })
```
jgrund, line 282:
```suggestion
        .map(|o| serde_json::from_slice::<Vec<serde_json::Value>>(&o.stdout).unwrap())
```

PR #2103, opened 2020-07-23
Make docker image tarball a required file

jgrund:
`iml-images.tgz` currently lives under /var/tmp
and is marked as ok if it's missing.

This can cause an issue if the tarball is removed, as `iml-docker` will
not be able to load the images and startup.

Move the tarball to `/var/lib`, and make it non-optional.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2103)

### DIFF
--- a/iml-docker.spec
+++ b/iml-docker.spec
@@ -37,8 +37,8 @@ mv iml-docker.service %{buildroot}%{_unitdir}
 
 %files
 %{_sysconfdir}/iml-docker
-%attr(750, root, root) %{_sharedstatedir}/iml-images.tgz
-%attr(754, root, root) %{_bindir}/iml
+%attr(0640, root, root) %{_sharedstatedir}/iml-images.tgz
+%attr(0755, root, root) %{_bindir}/iml
 %attr(754, root, root) %{_bindir}/update-embedded
 %attr(0644, root, root) %{_unitdir}/iml-docker.service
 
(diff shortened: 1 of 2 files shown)


iml-docker.spec
```diff
@@ -25,22 +25,22 @@ Requires: sed
...
+mv iml-images.tgz %{buildroot}%{_sharedstatedir}
 mv iml-cli-proxy.sh %{buildroot}%{_bindir}/iml
 mv update-embedded.sh %{buildroot}%{_bindir}/update-embedded
 mv iml-docker.service %{buildroot}%{_unitdir}
 
 
 %files
 %{_sysconfdir}/iml-docker
-%attr(750, root, root) %config(missingok) %{_tmppath}/iml-images.tgz
+%attr(750, root, root) %{_sharedstatedir}/iml-images.tgz
```
utopiabound, line 40: The attr be 0640

and the blow should be 0755

iml-docker.spec
```diff
@@ -25,22 +25,22 @@ Requires: sed
...
 %attr(754, root, root) %{_bindir}/iml
```
jgrund, line 41:
```suggestion
%attr(755, root, root) %{_bindir}/iml
```

iml-docker.spec
jgrund, line 40:
```suggestion
%attr(0640, root, root) %{_sharedstatedir}/iml-images.tgz
```

PR #2107, opened 2020-07-27
Build up targets from device-scanner output

jgrund:
The device-scanner has all the needed info to determine targets at
runtime.

This patch finds those targets and figures out all the hosts they can be
mounted on. We can then persist that info to the database and use it
instead of having to detect manually.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2107)


johnsonw:
This video highlights the following details:
1. The chroma_core_targets table is updated to reflect the active host and failover host information when the cluster is put into standby mode.
2. When a target is unmounted, the table will be updated such that the:
   a. The `mount_path` column is set to NULL
   b. The `active_host_id` column is set to NULL
   c. The `state` column is set to "unmounted"

targets_failover

johnsonw:
Latest tests running here: https://build.whamcloud.com/job/manager-for-lustre/1657/

device-scanner/device-types/src/devices.rs
```diff
@@ -164,3 +164,185 @@ pub enum Device {
     Zpool(Zpool),
     Dataset(Dataset),
 }
+
+#[derive(Debug, serde::Serialize, Eq, PartialEq, Ord, PartialOrd, Clone, Hash)]
```
jgrund, line 168: Can we split this out to it's own patch?

iml-services/iml-device/Cargo.toml
```diff
@@ -5,7 +5,7 @@ name = "iml-device"
 version = "0.3.0"
 
 [dependencies]
-device-types = "0.1"
+device-types = {path = "../../device-scanner/device-types", version = "0.1"}
```
jgrund, line 8: Once the above is it's own patch, we can publish it and bump the version here.

jgrund:
LGTM

PR #2122, opened 2020-07-31
action-client: Add invoke_rust_agent

utopiabound:
Add invoke_rust_agent() to rust code.
Add cancel-able variant.

Related Issue #1832 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2122)

### DIFF
--- a/iml-action-client/Cargo.toml
+++ b/iml-action-client/Cargo.toml
@@ -5,10 +5,15 @@ authors = ["IML Team <iml@whamcloud.com>"]
 edition = "2018"
 
 [dependencies]
+bytes = "0.5"
 futures = "0.3"
 iml-action-runner = { path = "../iml-services/iml-action-runner", version = "0.3" }
 iml-manager-env = { path = "../iml-manager-env", version = "0.3" }
 iml-wire-types = { path = "../iml-wire-types", version = "0.3" }
+hyper = { version = "0.13", default-features = false, features = ["tcp"] }
+# this can be used once hyperlocal#36 lands
+#hyperlocal = { version = "0.7", features = [ "client" ], default_features = false }
+hyperlocal = { version = "0.7" }
 reqwest = { version = "0.10", default-features = false, features = ["rustls-tls", "json", "stream"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
--- a/iml-action-client/Cargo.toml
+++ b/iml-action-client/Cargo.toml
@@ -18,4 +18,5 @@ reqwest = { version = "0.10", default-features = false, features = ["rustls-tls"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 thiserror = "1.0"
+tracing = "0.1"
 uuid = { version = "0.8", features = ["v4"] }
--- a/iml-action-client/src/lib.rs
+++ b/iml-action-client/src/lib.rs
@@ -52,7 +52,7 @@ fn client() -> ClientWrapper {
     }
 }
 
-// Return URI or painc
+// Return URI or panic
 fn connect_uri() -> hyper::Uri {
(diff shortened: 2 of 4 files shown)


iml-action-client/src/lib.rs
```diff
@@ -2,35 +2,75 @@
 // Use of this source code is governed by a MIT-style
 // license that can be found in the LICENSE file.
 
+use bytes::buf::BufExt as _;
```
jgrund, line 5: Add iml-tracing to this crate and log in debug for interesting points (probably request / responses).

There are times when debugging code that knowing payloads / responses are useful.

iml-action-client/src/lib.rs
```diff
@@ -2,35 +2,75 @@
...
+    if running_in_docker() {
+        ClientWrapper::Http(Client::new())
+    } else {
+        ClientWrapper::Unix(Client::unix())
+    }
+}
+
+fn connect_uri() -> Result<hyper::Uri, ImlActionClientError> {
+    if running_in_docker() {
+        get_action_runner_http()
```
jgrund, line 58: This is going to do a env lookup each for each connection, and then parse it to a Uri. this fn should instead return a Uri directly or panic if it can't be converted.

iml-action-client/src/lib.rs
```diff
@@ -41,38 +81,87 @@ pub fn invoke_rust_agent(
...
+    let client = client();
+    let request_id = Uuid::new_v4().to_hyphenated().to_string();
+
+    let host2 = host.clone();
+    let client2 = client.clone();
+    let request_id2 = request_id.clone();
+
+    let post = build_invoke_rust_agent(host, command, args, client, request_id);
+
+    let (p, c) = oneshot::channel::<()>();
```
jgrund, line 135: Consider https://docs.rs/futures/0.3.5/futures/future/fn.abortable.html here instead.

iml-action-client/src/lib.rs
```diff
@@ -2,35 +2,73 @@
...
+
+fn client() -> ClientWrapper {
+    if running_in_docker() {
+        ClientWrapper::Http(Client::new())
+    } else {
+        ClientWrapper::Unix(Client::unix())
+    }
+}
+
+// Return URI or painc
```
jgrund, line 55:
```suggestion
// Return URI or panic
```

iml-manager-env/src/lib.rs
```diff
@@ -247,16 +255,12 @@ pub fn get_use_stratagem() -> bool {
     string_to_bool(get_var("USE_STRATAGEM"))
 }
 
-pub fn get_action_runner_connect() -> String {
```
jgrund, line 250: nit: this was probably the cleaner approach

utopiabound: I need to use different  URI parsers (`hyper` and `hyperlocal`) in `iml-action-client`, so these needed to be separated.

iml-action-client/src/lib.rs
ip1981: Return URI or pain :)

iml-action-client/Cargo.toml
```diff
@@ -5,11 +5,17 @@ authors = ["IML Team <iml@whamcloud.com>"]
 edition = "2018"
 
 [dependencies]
+bytes = "0.5"
 futures = "0.3"
 iml-manager-env = { path = "../iml-manager-env", version = "0.3" }
 iml-wire-types = { path = "../iml-wire-types", version = "0.3" }
+hyper = { version = "0.13", default-features = false, features = ["tcp"] }
+# this can be used once hyperlocal#36 lands
```
johnsonw, line 13: Should we create a ticket as a reminder?

PR #2123, opened 2020-07-31
Extract remote_action code from sender

jgrund:
Currently, the only way to run a `remote_action` is to call it through
the `action-runner`.

This means even if there is code within the `action-runner` that wants
to call a local action it will need to schedule it via HTTPS / UDS.

As a first step to making this more efficient, move the `remote_action`
code out to a directly callable fn.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2123)


jgrund:
> Why do we do this?

Connections are closed when there are no more handles to them. This is a precaution to make sure we don't keep the handle open longer than necessary.

PR #2124, opened 2020-07-31
Use a single lustre install step

### LINKED ISSUE #2119: Create a LustreRPMsInstalled state
If we create a `LustreRPMsInstalled` state after the `Base` state, we can use that state for the start of all monitored testing.

The `LustreRPMsInstalled` state would require a provisioner: `install_ldiskfs_zfs_no_iml` that would just install Lustre + ZFS RPMs.

image

jgrund:
We should be able to do a single lustre install step for each monitored
test variant.

This should shave a bunch of time off test runs, as lustre will be
installed via snapshot for all branches.

Fixes #2119.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2124)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,12 +43,18 @@ local:
 check:
 	black --check ./
 	cargo fmt --all -- --check
-	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo check
+	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo check --locked --all-targets
 	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo clippy -- -W warnings
+	cargo check --locked --manifest-path iml-system-rpm-tests/Cargo.toml --tests
+	cargo clippy --manifest-path iml-system-rpm-tests/Cargo.toml --tests -- -W warnings
+	cargo check --locked --manifest-path iml-system-docker-tests/Cargo.toml --tests
+	cargo clippy --manifest-path iml-system-docker-tests/Cargo.toml --tests -- -W warnings
 
 fmt:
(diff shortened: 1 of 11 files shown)


jgrund:
graphviz (2)

jgrund:
Looks like this change saves ~17 minutes per run. I'll look into parallelizing more vagrant commands in a separate patch

Makefile
```diff
@@ -43,12 +43,18 @@ local:
 check:
 	black --check ./
 	cargo fmt --all -- --check
-	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo check
+	PQ_LIB_DIR=/usr/pgsql-9.6/lib cargo check --locked --all-targets
```
nlinker, line 46: is this to renew `Cargo.lock` file?

jgrund: This checks the code has no errors. The `--locked` flag ensures the Cargo.lock file is up to date. The  `--all-targets` flag ensures the check is run for all targets.

nlinker: Thanks

iml-system-test-utils/src/snapshots.rs
```diff
@@ -67,46 +66,16 @@ pub enum SnapshotName {
     StratagemDetected,
 }
 
-impl SnapshotName {
```
johnsonw, line 70: The ordering isn't needed?

jgrund: The ordering is still there, it's just derived by the compiler instead of us having to write it by hand.

PR #2131, opened 2020-08-04
Fix cargo formatting of system tests

utopiabound:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2131)


jgrund:
We should add a check for this to github actions so it gets caught in review

PR #2132, opened 2020-08-05
Add 'iml debugapi' to iml cli

utopiabound:
`debugapi` does not show up in help as it's for test/debug

Related #1832 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2132)

### DIFF
--- a/iml-manager-cli/src/api.rs
+++ b/iml-manager-cli/src/api.rs
@@ -4,9 +4,9 @@
 
 use crate::error::ImlManagerCliError;
 use console::Term;
-use structopt::{StructOpt, clap::arg_enum};
+use structopt::{clap::arg_enum, StructOpt};
 
-arg_enum!{
+arg_enum! {
     #[derive(PartialEq, Debug)]
     pub enum ApiType {
         Delete,
@@ -14,10 +14,6 @@ arg_enum!{
         Post,
         Put,
     }
-}
-
-#[derive(Debug, StructOpt)]
-pub struct ApiOpts {
 }
 
 #[derive(Debug, StructOpt)]
--- a/iml-manager-cli/src/api.rs
+++ b/iml-manager-cli/src/api.rs
@@ -23,15 +23,15 @@ pub struct ApiCommand {
 
     path: String,
 
-    /// Object to PUT
-    object: Option<String>,
+    /// Object to PUT or POST
+    body: Option<String>,
 }
 
 pub async fn api_cli(command: ApiCommand) -> Result<(), ImlManagerCliError> {
     let term = Term::stdout();
     let client = iml_manager_client::get_client()?;
     let uri = iml_manager_client::create_api_url(command.path)?;
-    let body = command.object.unwrap_or("{}".to_string());
+    let body = command.body.unwrap_or("{}".to_string());
 
     if let Some(resp) = match command.call {
         ApiType::Delete => Some(client.delete(uri).send().await?),
--- a/iml-manager-cli/src/api.rs
+++ b/iml-manager-cli/src/api.rs
@@ -24,6 +24,7 @@ pub struct ApiCommand {
     path: String,
 
     /// JSON formatted body to send
+    #[structopt(required_ifs(&[("call", "post"),("call", "put")]))]
     body: Option<String>,
 }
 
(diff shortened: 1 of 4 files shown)


utopiabound:
```
[root@adm ~]# iml debugapi get target
{"meta":{"limit":20,"next":null,"offset":0,"previous":null,"total_count":0},"objects":[]}
```

ip1981:
How is it better then curl? :) 

```
# curl http://127.0.0.1:8001/api/filesystem/
{"meta": {"limit": 20, "next": null, "offset": 0, "previous": null, "total_count": 1}, "objects": [{"cdt_mdt": "/api/target/2/", ...
```

iml-manager-cli/src/api.rs
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#[derive(Debug, StructOpt)]
+pub struct ApiCommand {
+    #[structopt(possible_values = &ApiType::variants(), case_insensitive = true)]
+    call: ApiType,
+
+    path: String,
+
+    /// Object to PUT
```
ip1981, line 26: nit: could be any method actually, at least POST and PUT. Better be called `body`.

iml-manager-cli/src/main.rs
```diff
@@ -35,8 +36,12 @@ pub enum App {
         command: filesystem::FilesystemCommand,
     },
     #[structopt(name = "update_repo")]
-    ///  Update Agent repo files
+    /// Update Agent repo files
     UpdateRepoFile(update_repo_file::UpdateRepoFileHosts),
+
+    #[structopt(name = "debugapi", setting = structopt::clap::AppSettings::Hidden)]
```
ip1981, line 42: Maybe shorter `api` ?

jgrund:
> How is it better than curl? :)

API may be setup as read-only and certain verbs (`POST`, `PUT`, ...) are usually protected via auth

iml-manager-cli/src/main.rs
utopiabound: I want it clear this isn't meant as a user facing, it's for debug and testing.

iml-manager-cli/src/api.rs
```diff
@@ -23,15 +23,15 @@ pub struct ApiCommand {
 
     path: String,
 
-    /// Object to PUT
-    object: Option<String>,
+    /// Object to PUT or POST
```
jgrund, line 26: Or PATCH :)

utopiabound:
POST tested and works

iml-manager-cli/src/api.rs
```diff
@@ -0,0 +1,52 @@
...
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::error::ImlManagerCliError;
+use console::Term;
+use structopt::{clap::arg_enum, StructOpt};
+
+arg_enum! {
+    #[derive(PartialEq, Debug)]
+    pub enum ApiType {
```
jgrund, line 11: Verb probably makes the most sense here. Also should include patch

iml-manager-cli/src/api.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+pub async fn api_cli(command: ApiCommand) -> Result<(), ImlManagerCliError> {
+    let term = Term::stdout();
+    let client = iml_manager_client::get_client()?;
+    let uri = iml_manager_client::create_api_url(command.path)?;
+    let body: serde_json::Value = serde_json::from_str(&command.body.unwrap_or("{}".to_string()))?;
+
+    if let Some(resp) = match command.call {
+        ApiType::Delete => Some(client.delete(uri).json(&body).send().await?),
```
jgrund, line 37: What happens if you don't want to send a body?

iml-manager-cli/src/api.rs
```diff
@@ -24,25 +24,38 @@ pub struct ApiCommand {
...
 
 pub async fn api_cli(command: ApiCommand) -> Result<(), ImlManagerCliError> {
     let term = Term::stdout();
     let client = iml_manager_client::get_client()?;
     let uri = iml_manager_client::create_api_url(command.path)?;
-    let body: serde_json::Value = serde_json::from_str(&command.body.unwrap_or("{}".to_string()))?;
+    let body: Option<serde_json::Value> = if let Some(data) = command.body {
+        Some(serde_json::from_str(&data)?)
+    } else {
+        None
```
ip1981, line 38: looks like `command.body.map(serde_json::from_str)`, no?

iml-manager-cli/src/api.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// JSON formatted body to send
+    #[structopt(required_ifs(&[("call", "post"),("call", "put")]))]
+    body: Option<String>,
+}
+
+pub async fn api_cli(command: ApiCommand) -> Result<(), ImlManagerCliError> {
+    let term = Term::stdout();
+    let client = iml_manager_client::get_client()?;
+    let uri = iml_manager_client::create_api_url(command.path)?;
+    let body: Option<serde_json::Value> = command.body.map(|s| serde_json::from_str(&s).unwrap());
```
ip1981, line 35: oh, I missed double Option. `.and_then()` would be the best then :)

utopiabound: Panic from a test tool if you use it wrong, I think, is okay.  Actually better than silently failing and setting body to `None`

jgrund: It doesn't really matter in this case, but https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.transpose would be useful to easily avoid the unwrap.

iml-manager-cli/src/api.rs
```diff
@@ -0,0 +1,71 @@
...
+    let client = iml_manager_client::get_client()?;
+    let uri = iml_manager_client::create_api_url(command.path)?;
+
+    let req = match command.method {
+        ApiMethod::Delete => client.delete(uri),
+        ApiMethod::Get => client.get(uri),
+        ApiMethod::Head => client.head(uri),
+        ApiMethod::Patch => client.patch(uri),
+        ApiMethod::Post => client.post(uri),
+        ApiMethod::Put => client.put(uri),
```
nlinker, line 44: Headers and bodies are not supported, is it intentional?

jgrund: Bodies are supported here: https://github.com/whamcloud/integrated-manager-for-lustre/pull/2132/files#diff-f33d40dcab0b4085abfc004fe3408e29R50-R54

Headers generally aren't needed (currently) when using the ```iml_manager_client``

PR #2156, opened 2020-08-13
Implement MountSnapshotJob / UnmountSnapshotJob

mkpankov:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2156)


mkpankov:
It works like this:
```
[root@mds1 vagrant]# lctl snapshot_list -F zfsmo 

filesystem_name: zfsmo
snapshot_name: test1
snapshot_fsname: 1494052f 
create_time: Mon Jul 27 14:09:53 2020
modify_time: Mon Jul 27 14:09:53 2020
status: not mount
```

```
➜  ~ jarg message="Test command" 'jobs[0][class_name]=MountSnapshotJob' 'jobs[0][args][hosts][0]=/api/host/1/' 'jobs[0][args][fsname]=zfsmo' 'jobs[0][args][name]=test1' | http --verbose --json --verify=false https://localhost:8443/api/command/ authorization:'ApiKey api:15ec8aca8b4139ad741e4bbf9157b1e193a07a52'
POST /api/command/ HTTP/1.1
Accept: application/json, */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Content-Length: 146
Content-Type: application/json
Host: localhost:8443
User-Agent: HTTPie/1.0.3
authorization: ApiKey api:15ec8aca8b4139ad741e4bbf9157b1e193a07a52

{
    "jobs": [
        {
            "args": {
                "fsname": "zfsmo",
                "hosts": [
                    "/api/host/1/"
                ],
                "name": "test1"
            },
            "class_name": "MountSnapshotJob"
        }
    ],
    "message": "Test command"
}

HTTP/1.1 201 Created
Connection: keep-alive
Content-Length: 212
Content-Type: application/json
Date: Thu, 13 Aug 2020 10:59:33 GMT
Location: /api/command/30/
Server: nginx/1.16.1
Vary: Accept
X-Frame-Options: SAMEORIGIN

{
    "cancelled": false,
    "complete": false,
    "created_at": "2020-08-13T10:59:33.597354",
    "errored": false,
    "id": 30,
    "jobs": [
        "/api/job/42/"
    ],
    "logs": "",
    "message": "Test command",
    "resource_uri": "/api/command/30/"
}
```

```
[root@mds1 vagrant]# lctl snapshot_list -F zfsmo 

filesystem_name: zfsmo
snapshot_name: test1
snapshot_fsname: 1494052f 
create_time: Mon Jul 27 14:09:53 2020
modify_time: Mon Jul 27 14:09:53 2020
status: mounted

```

mkpankov:
@jgrund dependencies and locks are missing, but I wanted to clarify if the approach taken is correct with respect to "DB calls should be made in Rust" as specified in the issue.

Since the state machine stays in Python and we need to save the `Job` itself, I guess these don't count as DB calls?

chroma_core/models/host.py
```diff
@@ -1816,3 +1816,29 @@ def run(self, kwargs):
         self.invoke_rust_agent_expect_result(
             kwargs["host"], "snapshot_unmount", {"fsname": kwargs["fsname"], "name": kwargs["name"]}
         )
+
+
+class MountSnapshotJob(HostListMixin):
+    fsname = models.CharField(max_length=512)
```
ip1981, line 1822: I think FS name is limited to 8 chars, so there might not be need to reserve more.

chroma_core/models/host.py
```diff
@@ -1816,3 +1816,29 @@ def run(self, kwargs):
...
+    def description(self):
+        if len(self.hosts) > 1:
+            return "Mount snapshot on %d hosts" % len(self.hosts)
+        else:
+            return "Mount snapshot on host %s" % self.hosts[0]
+
+    def get_steps(self):
+        steps = []
+        for host in self.hosts:
+            steps.append((MountSnapshotStep, {"host": "%s" % host, "fsname": self.fsname, "name": self.name}))
```
ip1981, line 1838: Can it be just `"host": host ...`?

jgrund: hosts.fqdn would be clearer than coercing to a string

jgrund:
> Since the state machine stays in Python and we need to save the Job itself, I guess these don't count as DB calls?

Yes, I'm more talking about associated queries needed to get related objects.

chroma_core/models/host.py
mkpankov: Done

chroma_core/models/host.py
```diff
@@ -1816,3 +1816,28 @@ def run(self, kwargs):
...
+        return help_text["mount_snapshot"]
+
+    def description(self):
+        if len(self.hosts) > 1:
+            return "Mount snapshot on %d hosts" % len(self.hosts)
+        return "Mount snapshot on host %s" % self.hosts[0]
+
+    def get_steps(self):
+        steps = []
+        for host in self.hosts:
```
jgrund, line 1836: Would this ever need to run on more than a single host?

chroma_core/models/host.py
mkpankov: Done

chroma_core/models/host.py
mkpankov: Scrapped that

mkpankov:
I've tested this with non-imported pools and no dependencies were necessary besides `ldev.conf`.

jgrund:
> I've tested this with non-imported pools and no dependencies were necessary besides ldev.conf.

At the very least the filesystem needs to be in the available state. Mounting a snapshot should not work if the fs is unavailable I think

mkpankov:
So, I've retested everything.
1. Mounting existing snapshot doesn't actually require running lustre FS. It works after ZFS pools are imported with `import-pools`.
2. With `DependOn(ManagedFilesystem.objects.get(name=self.fsname), "available")` as in current version, it does require a running lustre FS that is detected by IML.

chroma_core/models/host.py
```diff
@@ -1816,3 +1875,59 @@ def run(self, kwargs):
...
+        return help_text["mount_snapshot"]
+
+    def description(self):
+        return "Mount snapshot on host %s" % self.fqdn
+
+    def get_deps(self):
+        # To prevent circular imports
+        from chroma_core.models.filesystem import ManagedFilesystem
+
+        return DependOn(ManagedFilesystem.objects.get(name=self.fsname), "available")
```
jgrund, line 1896: We'll want to check if this is needed for ldiskfs since it's apparently not needed with zfs

chroma_help/help.py
```diff
@@ -196,4 +196,8 @@
     "revoke_ticket": "Revoking Ticket",
     "forget_ticket": "Forgetting Ticket",
     "create_task": "Creating Task",
+    "mount_snapshot": "Mounting Snapshot",
```
johnsonw, line 199: Not really a big deal but the long descriptions are shorter than the standard descriptions. You could store the values that you have under `description` in this help.py file and then call them:
```
 @classmethod
    def long_description(cls, stateful_object):
        return help_text["unmount_snapshot"] % stateful_object.fqdn

    def description(self):
        return self.long_description(self)
```

PR #2158, opened 2020-08-13
Make db pool size configurable

### LINKED ISSUE #2157: iml-task-runner: limit parallelism to number of connections in pool
In iml-task-runner, we do some work each tick (5 seconds) and ship it off to a separate task. Within that task, we find some fid actions and run them concurrently. Each block of fid actions is run within a transaction, so we can rollback if there is a failure of any kind. 

The amount of parallel fid actions in flight is bounded by both the number of usable workers and the db pool used for connections.

If workers are < the pool size, then when all workers are in flight there are still open connections in the pool and things work well.

If workers are >= the pool size, then we will attempt to schedule work on each tick. Each block of work will attempt to get a transaction, which will block until a connection is available. 

The connection timeout is currently set to 30 seconds; once a transaction blocks, it has this long for a previous transaction to be released or an error will be raised.

To complicate matters further, Postgres has a bounded number of total connections. Once they are all in use, all processes connected to the database will start erroring. This value is set to 100 in a default install:

```
chroma=> SHOW max_connections;
 max_connections
-----------------
 100
(1 row)
```

Given actions can be run for an undefined amount of time; we need to hold a connection open for the entire lifetime of an action, and we have a limited number of connections, we should bound parallelism of transactions to the amount of connections available in the pool.

By doing this we avoid buffering work that has 30 seconds to get a connection or suffer an error.

It should be noted that the transactional approach (which seems to be the correct one) trades performance for reliability; If we were to remove the transaction requirement I think we could keep the current approach.

utopiabound:
Allow override of POOL_LIMIT on a per service basis
Adjust most limits down
Adjust task-runner limit up to 10

Fixes #2157

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2158)

### DIFF
--- a/iml-api/src/main.rs
+++ b/iml-api/src/main.rs
@@ -13,7 +13,7 @@ use iml_rabbit::{self, create_connection_filter};
 use iml_wire_types::Conf;
 use warp::Filter;
 
-// Default pool limit if not overwridden by POOL_LIMIT
+// Default pool limit if not overridden by POOL_LIMIT
 const POOL_LIMIT: u32 = 5;
 
 #[tokio::main]
--- a/iml-api/src/main.rs
+++ b/iml-api/src/main.rs
@@ -7,10 +7,14 @@ mod command;
 mod error;
 mod task;
 
+use iml_manager_env::get_pool_limit;
 use iml_postgres::get_db_pool;
 use iml_rabbit::{self, create_connection_filter};
 use iml_wire_types::Conf;
 use warp::Filter;
+
+// Default pool limit if not overwridden by POOL_LIMIT
+const POOL_LIMIT: u32 = 5;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -33,7 +37,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let conn_filter = create_connection_filter(pool);
 
-    let pool = get_db_pool(5).await?;
+    let pool = get_db_pool(get_pool_limit().unwrap_or(POOL_LIMIT)).await?;
     let db_pool_filter = warp::any().map(move || pool.clone());
 
     let routes = warp::path("conf")
--- a/iml-api/src/main.rs
+++ b/iml-api/src/main.rs
@@ -14,7 +14,7 @@ use iml_wire_types::Conf;
 use warp::Filter;
 
 // Default pool limit if not overridden by POOL_LIMIT
-const POOL_LIMIT: u32 = 5;
+const DEFAULT_POOL_LIMIT: u32 = 5;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
@@ -37,7 +37,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let conn_filter = create_connection_filter(pool);
 
-    let pool = get_db_pool(get_pool_limit().unwrap_or(POOL_LIMIT)).await?;
+    let pool = get_db_pool(get_pool_limit().unwrap_or(DEFAULT_POOL_LIMIT)).await?;
     let db_pool_filter = warp::any().map(move || pool.clone());
 
     let routes = warp::path("conf")
--- a/iml-mailbox/src/main.rs
+++ b/iml-mailbox/src/main.rs
@@ -18,7 +18,7 @@ use lazy_static::lazy_static;
 use std::pin::Pin;
 use warp::Filter as _;
 
-// Default pool limit if not overwridden by POOL_LIMIT
+// Default pool limit if not overridden by POOL_LIMIT
 lazy_static! {
     static ref POOL_LIMIT: u32 = get_pool_limit().unwrap_or(8);
 }
@@ -27,7 +27,7 @@ lazy_static! {
 async fn main() -> Result<(), Box<dyn std::error::Error>> {
     iml_tracing::init();
 
-    let pool = get_db_pool(POOL_LIMIT).await?;
+    let pool = get_db_pool(*POOL_LIMIT).await?;
     let db_pool_filter = warp::any().map(move || pool.clone());
 
     let addr = iml_manager_env::get_mailbox_addr();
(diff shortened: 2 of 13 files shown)


iml-mailbox/src/main.rs
```diff
@@ -11,16 +11,23 @@
...
 use futures::{Stream, StreamExt};
 use iml_mailbox::ingest_data;
+use iml_manager_env::get_pool_limit;
 use iml_postgres::{get_db_pool, sqlx::PgPool};
 use iml_tracing::tracing;
+use lazy_static::lazy_static;
 use std::pin::Pin;
 use warp::Filter as _;
 
+// Default pool limit if not overwridden by POOL_LIMIT
```
jgrund, line 21:
```suggestion
// Default pool limit if not overridden by POOL_LIMIT
```

iml-api/src/main.rs
```diff
@@ -7,11 +7,15 @@ mod command;
 mod error;
 mod task;
 
+use iml_manager_env::get_pool_limit;
 use iml_postgres::get_db_pool;
 use iml_rabbit::{self, create_connection_filter};
 use iml_wire_types::Conf;
 use warp::Filter;
 
+// Default pool limit if not overwridden by POOL_LIMIT
```
jgrund, line 16:
```suggestion
// Default pool limit if not overridden by POOL_LIMIT
```

iml-services/iml-action-runner/src/main.rs
```diff
@@ -16,6 +17,9 @@ use warp::{self, Filter as _};
 
 pub static AGENT_TX_RUST: &str = "agent_tx_rust";
 
+// Default pool limit if not overwridden by POOL_LIMIT
```
jgrund, line 20:
```suggestion
// Default pool limit if not overridden by POOL_LIMIT
```

iml-services/iml-device/src/main.rs
```diff
@@ -23,13 +24,16 @@ use std::{
 };
 use warp::Filter;
 
+// Default pool limit if not overwridden by POOL_LIMIT
```
jgrund, line 27:
```suggestion
// Default pool limit if not overridden by POOL_LIMIT
```

iml-services/iml-journal/src/main.rs
```diff
@@ -19,6 +20,9 @@ lazy_static! {
     static ref DBLOG_LW: i64 = iml_manager_env::get_dblog_lw() as i64;
 }
 
+// Default pool limit if not overwridden by POOL_LIMIT
```
jgrund, line 23:
```suggestion
// Default pool limit if not overridden by POOL_LIMIT
```

iml-services/iml-ntp/src/main.rs
```diff
@@ -3,15 +3,19 @@
 // license that can be found in the LICENSE file.
 
 use futures::TryStreamExt;
+use iml_manager_env::get_pool_limit;
 use iml_postgres::{alert, get_db_pool, sqlx};
 use iml_service_queue::service_queue::consume_data;
 use iml_wire_types::{db::ManagedHostRecord, time::State, AlertRecordType, AlertSeverity};
 
+// Default pool limit if not overwridden by POOL_LIMIT
```
jgrund, line 11:
```suggestion
// Default pool limit if not overridden by POOL_LIMIT
```

iml-task-runner/src/main.rs
```diff
@@ -27,6 +28,8 @@ pub mod error;
 const FID_LIMIT: i64 = 2000;
 // Number of seconds between cycles
 const DELAY: Duration = Duration::from_secs(5);
+// Default pool limit if not overwridden by POOL_LIMIT
```
jgrund, line 31:
```suggestion
// Default pool limit if not overridden by POOL_LIMIT
```

iml-task-runner/src/main.rs
```diff
@@ -28,6 +31,11 @@ const FID_LIMIT: i64 = 2000;
 // Number of seconds between cycles
 const DELAY: Duration = Duration::from_secs(5);
 
+// Default pool limit if not overridden by POOL_LIMIT
+lazy_static! {
+    static ref POOL_LIMIT: u32 = get_pool_limit().unwrap_or(8);
```
jgrund, line 36: Why 8 instead of 10?

iml-services/iml-action-runner/src/main.rs
```diff
@@ -16,6 +17,9 @@ use warp::{self, Filter as _};
 
 pub static AGENT_TX_RUST: &str = "agent_tx_rust";
 
+// Default pool limit if not overridden by POOL_LIMIT
+const POOL_LIMIT: u32 = 2;
```
ip1981, line 21:
```suggestion
const DEFAULT_POOL_LIMIT: u32 = 2;
```

:)

iml-task-runner/src/main.rs
utopiabound: Targets tend to be multiples of 4, so 8 seemed to fit better.

iml-mailbox/src/main.rs
```diff
@@ -11,16 +11,23 @@
...
 use iml_mailbox::ingest_data;
+use iml_manager_env::get_pool_limit;
 use iml_postgres::{get_db_pool, sqlx::PgPool};
 use iml_tracing::tracing;
+use lazy_static::lazy_static;
 use std::pin::Pin;
 use warp::Filter as _;
 
+// Default pool limit if not overridden by POOL_LIMIT
+lazy_static! {
```
ip1981, line 22: Why lazy_static?

utopiabound: So it's only eval'd once, and this allows it to be used inside the `async move` without passing in as an extra filter.

jgrund:
I've tested this on google cloud with 65 mounted clients and 66 servers. All fids were processed and active_clients were empty afterwards

```
-[ RECORD 1 ]---+-----------------------------------------------------------------------------------
id              | 7
name            | 38303150-094b-4dd1-abf0-5fca2948b0a7-warn_fids-fids_expiring_soon
start           | 2020-08-15 00:30:22.519986+00
finish          | 2020-08-15 00:31:02.172049+00
state           | removed
fids_total      | 216301
fids_completed  | 216301
fids_failed     | 0
data_transfered | 0
single_runner   | f
keep_failed     | f
actions         | {stratagem.warning}
args            | {"report_name": "expiring_fids-exacloud-38303150-094b-4dd1-abf0-5fca2948b0a7.txt"}
filesystem_id   | 1
running_on_id   |
```

PR #2160, opened 2020-08-17
Add ManagedFilesystem.get_server_groups()

### LINKED ISSUE #2139: HA Clusters
Instead of `managedtarget.ha_cluster_peers` it would be better to have a list HA clusters with member hosts.

The use case for this is to find 1 host from each cluster for a given filesystem to run resource creates on.

Given the current setup: finding HA Groups for a filesystem could be accomplished by mapping all ManagedTargetMounts (grouped by target_id) to groups of MangedHosts and then finding unique sets.

utopiabound:
Assemble a list of server groups in roughly O(get_servers())

Fixes #2139

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2160)

### DIFF
--- a/chroma_core/models/filesystem.py
+++ b/chroma_core/models/filesystem.py
@@ -109,6 +109,20 @@ class ManagedFilesystem(StatefulObject):
         # (http://stackoverflow.com/questions/4764110/django-template-cant-loop-defaultdict)
         return dict(servers)
 
+    def get_server_groups(self):
+        """Return: Array(Array(ManagedHost)) """
+        targets = self.get_targets()
+
+        groups = {}
+
+        for t in targets:
+            hosts = [x.host for x in t.managedtargetmount_set.all()]
+            name = ",".join(sorted([x.fqdn for x in hosts]))
+            if name not in groups:
+                groups[name] = hosts
+        
+        return groups.values()
+
     def get_pools(self):
         return OstPool.objects.filter(filesystem=self)
 


chroma_core/models/filesystem.py
```diff
@@ -109,6 +109,20 @@ def get_servers(self):
         # (http://stackoverflow.com/questions/4764110/django-template-cant-loop-defaultdict)
         return dict(servers)
 
+    def get_server_groups(self):
```
jgrund, line 112: Where does this get called?

utopiabound: from my task-hotpools patch, I could include it there instead.

utopiabound:
> Why can't we read actual corosync / pacemaker output to get this?

We need EACH ha group and individual corosync clusters don't know each-other.

jgrund:
> We need EACH ha group and individual corosync clusters don't know each-other.

But wouldn't we collect them in aggregate in the manager?

utopiabound:
> But wouldn't we collect them in aggregate in the manager?

We could, but I'm not sure of a structure that would be easier to pull the information out of.  This code should do roughly the same number of database lookups as `get_servers()`

jgrund:
Using `Target.managedtargetmount_set.all()` is not equivalent to where a target can run because it doesn't factor in (at least) location constraints. For example:

```
In [21]: targets[0].managedtargetmount_set.all()                                                                                                                        │:            if name not in groups:
Out[21]: <QuerySet [<ManagedTargetMount: ai400-0a9c-vm00:primary:MGS>, <ManagedTargetMount: ai400-0a9c-vm03:failover:MGS>, <ManagedTargetMount: ai400-0a9c-vm01:failover│:                groups[name] = hosts
:MGS>, <ManagedTargetMount: ai400-0a9c-vm02:failover:MGS>]>    
```

Shows that a target could be mounted on four hosts (based on backing storage). However if we look at negative location constraints:

```
Negative Location Constraints:
 mgs-vg-placement-rule  prevents mgs-vg from running on ai400-0a9c-vm02 (2)
 mgs-vg-placement-rule  prevents mgs-vg from running on ai400-0a9c-vm03 (1)
 mgs-placement-rule     prevents mgs from running on ai400-0a9c-vm02 (2)
 mgs-placement-rule     prevents mgs from running on ai400-0a9c-vm03 (1)
```

We can see that the resource can only run on two of the four places listed.

If we imagine a scenario where all targets in a fs had a location constraint (suppose the mdts / osts had the same constraint), We would overreport the servers that could run the resource.

I wonder if instead, we should report the union of where a resource is currently running (grouped by cluster). This would exclude nodes that should not be present due to current resource constraints.

jgrund (changes requested):
#2204 should be used instead.

jgrund:
#2204  has landed and should be used instead. Closing this.

Issue #1464, opened 2020-01-11
Hotpools


jgrund:
Tracker for all pieces needed for Hotpools

PR #2178, opened 2020-08-21
Add create and destroy resource

utopiabound:
Related #1464

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2178)


iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -192,16 +192,16 @@ fn process_resource(output: &[u8]) -> Result<ResourceAgentInfo, ImlAgentError> {
     Ok(create(&element))
 }
 
-pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAgentError> {
+async fn resource_list() -> Result<Vec<String>, ImlAgentError> {
     let resources = match crm_resource(&["--list-raw"]).await {
-        Ok(res) => res,
+        Ok(res) => res.to_string(),
```
johnsonw, line 197: Can this just be `res`?

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -216,17 +216,208 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
-                process_resource(xml.as_bytes())
-            } else {
-                Err(ImlAgentError::MarkerNotFound)
-            }
-        });
+        .map(str::to_string)
+        .collect())
+}
+
+pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAgentError> {
```
johnsonw, line 223: Is the parameter needed?

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -216,17 +216,208 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
+        if let Some(xml) = output.split("xml:").last() {
+            process_resource(xml.as_bytes())
+        } else {
+            Err(ImlAgentError::MarkerNotFound)
+        }
+    });
     try_join_all(xs).await
 }
 
+fn xml_add_op<'a>(
```
johnsonw, line 235: Is the lifetime parameter necessary?

iml-agent/src/action_plugins/high_availability/mod.rs
johnsonw: The reason I ask is that it's saying this may be an unnecessary clone.

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -216,17 +219,324 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
+    let ids = resource_list().await?;
+
+    if ids.contains(&format!("cl-{}-client", fsname)) {
+        return Ok(());
+    }
+
+    let unit = format!(
+        "{}.mount",
+        mountpoint
+            .trim_start_matches("/")
```
johnsonw, line 317: Should a character be used here? Do you see this clippy error locally?

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -216,17 +219,324 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
+            id: format!("{}-client-after-mgs", fsname),
+            first: "mgs".to_string(),
+            first_action: Some(PacemakerActions::Start),
+            then: agent.id.clone(),
+            then_action: Some(PacemakerActions::Start),
+            kind: Some(PacemakerKindOrScore::Score(PacemakerScore::Value(0))),
+        });
+    }
+    let o = cibxpath("query", "//template[@type=\"lustre-server\"]", &["-e"]).await?;
+    // "/cib/configuration/resources/template[@id='lustre-fs0a9c-mdt']"
```
johnsonw, line 353: Is this comment needed?

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -216,17 +219,324 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
+    for line in o.lines() {
+        if let Some(id) = line.split("'").nth(2) {
```
johnsonw, line 355: Should this be a character?

johnsonw:
Looks like there are a couple of clippy errors and wondering if we need to keep the comment.

iml-agent/src/action_plugins/high_availability/fixtures/create-client-cloned.xml
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" encoding="utf-8"?><clone id="cl-lustre-client"><primitive class="systemd" id="lustre-client" type="mnt-lustre.mount"><operations><op id="lustre-client-start-timeout-900s" name="start" timeout="900s" /><op id="lustre-client-monitor-timeout-60s" interval="60s" name="monitor" /></operations><meta_attributes id="lustre-client-meta_attributes"><nvpair id="lustre-client-meta_attributes-target-role" name="target-role" value="Stopped" /></meta_attributes></primitive></clone>
```
ip1981, line 1: format with `xmllint --format` ?

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -216,17 +219,324 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
+    }
+
+    let op = res.append_new_child("op");
+    op.set_attr("name", name);
+    if let Some(value) = interval {
+        op.set_attr("id", format!("{}-{}-timeout-{}", id, name, value))
+            .set_attr("interval", value);
+    }
+    if let Some(value) = timeout {
+        op.set_attr("id", format!("{}-{}-timeout-{}", id, name, value))
```
ip1981, line 256: This line looks the same as above.

ip1981, line 256: Is it possible both timeout and interval are Some? What would be the id if `op`?
Why isn't `op` mutable? :)

iml-agent/src/action_plugins/high_availability/fixtures/create-client-cloned.xml
utopiabound: no, then I can't do string matching, and elementtree does not implement Eq for Elements

iml-agent/src/action_plugins/high_availability/mod.rs
utopiabound: Possible: yes (Monitor operation can techincally have both an interval and a timeout, but practically, not really).
If both set, it would be "interval", but id doesn't really matter, it just needs to be unique.
Compiler says it doesn't need to be mutable.

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -45,6 +45,19 @@ pub fn create_registry() -> action_plugins::Actions {
...
+            "ha_resource_create",
+            high_availability::create_single_resource,
+        )
+        .add_plugin(
+            "ha_resource_create_cloned",
+            high_availability::create_cloned_resource,
+        )
+        .add_plugin("ha_resource_destroy", high_availability::destroy_resource)
+        .add_plugin(
+            "create_cloned_client",
```
jgrund, line 58:
```suggestion
            "ha_create_cloned_client",
```

So it's clearer to RPC callers.

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -45,6 +45,19 @@ pub fn create_registry() -> action_plugins::Actions {
...
+            high_availability::create_single_resource,
+        )
+        .add_plugin(
+            "ha_resource_create_cloned",
+            high_availability::create_cloned_resource,
+        )
+        .add_plugin("ha_resource_destroy", high_availability::destroy_resource)
+        .add_plugin(
+            "ha_create_cloned_client",
+            high_availability::create_cloned_client,
```
ip1981, line 59: There is obvious inconstancy in naming :)

ip1981 (dismissed):

iml-agent/src/action_plugins/action_plugin.rs
johnsonw: What do you think of, `ha_client_create_clone`?

johnsonw (dismissed):

ip1981 (dismissed):

mkpankov:
Looks like it fails to build with an issue that's fixed on master, you'll probably want to rebase

mkpankov (approved):

mkpankov (approved):

iml-agent/src/action_plugins/lustre/client.rs
```diff
@@ -96,12 +96,16 @@ pub async fn mount_many(xs: Vec<Mount>) -> Result<(), ImlAgentError> {
 pub async fn unmount(unmount: Unmount) -> Result<(), ImlAgentError> {
     delete_fstab_entry(&unmount.mountpoint).await?;
 
-    Command::new("/bin/umount")
-        .arg(unmount.mountpoint)
-        .checked_output()
-        .err_into()
-        .await
-        .map(drop)
+    if is_mounted(&unmount.mountpoint, false).await? {
```
jgrund, line 99: Why?

iml-wire-types/src/lib.rs
```diff
@@ -1590,16 +1590,62 @@ impl PacemakerOperations {
             stop: stop.into(),
         }
     }
+
+    pub fn is_some(&self) -> bool {
```
jgrund, line 1594: This name is confusing as it makes it appear that entire struct is wrapped in an Option.

jgrund (changes requested):
https://github.com/whamcloud/integrated-manager-for-lustre/blob/2cd9ed8465aee05e85b4afebc199da34888469ec/iml-agent/src/high_availability.rs

Is where fns related to high-availability should live. In addition, we should be using a streaming parser to read the XML instead of searching through the returned document.

iml-agent/src/action_plugins/lustre/client.rs
utopiabound: So it's idempotent

jgrund: Also masks the error where you ask something to unmount that's already unmounted.

utopiabound: Should that be an error?

jgrund: I think so

```
[root@mds1 ~]# umount /foo/bar
umount: /foo/bar: mountpoint not found
[root@mds1 ~]# echo $?
32
```

iml-agent/src/action_plugins/action_plugin.rs
jgrund: These should be consistent

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -47,6 +47,23 @@ pub fn create_registry() -> action_plugins::Actions {
...
+            high_availability::create_cloned_resource,
+        )
+        .add_plugin("ha_resource_destroy", high_availability::destroy_resource)
+        .add_plugin(
+            "ha_cloned_client_create",
+            high_availability::create_cloned_client,
+        )
+        .add_plugin(
+            "ha_cloned_client_destroy",
+            high_availability::destroy_cloned_client,
```
jgrund, line 65: These should be consistent

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -47,6 +47,23 @@ pub fn create_registry() -> action_plugins::Actions {
         .add_plugin("unmount_many", lustre::client::unmount_many)
         .add_plugin("ha_resource_start", high_availability::start_resource)
         .add_plugin("ha_resource_stop", high_availability::stop_resource)
+        .add_plugin(
+            "ha_resource_create",
+            high_availability::create_single_resource,
```
jgrund, line 52: These should be consistent

iml-agent/src/action_plugins/action_plugin.rs
```diff
@@ -47,6 +47,23 @@ pub fn create_registry() -> action_plugins::Actions {
...
+        )
+        .add_plugin(
+            "ha_resource_create_cloned",
+            high_availability::create_cloned_resource,
```
jgrund, line 56: These should be consistent

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -335,8 +658,8 @@ pub async fn stop_resource(resource: String) -> Result<(), ImlAgentError> {
 #[cfg(test)]
 mod tests {
     use super::{
-        process_resource, resource_status, PacemakerOperations, ResourceAgentInfo,
-        ResourceAgentType,
+        process_resource, resource_status, resource_xml_string, PacemakerOperations,
+        ResourceAgentInfo, ResourceAgentType,
     };
     use elementtree::Element;
```
jgrund, line 664: Clippy warning

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -65,46 +73,10 @@ fn create(elem: &Element) -> ResourceAgentInfo {
...
-    let o = Command::new("crm_mon")
-        .args(&["--as-xml", "--inactive"])
-        .checked_output()
-        .await?;
-
-    Ok(Element::from_reader(o.stdout.as_slice())?)
-}
-
-fn resource_status(tree: Element) -> HashMap<String, String> {
+fn resource_status(rawxml: &[u8]) -> Result<HashMap<String, String>, ImlAgentError> {
```
jgrund, line 76: This should be added to the `Cluster` struct generated by `read_crm_output` instead of parsing the raw cibxml.

We don't need two different ways to do the same thing, and using a streaming reader is more error tolerant than searching a document + find + unwrapping.

jgrund (changes requested):
I don't think we should use `Elementtree` when we have a streaming parser that does the same thing in a more error-tolerant way.

iml-agent/src/action_plugins/high_availability/mod.rs
utopiabound: `read_crm_output` is insufficient, and limited in scope like the python code, and these action plugins aren't self-limiting to certain resource agents.

jgrund: > read_crm_output is insufficient, and limited in scope like the python code

How? It can parse the entire cibxml.

utopiabound: It does a similar but smaller check than `_get_resource_locations` in the python code only returning certain resource types, and it doesn't do any accounting for cloned or grouped resources.
```rust
fn is_lustre_ra(x: &str) -> bool {
    x == "ocf::ddn:lustre-server" || x == "ocf::lustre:Lustre"
}
...
                if x.get("resource_agent")
                    .filter(|x| is_lustre_ra(x))
                    .is_none()
                {
                    continue;
                }
```
If I change that, than what gets reported by the corosync daemon plugin changes.

jgrund: I think it would be better to expand the `Cluster` struct to account for all resource types + clones and groups and then filter downstream in the places we need just Lustre resources, rather than have two ways of doing mostly the same thing.

iml-agent/src/action_plugins/lustre/client.rs
utopiabound: If something fails, it becomes harder to unwedge.

jgrund:
Looks like this has a conflict on latest master.

iml-agent/src/action_plugins/lustre/client.rs
utopiabound: This also makes it consistent with `mount()` as calling mount twice doesn't result in double mounting the target.

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -176,15 +141,377 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
+
+    // The reason the resource is ALWAYS created in the stopped state,
+    // is that we need to add constraints prior to starting so it will
+    // start on the correct node, or wait for other resources to
+    // start.
+    if started {
+        cibxpath(
+            "delete",
+            &format!(
+                "//resources/primitive[@id=\"{}\"]/meta_attributes",
```
ip1981, line 491: nit:
```suggestion
                r#"//resources/primitive[@id="{}"]/meta_attributes"#,
```

ip1981 (approved):

johnsonw (approved):

iml-agent/src/action_plugins/high_availability/mod.rs
```diff
@@ -176,15 +141,377 @@ pub async fn get_ha_resource_list(_: ()) -> Result<Vec<ResourceAgentInfo>, ImlAg
...
+    nvpair
+        .set_attr("id", format!("{}-{}", &id, name))
+        .set_attr("name", name)
+        .set_attr("value", value);
+}
+
+pub async fn create_single_resource(
+    (agent, constraints): (ResourceAgentInfo, Vec<ResourceConstraint>),
+) -> Result<(), ImlAgentError> {
+    let ids = resource_list().await?;
```
jgrund, line 197: An reason `read_crm_output` could not be used here instead?

PR #2186, opened 2020-08-25
Graphql

jgrund:
Implement Experimental GraphQL support

This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2186)


jgrund:
This patch adds two new endpoints:

`/graphql` - Accepts a POST and data as described in https://graphql.org/

`/graphiql` - Loads an interactive query visualizer in the browser. 

Screen Shot 2020-08-25 at 5 54 24 PM

iml-api/src/graphql.rs
```diff
@@ -0,0 +1,128 @@
...
+        match self {
+            Self::Asc => "asc",
+            Self::Desc => "desc",
+        }
+    }
+}
+
+#[juniper::graphql_object(Context = Context)]
+impl QueryRoot {
+    #[graphql(arguments(
```
johnsonw, line 62: This effectively gives a default it looks like. Seems like a good solution to having to explicitly specify a limit/direction on every call.

jgrund: This is just docs for the query browser. The default is due to the `unwrap_or` calls within this fn.

mkpankov (approved):
> Reviewed 8 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2186)* status: all files reviewed, 2 unresolved discussions (waiting on @jgrund and @johnsonw)

___
*[chroma-manager.conf.template, line 226 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2186#-MFeT_yqAGbkmw90jcdI:-MFeT_yqAGbkmw90jcdJ:b-xeg74d) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/9325dc8b2013389632c961b3f53eaf34418f9013/chroma-manager.conf.template#L226)):*
> ```
>         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
>         proxy_pass {{IML_API_PROXY_PASS}}/graphiql;
>     }
> ```

Why no compression here?
___
*[iml-api/src/graphql.rs, line 62 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2186#-MF_a1XLK-DVhjELXAtQ-r1-62:-MFeUHP5CV0APjmMlgbk:bcfspgu) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/9325dc8b2013389632c961b3f53eaf34418f9013/iml-api/src/graphql.rs#L62)):*

This is just docs for the query browser. The default is due to the `unwrap_or` calls within this fn.
</blockquote></details>

In theory, could we just accept same JSON with limit, offset and direction and not use GraphQL here? Just trying to understand what does GraphQL bring in specifically.

jgrund:
___
*[chroma-manager.conf.template, line 226 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2186#-MFeT_yqAGbkmw90jcdI:-MFeVvnYDBHYxF07pFeV:b-6d3vn) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/9325dc8b2013389632c961b3f53eaf34418f9013/chroma-manager.conf.template#L226)):*

Why no compression here?
</blockquote></details>

I'll add in a follow-on

jgrund:
___
*[iml-api/src/graphql.rs, line 62 at r1](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2186#-MF_a1XLK-DVhjELXAtQ-r1-62:-MFeVyoP6Eh1D7pfUkbR:b-rt6uz) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/9325dc8b2013389632c961b3f53eaf34418f9013/iml-api/src/graphql.rs#L62)):*

In theory, could we just accept same JSON with limit, offset and direction and not use GraphQL here? Just trying to understand what does GraphQL bring in specifically.
</blockquote></details>

Yes, we could do the same with query string params.

currently, the biggest draws are:

1. API Documentation generated from code (which is very important for external consumers)
1. A query explorer
1. A consistently enforced way to structure endpoints
1. The ability to request just the data that is needed from clients

Of all of those, the ability to generate extensive documentation (which we will need to stay on top of) is the most important imo.

PR #2196, opened 2020-08-28
Account for multiple fs with the same target name

johnsonw:
The MGS target has the same name for multiple filesystems. As an
example, a filesystem created with an MGT on MDS1 and a filesystem
created with an MGT on OSS1 will both have a target name of `MGS`. This
needs to be accounted for when building the target table.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2196)

### DIFF
--- a/iml-influx/Cargo.toml
+++ b/iml-influx/Cargo.toml
@@ -8,3 +8,4 @@ edition = "2018"
 [dependencies]
 iml-wire-types = { path = "../iml-wire-types", version = "0.3" }
 serde = { version = "1", features = ['derive'] }
+serde_json = { version = "1" }
--- a/iml-influx/Cargo.toml
+++ b/iml-influx/Cargo.toml
@@ -9,3 +9,6 @@ edition = "2018"
 iml-wire-types = { path = "../iml-wire-types", version = "0.3" }
 serde = { version = "1", features = ['derive'] }
 serde_json = { version = "1" }
+
+[dev-dependencies]
+influx_db_client = {version = "0.4", default-features = false, features = ["rustls-tls"]}
--- a/iml-services/iml-device/src/lib.rs
+++ b/iml-services/iml-device/src/lib.rs
@@ -18,13 +18,34 @@ use iml_postgres::sqlx::{self, PgPool};
 use iml_tracing::tracing;
 use iml_wire_types::Fqdn;
 use influx_db_client::{keys::Node, Client, Precision};
+use serde_json::{Map, Value};
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap},
     sync::Arc,
 };
 
 pub type Cache = Arc<Mutex<HashMap<Fqdn, Device>>>;
-pub type TargetFsRecord = HashMap<String, Vec<String>>;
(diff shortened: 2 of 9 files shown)


jgrund:
Looks like this is introducing a number of clippy warnings.

iml-services/iml-device/src/lib.rs
```diff
@@ -469,35 +481,69 @@ pub fn build_updates(x: Changes<'_, Target>) -> Vec<Target> {
...
+                        x.values
                             .into_iter()
-                            .zip(x.values.into_iter().flatten())
-                            .collect::<HashMap<String, serde_json::Value>>()
+                            .map(|v| {
+                                columns
+                                    .iter()
+                                    .cloned()
+                                    .zip(v.into_iter())
+                                    .collect::<HashMap<String, serde_json::Value>>()
```
jgrund, line 498: I get the feeling that we should create a `From` impl that takes in a tuple of (columns, values) and produces a https://docs.serde.rs/serde_json/value/enum.Value.html#variant.Object. 

Once we have an `Object` we can then deserialize it directly to a struct instead of having to do the check / panics we are doing here.

johnsonw: I took a stab at this in the following patch. If this isn't quite what you're looking for can we discuss further?

jgrund: I had something slightly different in mind. In the interest of time, I'll push what I was thinking and we can discuss if you'd like.

jgrund: Pushed

iml-influx/Cargo.toml
```diff
@@ -8,3 +8,7 @@ edition = "2018"
 [dependencies]
 iml-wire-types = { path = "../iml-wire-types", version = "0.3" }
 serde = { version = "1", features = ['derive'] }
+serde_json = { version = "1" }
```
jgrund, line 11: Nit:

```suggestion
serde_json = "1" 
```

iml-services/iml-device/src/lib.rs
```diff
@@ -24,7 +25,27 @@ use std::{
 };
 
 pub type Cache = Arc<Mutex<HashMap<Fqdn, Device>>>;
-pub type TargetFsRecord = HashMap<String, Vec<String>>;
+pub type TargetFsRecord = HashMap<String, Vec<(Fqdn, String)>>;
+
+#[derive(serde::Serialize, serde::Deserialize)]
```
jgrund, line 30: nit, probably don't need to derive `Serialize`.

iml-services/iml-device/src/lib.rs
```diff
@@ -24,7 +25,27 @@ use std::{
...
+struct FsRecord {
+    host: String,
+    target: String,
+    fs: Option<String>,
+    mgs_fs: Option<String>,
+}
+
+impl FsRecord {
+    fn filesystems(&self) -> String {
```
jgrund, line 39: nit, this could probably just return a `&str`, which would elimiate the clone.

ip1981 (approved):

PR #2206, opened 2020-09-01
Add test_setup fn to iml_postgres

jgrund:
Add a test_setup fn to iml_postgres that will run hand out a single
pool connection and run it within a transaction. When the connection is
closed the transaction is rolled back so it is safe to call in parllel
and will not pollute db testing.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2206)

### DIFF
--- a/iml-postgres/src/lib.rs
+++ b/iml-postgres/src/lib.rs
@@ -91,3 +91,21 @@ where
 
     client.query_raw(&s, params).await
 }
+
+#[cfg(feature = "test")]
+use dotenv::dotenv;
+
+/// Setup for a test run. This fn hands out a pool
+/// With a single connection and starts a transaction for it.
+/// The transaction is rolled back when the connection closes
+/// So nothing is written to the database.
+#[cfg(feature = "test")]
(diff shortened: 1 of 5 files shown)


iml-postgres/src/lib.rs
```diff
@@ -91,3 +91,21 @@ where
 
     client.query_raw(&s, params).await
 }
+
+#[cfg(feature = "test")]
+use dotenv::dotenv;
+
+/// Setup for a test run. This fn hands out a pool
+/// With a single connection and starts a transaction for it.
```
johnsonw, line 99:
```suggestion
/// with a single connection and starts a transaction for it.
```

iml-postgres/src/lib.rs
```diff
@@ -91,3 +91,21 @@ where
...
+/// The transaction is rolled back when the connection closes
+/// So nothing is written to the database.
```
johnsonw, line 101:
```suggestion
/// so nothing is written to the database.
```

ip1981 (approved):

PR #2210, opened 2020-09-02
Implement Snapshot listing GraphQL API

mkpankov:
Example:

```
➜  ~ echo '{ "query": "{ snapshots(args:{fsname:\"zfsmo\", name:\"test1\", detail:true}) { filesystemName snapshotName details { snapshotFsname } } }"}' | http --json --verify=false POST https://localhost:8443/graphql/ authorization:'ApiKey api:09e8a32507c26f8ea44566b2cf9a81c15a9b5097'

HTTP/1.1 200 OK
Connection: keep-alive
Content-Length: 116
Content-Type: text/plain; charset=utf-8
Date: Thu, 03 Sep 2020 09:21:01 GMT
Server: nginx/1.16.1

{
    "data": {
        "snapshots": [
            {
                "details": [
                    {
                        "snapshotFsname": "0b93f8b7"
                    }
                ],
                "filesystemName": "zfsmo",
                "snapshotName": "test1"
            }
        ]
    }
}


```

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2210)


iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+        offset: Option<i32>,
+        dir: Option<SortDir>,
+        args: List,
+    ) -> juniper::FieldResult<Vec<Snapshot>> {
+        let dir = dir.unwrap_or_default();
+
+        if args.detail {
+            let xs = sqlx::query!(
+                r#"
+                    SELECT filesystem_name, snapshot_name, create_time, modify_time, snapshot_fsname, mounted, comment from snapshot s
```
ip1981, line 166: nit:
```suggestion
                    SELECT filesystem_name, snapshot_name, create_time, modify_time, snapshot_fsname, mounted, comment FROM snapshot s
```

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+                .into_iter()
+                .map(|x| Snapshot {
+                    snapshot_name: x.snapshot_name,
+                    filesystem_name: x.filesystem_name,
+                    details: vec![Detail {
+                        comment: x.comment,
+                        create_time: x.create_time,
+                        modify_time: x.modify_time,
+                        snapshot_fsname: x.snapshot_fsname,
+                        // FIXME
```
ip1981, line 191: Fix what?

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+    async fn snapshots(
+        context: &Context,
+        limit: Option<i32>,
+        offset: Option<i32>,
+        dir: Option<SortDir>,
+        args: List,
+    ) -> juniper::FieldResult<Vec<Snapshot>> {
+        let dir = dir.unwrap_or_default();
+
+        if args.detail {
```
ip1981, line 163: I think there is no need to support it yet (because there is not support in DB yet, and actual listing is extremely slow). And in any case, `.details[]` exists: by default it includes only one record with `snapshot_role: None` (I believe it is MDT). In a detailed listing it would include components of the snapshot with `snapshot_role` like `OST0001`.

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+                    snapshot_name: x.snapshot_name,
+                    filesystem_name: x.filesystem_name,
+                    details: vec![Detail {
+                        comment: x.comment,
+                        create_time: x.create_time,
+                        modify_time: x.modify_time,
+                        snapshot_fsname: x.snapshot_fsname,
+                        // FIXME
+                        snapshot_role: None,
+                        status: x.mounted.map(|b| {
```
ip1981, line 193: I am ok with getting rid of `Status` and using `Option<bool>`. Don't know what's better.

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+
+            Ok(snapshots)
+        } else {
+            let xs = sqlx::query!(
+                r#"
+                    SELECT filesystem_name, snapshot_name from snapshot s
+                    WHERE filesystem_name = $4 AND $5::text IS NULL OR snapshot_name = $5
+                    ORDER BY
+                        CASE WHEN $3 = 'asc' THEN s.snapshot_name END ASC,
+                        CASE WHEN $3 = 'desc' THEN s.snapshot_name END DESC
```
ip1981, line 212: Maybe ordering by creation time would make more sense?

iml-api/src/graphql.rs
mkpankov: How is there no support in DB if the `details` field of `Snapshot` is saved to DB? So we either ask for them or not.

iml-api/src/graphql.rs
mkpankov: I think we either need to make `Unknown` a variant in the enum and get rid of the `Option`, or get rid of the enum. Depends if there are other statuses besides `Mounted`  and `Unmounted`.

iml-api/src/graphql.rs
mkpankov: See my review on your PR.

iml-api/src/graphql.rs
johnsonw: If you had the `mount_path` as an option you could check if it is a `some` and if it is that would evaluate to a state of mounted.

johnsonw: The other option is that you could implement the `From` trait and then do:
```
status: x.mounted.into()
```

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+                    }],
+                })
+                .collect();
+
+            Ok(snapshots)
+        } else {
+            let xs = sqlx::query!(
+                r#"
+                    SELECT filesystem_name, snapshot_name FROM snapshot s
+                    WHERE filesystem_name = $4 AND $5::text IS NULL OR snapshot_name = $5
```
johnsonw, line 209: Probably need to add parenthesis around conjunctive operators:
```
WHERE filesystem_name = $4 AND ($5::text IS NULL OR snapshot_name = $5)
```

iml-api/src/graphql.rs
johnsonw: I think it would be more beneficial to order by time, personally. If the names are ordered then it's more of a sort by alphabetical name but there is no relative meaning behind the sort. If the snapshots are ordered by time, it could help determine what snapshot to restore. For example, if I had 30 snapshots and I want to quickly go back to the first snapshot when I setu p my system, I would list out the snapshots and scroll to the first one and immediately know that this is the snapshot I want to use.

jgrund: I think the pertinent usecase will be to delete the oldest snapshots so I agree that sorting by time is useful.

mkpankov: Done

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+        dir: Option<SortDir>,
+        args: List,
+    ) -> juniper::FieldResult<Vec<Snapshot>> {
+        let dir = dir.unwrap_or_default();
+
+        if args.detail {
+            let xs = sqlx::query!(
+                r#"
+                    SELECT filesystem_name, snapshot_name, create_time, modify_time, snapshot_fsname, mounted, comment FROM snapshot s
+                    WHERE filesystem_name = $4 AND $5::text IS NULL OR snapshot_name = $5
```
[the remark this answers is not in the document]
mkpankov: Done

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,94 @@ impl QueryRoot {
...
+                args.name,
+            )
+            .fetch_all(&context.client)
+            .await?;
+
+            let snapshots: Vec<_> = xs
+                .into_iter()
+                .map(|x| Snapshot {
+                    snapshot_name: x.snapshot_name,
+                    filesystem_name: x.filesystem_name,
```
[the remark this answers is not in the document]
mkpankov: It's not a tuple but a struct, so this doesn't work

```
    Checking iml-api v0.3.0 (/home/mkpankov/work/integrated-manager-for-lustre/iml-api)
error[E0308]: mismatched types
   --> iml-api/src/graphql.rs:224:23
    |
224 |                 .map(|(snapshot_name, filesystem_name)| Snapshot {
    |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                       |
    |                       expected struct `<graphql::QueryRoot as juniper::GraphQLValueAsync>::resolve_field_async::{{closure}}#2::{{closure}}#0::Record`, found tuple
    |                       expected due to this
    |
    = note: expected struct `<graphql::QueryRoot as juniper::GraphQLValueAsync>::resolve_field_async::{{closure}}#2::{{closure}}#0::Record`
                found tuple `(_, _)`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `iml-api`.
```

ip1981:
BTW, I think we need an index for `create_time`.

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,93 @@ impl QueryRoot {
...
+    async fn snapshots(
+        context: &Context,
+        limit: Option<i32>,
+        offset: Option<i32>,
+        dir: Option<SortDir>,
+        args: List,
+    ) -> juniper::FieldResult<Vec<Snapshot>> {
+        let dir = dir.unwrap_or_default();
+
+        if args.detail {
```
ip1981, line 163: We are going to have it always "false". So you can just remove this condition, but keep the branch (remove "else").

iml-api/src/graphql.rs
jgrund: I think `detail` is being overloaded. 

A non-detail call has this result:

```
lctl snapshot_list -F zfsmo --name zfsmo-snapshot1

filesystem_name: zfsmo
snapshot_name: zfsmo-snapshot1
create_time: Wed Sep  2 15:29:16 2020
modify_time: Wed Sep  2 15:29:16 2020
snapshot_fsname: 462af955
status: not mount
```

Where a detail call looks like:

```
[root@mds1 ~]# lctl snapshot_list -F zfsmo --name zfsmo-snapshot1 --detail

filesystem_name: zfsmo
snapshot_name: zfsmo-snapshot1

snapshot_role: MGS
snapshot_fsname: 462af955
modify_time: Wed Sep  2 15:29:16 2020
create_time: Wed Sep  2 15:29:16 2020
status: not mount

snapshot_role: MDT0000
create_time: Wed Sep  2 15:29:16 2020
modify_time: Wed Sep  2 15:29:16 2020
snapshot_fsname: 462af955
status: not mount

...
```

Both calls have the same fields, the difference is the detail call is listing each target. 

Also, details will be empty if the MGT / MDT are not co-located on the same node.

iml-api/src/graphql.rs
jgrund: Why is there an option currently in 

https://github.com/whamcloud/integrated-manager-for-lustre/blob/ce8f0c996d49a59a7105c826410fc86fc7cfcf8a/iml-wire-types/src/snapshot.rs#L50

Is it possible to have a `None`?

iml-api/src/graphql.rs
```diff
@@ -143,6 +144,93 @@ impl QueryRoot {
...
+        dir(description = "Sort direction, defaults to asc"),
+        args(description = "Snapshot listing arguments")
+    ))]
+    /// Fetch the list of snapshots
+    async fn snapshots(
+        context: &Context,
+        limit: Option<i32>,
+        offset: Option<i32>,
+        dir: Option<SortDir>,
+        args: List,
```
jgrund, line 159: I think we need to expand the list args into individual params in the interest of user-friendliness.

The downside of reusing the args / types from wire-types is we assume that the structs are one-size-fits-all, which is not always going to be the case.

iml-api/src/graphql.rs
jgrund: In addition, I wonder if this struct should not be nested as the result of this API call.

iml-api/src/graphql.rs
ip1981: In case of snapshots, it seems fine. The structures are small and most fields are required. In any case, de-serializing will handle optional fields.

jgrund: How does this look in the `graphiql` client? I.E. how does the user input this?

jgrund (approved):
I think we're going to change the underlying `Snapshot` struct to remove detail, but let's do that in a follow-on.

iml-api/src/graphql.rs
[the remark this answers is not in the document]
johnsonw: Odd, I pulled your patch down and it seemed to compile fine.

johnsonw (approved):

PR #2215, opened 2020-09-03
Map multipath bindings to devices on vagrant

### LINKED ISSUE #2197: map multipath bindings to devices on vagrant
Iscsi creates many devices to be used on the MDS and OSS nodes. These devices are typically 5GB for MDT's and OST's. MGT's however, do not require nearly this much space. The concern is that the provisioners format and mount devices and depend on these devices being mapped to the correct mpath. For example, consider that we want to format and mount the first 5 OST devices, each being 5GB. We would expect:

/dev/mapper/mpatha => /mnt/ost1,
/dev/mapper/mpathb => /mnt/ost2,
/dev/mapper/mpathc => /mnt/ost3,
/dev/mapper/mpathd => /mnt/ost4,
/dev/mapper/mpathe => /mnt/ost5

But, what if we added a new mgt to the OSS node pair and we expected it to be mapped to the very last mpath entry (/dev/mapper/mpathu). If there is nothing to guarantee that the devices are mapped as above, then this new MGT, which is only 512MB could end up being mapped to /dev/mapper/mpathn, which should have been assigned to OST14. This creates issues where we think we are formatting and mounting ostx but we are really working on osty. To prevent this issue, we can simply ensure the bindings file maps each of the WWID's accordingly, stop the multipath service, flush it, and then start the multipath service. This will ensure that the devices are mapped as above and the new MGT, in this case, will be assigned to /dev/mapper/mpathu instead of /dev/mapper/mpathn.

johnsonw:
- Fixes #2197.

- Map the devices using a natural sort order.
- Update the zfs filesystem creation in the Vagrantfile to create a
second fs that lives on the OSS.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2215)

### DIFF
--- a/vagrant/Vagrantfile
+++ b/vagrant/Vagrantfile
@@ -521,6 +521,12 @@ Vagrant.configure('2') do |config|
 
       provision_mpath oss
 
+      oss.vm.provision "file",
+                      source: "./scripts/update_oss_mpath.py",
+                      destination: "/tmp/update_oss_mpath.py"
+
+      configure_mpath oss
+
       provision_fence_agents oss
 
       cleanup_storage_server oss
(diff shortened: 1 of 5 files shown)


vagrant/scripts/update_oss_mpath.py
```diff
@@ -0,0 +1,34 @@
...
+import fileinput
+import re
+
+
+def natural_sort_key(s, _nsre=re.compile("([0-9]+)")):
+    return [int(text) if text.isdigit() else text.lower() for text in _nsre.split(s)]
+
+
+devices = {}
+for line in fileinput.input():
```
nlinker, line 15: Could you show an example, how `fileinput` looks like?

vagrant/Vagrantfile
```diff
@@ -600,26 +622,42 @@ Vagrant.configure('2') do |config|
...
-                              mount -t lustre ost7/ost7 /lustre/zfsmo/ost7
-                              mount -t lustre ost8/ost8 /lustre/zfsmo/ost8
-                              mount -t lustre ost9/ost9 /lustre/zfsmo/ost9
+                         SHELL
+
+          oss.vm.provision 'mount-zfs-fs2',
+                         type: 'shell',
+                         run: 'never',
+                         inline: <<-SHELL
+                              mkdir -p /lustre/zfsmo2/{mgt2,mdt2-0}
```
mkpankov, line 655: Can we setup `mgt2` on `mds2` as main server and `mds1` as fallback (mirrored from how `mgt` is setup)? Would be more symmetrical.

ip1981 (approved):

PR #2221, opened 2020-09-06
Add deadline support for daemon plugins

### LINKED ISSUE #2219: customize daemon plugin deadline
Daemon plugins are currently checked once a second. 

If a plugin is active and it's `instant` is in the past, it's `poll` method is called. 

Once the `poll` method has been called, the plugin has 1s to complete it's work before it is checked again.

If the plugin does not complete the poll before it's checked, the current poll task is dropped and it is rescheduled for 10s in the future.

This has worked ok so far, but for slower plugins we need to either run their monitoring in a background task or parallelize their monitoring to run in under 1s. 

Instead, we should allow for plugin authors to define the deadline per-plugin. We can default to 10s and allow it to be overridden in the plugin definition.

jgrund:
Fixes #2219

This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2221)

### DIFF
--- a/iml-agent/README.md
+++ b/iml-agent/README.md
@@ -40,7 +40,7 @@ Daemon plugins live in the `iml-agent/src/daemon_plugins` directory.
 
 Unlike action plugins, daemon plugins are stateful. They send an initial data payload to the manager on startup and then optionally send an update payload every 10 seconds. While action plugins are useful for management actions, daemon-plugins are useful for monitoring the distributed system.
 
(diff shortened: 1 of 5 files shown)


ip1981 (approved):

mkpankov (dismissed):
> Reviewed 5 of 5 files at r1.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2221)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jgrund)

johnsonw (approved):

ip1981 (approved):

PR #2235, opened 2020-09-10
Add barrier support to snapshot create

jgrund:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2235)


ip1981:
AFAIK barrier is used by default. Do we need to turn it off?

jgrund:
> AFAIK barrier is used by default. Do we need to turn it off?

Yes, it should default to off and needs to be configurable.

ip1981 (approved):

johnsonw (approved):

PR #2236, opened 2020-09-11
Add iml-graphql-queries crate

jgrund:
Add a new crate, `iml-graphql-queries` that serves as a simple client
query builder.

It has no dependency on transport and can be used in both wasm and
native contexts.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2236)


johnsonw (approved):
Do you think it would be beneficial to have a `run` command that would make the HTTP call to the graphql endpoint with the specified query?

jgrund:
> Do you think it would be beneficial to have a run command that would make the HTTP call to the graphql endpoint with the specified query?

No, this is purposely client agnostic so it can be used with wasm and native code, i.e. seed on the front-end and reqwest on the backend.

johnsonw (dismissed):

iml-graphql-queries/src/snapshot.rs
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+pub mod create {
+    use crate::Query;
+
+    pub const QUERY: &str = r#"
+            mutation CreateSnapshot($fsname: String!, $name: String!, $comment: String, $use_barrier: Boolean) {
```
ip1981, line 9: I think it's `createSnapshot` (camelCase) :)

iml-graphql-queries/src/snapshot.rs
```diff
@@ -0,0 +1,48 @@
...
+              createSnapshot(fsname: $fsname, name: $name, comment: $comment, use_barrier: $use_barrier) {
+                cancelled
+                complete
+                createdAt
```
ip1981, line 13:
```suggestion
                created_at: createdAt
```
For smooth transition.

iml-graphql-queries/src/snapshot.rs
jgrund: not for the wrapper, ref: https://graphql.org/learn/queries/#mutations

iml-graphql-queries/src/snapshot.rs
jgrund: should we alias all of the fields then? (probably)

jgrund: nm.

jgrund: nice fix.

ip1981:
Might need a helper to decode the responses. Like

```rust
serde_json::from_value(resp.get("data").unwrap().get("createSnapshot").unwrap().clone())
```

ip1981 (approved):

jgrund:
> Might need a helper to decode the responses.

Yeah, I'll try to do that in a follow-on

jgrund:
Went ahead and implemented response support as well:

Here's how I'm creating a request in seed:

```rust
snapshot::create::build(&model.fs_name, &model.name, model.comment.as_ref(), Some(model.barrier));

let req = fetch::Request::graphql_query(&query);

orders.perform_cmd(req.fetch_json_data(|x| Msg::SnapshotCreateResp(x)));
```

and here's how the response looks like in `msg`:

```rust
SnapshotCreateResp(fetch::ResponseDataResult<Response<snapshot::create::Resp>>)
```

Will probably write a seed helper to collapse the nesting of that response, but that can be in another patch.

johnsonw (approved):

ip1981 (approved):

PR #2247, opened 2020-09-15
Add snapshot configuration API

jgrund:
This patch adds support for creating and deleting snapshot intervals
and snapshot retention rules via the API.

It also adds support for these tables to warp-drive.

Screen Shot 2020-09-16 at 10 08 02 AM

Screen Shot 2020-09-16 at 10 17 30 AM

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2247)

### DIFF
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -4,21 +4,31 @@
 
 use crate::{command::get_command, error::ImlApiError};
 use futures::{TryFutureExt, TryStreamExt};
-use iml_postgres::{sqlx, PgPool};
+use iml_postgres::{sqlx, sqlx::postgres::types::PgInterval, PgPool};
 use iml_rabbit::Pool;
-use iml_wire_types::{snapshot::Snapshot, Command, EndpointName, Job};
+use iml_wire_types::{
+    graphql_duration::GraphQLDuration,
+    snapshot::{DeleteUnit, DeleteWhen, Snapshot, SnapshotInterval, SnapshotRetention},
+    Command, EndpointName, Job,
+};
 use itertools::Itertools;
 use juniper::{
     http::{graphiql::graphiql_source, GraphQLRequest},
     EmptySubscription, FieldError, GraphQLEnum, RootNode, Value,
 };
-use std::ops::Deref;
 use std::{
     collections::{HashMap, HashSet},
-    convert::Infallible,
+    convert::{Infallible, TryFrom as _},
+    ops::Deref,
     sync::Arc,
 };
 use warp::Filter;
+
+#[derive(juniper::GraphQLInputObject)]
+struct InputDeleteWhen {
+    value: i32,
+    unit: DeleteUnit,
+}
 
 #[derive(juniper::GraphQLObject)]
 /// A Corosync Node found in `crm_mon`
@@ -104,9 +114,6 @@ struct BannedResource {
     master_only: bool,
 }
 
-pub(crate) struct QueryRoot;
-pub(crate) struct MutationRoot;
-
 #[derive(GraphQLEnum)]
 enum SortDir {
     Asc,
@@ -129,6 +136,8 @@ impl Deref for SortDir {
         }
     }
 }
+
+pub(crate) struct QueryRoot;
 
 #[juniper::graphql_object(Context = Context)]
 impl QueryRoot {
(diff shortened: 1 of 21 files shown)


iml-api/src/graphql.rs
```diff
@@ -3,26 +3,60 @@
...
+impl<S> GraphQLScalar for GraphQLDuration
+where
+    S: juniper::ScalarValue,
+{
+    fn resolve(&self) -> juniper::Value {
+        juniper::Value::scalar(humantime::format_duration(self.0).to_string())
+    }
+
+    fn from_input_value(value: &juniper::InputValue) -> Option<GraphQLDuration> {
+        let x = value
```
johnsonw, line 40: Just wondering if we could do something like:
```
value
    .as_string_value()?
.parse::<humantime::Duration>()
.ok()
.map(|x| GraphQLDuration(x))
```
?

iml-api/src/graphql.rs
```diff
@@ -392,6 +384,56 @@ impl QueryRoot {
...
+        Ok(xs)
+    }
+    /// List all snapshot retention policies
+    async fn list_snapshot_retention_policies(
+        context: &Context,
+    ) -> juniper::FieldResult<Vec<SnapshotRetention>> {
+        let xs: Vec<SnapshotRetention> = sqlx::query!(
+            r#"
+                SELECT
+                    id,
```
johnsonw, line 412: I didn't think about this earlier but I guess this could be:
```
SELECT * FROM snapshot_retention
```

iml-api/src/graphql.rs
jgrund: yeah, it could be all combinators.

iml-api/src/graphql.rs
jgrund: no, because delete_unit is being cast.

iml-api/src/graphql.rs
```diff
@@ -1039,3 +899,37 @@ fn validate_snapshot_name(x: &str) -> Result<(), FieldError> {
...
+
+async fn active_mgs_host_fqdn(
+    fsname: &str,
+    pool: &PgPool,
+) -> Result<Option<String>, iml_postgres::sqlx::Error> {
+    let fsnames = &[fsname.into()][..];
+    let maybe_active_mgs_host_id = sqlx::query!(
+        r#"
+            SELECT active_host_id from target WHERE filesystems @> $1 and name='MGS'
+        "#,
```
johnsonw, line 911: Any reason we can't LEFT JOIN on chroma_core_managedhost? If we could do that it would be a single db call wouldn't it?

jgrund: probably an inner join because we'd want to return None if either of the sides were empty.

That said, this fn was not created in this patch, just moved from another spot in the file. I'm not too keen on changing the query here.

johnsonw: sounds good

mkpankov (dismissed):
> Reviewed 15 of 15 files at r5.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2247)* status: all files reviewed, 3 unresolved discussions (waiting on @jgrund and @johnsonw)

migrations/20200831174542_snapshot.sql
```diff
@@ -31,11 +31,64 @@ DROP TRIGGER IF EXISTS snapshot_notify_delete ON snapshot;
...
+INSERT ON snapshot FOR EACH ROW EXECUTE PROCEDURE table_update_notify();
 
 CREATE TRIGGER snapshot_notify_delete
-AFTER DELETE ON snapshot FOR EACH ROW EXECUTE PROCEDURE table_snapshot_update_notify();
\ No newline at end of file
+AFTER DELETE ON snapshot FOR EACH ROW EXECUTE PROCEDURE table_update_notify();
+
+CREATE TABLE IF NOT EXISTS snapshot_interval (
+  id serial PRIMARY KEY,
+  filesystem_name TEXT NOT NULL,
```
ip1981, line 45: Without a constraint it will be possible to create multiple intervals. no?

iml-wire-types/src/graphql_duration.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use humantime;
+#[cfg(feature = "postgres-interop")]
+use sqlx::postgres::types::PgInterval;
+use std::{convert::TryFrom, convert::TryInto, fmt, time::Duration};
```
nlinker, line 8: Clippy complains

iml-wire-types/src/graphql_duration.rs
```diff
@@ -0,0 +1,62 @@
...
+        x.parse::<humantime::Duration>()
+            .map(|x| GraphQLDuration(x.into()))
+    }
+}
+
+impl From<GraphQLDuration> for String {
+    fn from(x: GraphQLDuration) -> Self {
+        x.to_string()
+    }
+}
```
nlinker, line 56: Where is this needed, looks like unnecessary magic is added, no?

migrations/20200831174542_snapshot.sql
jgrund: Yes, I still need to address this.

iml-wire-types/src/graphql_duration.rs
jgrund: Needed for the `serde = into` attribute on line 11.

migrations/20200831174542_snapshot.sql
jgrund: Fixed.

mkpankov (approved):
> Reviewed 8 of 9 files at r6, 5 of 5 files at r7.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2247)* status: all files reviewed, 6 unresolved discussions (waiting on @ip1981, @jgrund, @johnsonw, and @nlinker)

ip1981 (approved):

iml-api/src/graphql.rs
```diff
@@ -591,39 +635,100 @@ impl MutationRoot {
...
+            "#,
+            fsname,
+            use_barrier.unwrap_or_default(),
+            PgInterval::try_from(interval.0)?,
+        )
+        .execute(&context.pg_pool)
+        .await?;
 
-    tracing::trace!("Maybe active MGS host id: {:?}", maybe_active_mgs_host_id);
+        Ok(true)
```
nlinker, line 670: Usually we are returning the `id` of the row inserted, so that then we can pass it to `remove_snapshot_interval`. Is the idiom not needed here?

iml-api/src/graphql.rs
```diff
@@ -591,39 +635,100 @@ impl MutationRoot {
...
+    /// Creates a new snapshot retention policy for the given `fsname`.
+    /// The filesystem will be checked periodically, and if the policy is met,
+    /// The oldest snapshot will be deleted until the policy no longer holds.
+    async fn create_snapshot_retention(
+        context: &Context,
+        fsname: String,
+        delete_num: i32,
+        delete_unit: DeleteUnit,
+        keep_num: i32,
+    ) -> juniper::FieldResult<bool> {
```
nlinker, line 703: The same question as above for `snapshot_interval` table.

nlinker (approved):
Some questions though.

iml-api/src/graphql.rs
jgrund: Typical graphql idiom is to return true if the mutation didn't have errors. It will be fairly atypical to remove the interval right after adding it. I suppose there could be some value to returning the entire created object, but I'd rather wait to see if that ends up being needed than implementing it now.

PR #2265, opened 2020-09-22
Implement fetching commands by ids

nlinker:
For the ids supplied return the commands, the order is guaranteed to be preserved.
Screenshot 2020-09-22 at 23 45 25
Screenshot 2020-09-22 at 23 47 36

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2265)

### DIFF
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -333,7 +333,7 @@ impl QueryRoot {
         limit(description = "paging limit, defaults to 20"),
         offset(description = "Offset into items, defaults to 0"),
         dir(description = "Sort direction, defaults to ASC"),
-        is_active(description = "Command status, active means not completed, can be omitted"),
+        is_active(description = "Command status, active means not completed, default is true"),
         msg(description = "Substring of the command's message, null or empty matches all"),
     ))]
     async fn commands(
@@ -345,7 +345,7 @@ impl QueryRoot {
         msg: Option<String>,
     ) -> juniper::FieldResult<Vec<Command>> {
         let dir = dir.unwrap_or_default();
-        let is_completed = is_active.map(|active| !active);
+        let is_completed = !is_active.unwrap_or(true);
         let commands: Vec<Command> = sqlx::query_as!(
             CommandTmpRecord,
             r#"
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -459,21 +459,15 @@ impl QueryRoot {
                 .collect::<Vec<_>>()
         })
         .await?;
-        // convert the commands into the hashmap id -> command
         let mut hm = unordered_cmds
             .into_iter()
             .map(|c| (c.id, c))
             .collect::<HashMap<i32, Command>>();
         let commands = ids
             .iter()
-            .map(|id| {
-                if let Some(c) = hm.remove(id) {
-                    Some(c)
-                } else {
-                    None
-                }
-            })
-            .collect();
+            .map(|id| hm.remove(id))
+            .collect::<Vec<Option<Command>>>();
+
         Ok(commands)
     }
 
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -332,9 +332,8 @@ impl QueryRoot {
         limit(description = "paging limit, defaults to 20"),
         offset(description = "Offset into items, defaults to 0"),
         dir(description = "Sort direction, defaults to ASC"),
-        is_active(description = "Command status, active means not completed, default is false"),
+        is_active(description = "Command status, active means not completed, can be omitted"),
         msg(description = "Substring of the command's message, null or empty matches all"),
-        ids(description = "The list of command ids to fetch, default is NULL means no restriction"),
     ))]
     async fn commands(
         context: &Context,
@@ -343,11 +342,9 @@ impl QueryRoot {
         dir: Option<SortDir>,
         is_active: Option<bool>,
         msg: Option<String>,
-        ids: Option<Vec<i32>>,
     ) -> juniper::FieldResult<Vec<Command>> {
         let dir = dir.unwrap_or_default();
-        let is_completed = !is_active.unwrap_or(false);
-        let ids: Option<&[i32]> = ids.as_deref();
+        let is_completed = is_active.map(|active| !active);
         let commands: Vec<Command> = sqlx::query!(
             r#"
                 SELECT
@@ -360,9 +357,8 @@ impl QueryRoot {
                     message
                 FROM chroma_core_command c
                 JOIN chroma_core_command_jobs cj ON c.id = cj.command_id
-                WHERE complete = $4
+                WHERE ($4::BOOL IS NULL OR complete = $4)
                   AND ($5::TEXT IS NULL OR c.message ILIKE '%' || $5 || '%')
-                  AND ($6::INT[] IS NULL OR c.id = ANY ($6::INT[]))
                 GROUP BY c.id
                 ORDER BY
                     CASE WHEN $3 = 'asc' THEN c.id END ASC,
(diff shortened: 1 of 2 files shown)


jgrund (changes requested):
Needs to be on latest master.

iml-api/src/graphql.rs
```diff
@@ -397,6 +398,85 @@ impl QueryRoot {
...
+            .map(|c| (c.id, c))
+            .collect::<HashMap<i32, Command>>();
+        let commands = ids
+            .iter()
+            .map(|id| {
+                if let Some(c) = hm.remove(id) {
+                    Some(c)
+                } else {
+                    None
+                }
```
nlinker, line 474: eh

iml-api/src/graphql.rs
```diff
@@ -344,7 +344,7 @@ impl QueryRoot {
         msg: Option<String>,
     ) -> juniper::FieldResult<Vec<Command>> {
         let dir = dir.unwrap_or_default();
-        let is_completed = !is_active.unwrap_or(false);
+        let is_completed = is_active.map(|active| !active);
```
jgrund, line 347:
```suggestion
        let is_completed = is_active.unwrap_or_default();
```

jgrund (dismissed):

iml-api/src/graphql.rs
ip1981: Both variants aren't obvious :)

iml-api/src/graphql.rs
```diff
@@ -397,6 +398,79 @@ impl QueryRoot {
...
+                            .unwrap_or_default()
+                            .into_iter()
+                            .map(|job_id: i32| {
+                                format!("/api/{}/{}/", Job::<()>::endpoint_name(), job_id)
+                            })
+                            .collect::<Vec<_>>()
+                    },
+                    logs: "".to_string(),
+                    message: x.message.clone(),
+                    resource_uri: format!("/api/{}/{}/", Command::endpoint_name(), x.id),
```
ip1981, line 457: Looks like a lot of duplication. Is it possible to extend the `commands` query to support optional id list?

nlinker: I tried, this `x` has very complicated synthetic type, derived by `sqlx` macro. Can be worked around with a transient local type though, say `TransientCommand`:
```
        let snapshots = sqlx::query_as!(TransientCommand, "SQL goes here")
              .fetch_all(&pool)
              .map(convertToCommand)
```

nlinker: Or `CommandTemp`

iml-api/src/graphql.rs
nlinker: 1. Does it make sense to rename the argument to `is_complete` then to reduce the "impedance mismatch" or leave it `is_active`? What is more natural/frequent use case?

2. Also having it optional makes us to completely ignore the filter, so
- `commands(isComplete: true)` returns completed commands
- `commands(isComplete: false)` returns incomplete commands
- `commands()` returns all commands
It sounds as an unnecessary feature, right?

jgrund: I think the most common case will be looking for active commands

nlinker: Screenshot 2020-09-23 at 08 43 27

mkpankov (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2265)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @ip1981, @jgrund, and @nlinker)

___
*[iml-api/src/graphql.rs, line 396 at r4](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2265#-MHtsuPD0XvBuk4pDC8_:-MHtsuPD0XvBuk4pDC8a:bszjouh) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/f2c4ec79191b1f2bb65d01d611560cf7876aa661/iml-api/src/graphql.rs#L396)):*
> ```Rust
>         limit: Option<i32>,
>         offset: Option<i32>,
>         ids: Vec<i32>,
> ```

I'd like to make it an optional argument to `commands`. This feels like special-casing `id` field while in fact it's just another field of entity we're querying.

nlinker:
> ```rust
>         limit: Option<i32>,
>         offset: Option<i32>,
>         ids: Vec<i32>,
> ```
> I'd like to make it an optional argument to `commands`. This feels like special-casing `id` field while in fact it's just another field of entity we're querying.

It is indeed a special case, it is not just another filtering criteria, there are several reasons:

1. From the application POV it is pretty strange to get entities with the specified ids and messages like something. There is nothing like "approximate id", while approximate messages are. If we are include `ids` filter into `commands` method, then there will be two huge cases inside: one optional is fuzzy match based on some fields, with ordering needed and grouping, and optional exact match based on `id` optimized by performance.

2. From the interface POV `commands` by fuzzy filter returns several commands list or just empty list. The ordering reflects the will of the end user. Commands by the exact search should convey also the information in the easy-to-use form to the code, that requests it. So if some of the ids are absent, the user wants it just filtered out, but the code needs it to be detected. 
Hence the fuzzy search has the type `FuzzyFilter -> Vec<Command>` and exact search has the type `Ids -> Vec<Option<Command>>`, the results of the exact search are easy to be consumed with `map`s, `flat_map`s and `filter`s.

3. From the performance POV exact and fuzzy search are different — there are no need for `order by` and `like`. And to tackle the `N+1` problem (I'm not sure, we will face it soon, but will eventually!) we must have a method, something of type `Vec<Id> -> Vec<Option<Command>>` and use https://docs.rs/crate/juniper-eager-loading/0.5.1 library to reduce roundtrips.

A good example of the pretty complex GraphQL API is https://github.com/nodkz/wrike-graphql (russian docs, sorry), you can see the patterns:
```sdl
  taskByIds(ids: [TaskID!]!): [Task!]
  taskFindMany(limit: Int, pageSize: Int, nextPageToken: String, filter: TaskFindManyFilter, sort: TaskFindManySortEnum): [Task!]
...
  timelogByIds(ids: [TimelogID!]!, plainText: Boolean): [Timelog!]
  timelogFindMany(filter: TimelogFindManyFilter, plainText: Boolean! = false): [Timelog!]
...
  userById(id: ContactID!): User
  userScheduleExclusionFindById(id: UserScheduleExclusionID!): UserScheduleExclusion
```
@ip1981 @jgrund @mkpankov If you find all the above irrelevant to the current task, I'll glue the methods `commands` and `commands_by_id`. Thanks.

mkpankov:
@nlinker you convinced me

mkpankov (approved):

ip1981:
Well, this is exactly the part of GraphQL we miss - dynamic queries :)

From the client point, a single query seems more reasonable:

```graphql
commands(ids: [1,2,3]) {
...
}
```

or 
```graphql
commands(msg: "Foo") {
...
}
```

So I am in favor of a single entry point, but with flexible backend if needed.

jgrund:
> So I am in favor of a single entry point, but with flexible backend if needed.

In general this idea would allow for more flexibility on the interface, but means we have to resort to dynamic query building and lose out on the compile-time checking of `query!` / `query_as!`. The other option is to run different `query!` / `query_as!`  dependent on the passed in params, which also is not ideal.

iml-api/src/graphql.rs
```diff
@@ -357,46 +359,81 @@ impl QueryRoot {
...
+        limit(description = "paging limit, defaults to 20"),
+        offset(description = "Offset into items, defaults to 0"),
+        ids(description = "The list of command ids to fetch, ids may be empty"),
+    ))]
+    async fn commands_by_ids(
+        context: &Context,
+        limit: Option<i32>,
+        offset: Option<i32>,
+        ids: Vec<i32>,
+    ) -> juniper::FieldResult<Vec<Option<Command>>> {
```
jgrund, line 397: nit:

If we pass in an invalid id, it might be better to just return an error. With this approach, we will have to `.filter_map` every result we get back from this call.

iml-api/src/graphql.rs
```diff
@@ -734,6 +771,37 @@ impl MutationRoot {
...
+    pub cancelled: bool,
+    pub complete: bool,
+    pub created_at: DateTime<Utc>,
+    pub errored: bool,
+    pub id: i32,
+    pub job_ids: Option<Vec<i32>>,
+    pub message: String,
+}
+
+fn to_command(x: CommandTmpRecord) -> Command {
```
jgrund, line 785: nit: could use `From` trait here.

jgrund (approved):

iml-api/src/graphql.rs
nlinker: I tried, It was a little magical.

iml-api/src/graphql.rs
nlinker: I'll tackle it in follow up PR.

PR #2269, opened 2020-09-23
GH-2268 Update the limit param in graphql api's

### LINKED ISSUE #2268: Update the limit param in graphql api's
Currently, the `limit` param allows you to specify a limit when retrieving items back from an API call. This is a numeric value that can be passed in but in some cases, we may want all results. To address this, postress allows us to pass in `ALL` instead of a numeric value, which is effectively the same as not passing a limit at all. Update all of the graphql API endpoints such that if `None` is passed in it will pass `ALL` to the LIMIT clause in the query. This will allow us to get all entries without modifying the existing api.

johnsonw:
Fixes #2268.

Currently, the limit param allows you to specify a limit when retrieving items back from an API call. This is a numeric value that can be passed in but in some cases, we may want all results. To address this, postress allows us to pass in ALL instead of a numeric value, which is effectively the same as not passing a limit at all. Update all of the graphql API endpoints such that if None is passed in it will pass ALL to the LIMIT clause in the query. This will allow us to get all entries without modifying the existing api.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2269)

### DIFF
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -136,7 +136,7 @@ pub(crate) struct QueryRoot;
 #[juniper::graphql_object(Context = Context)]
 impl QueryRoot {
     #[graphql(arguments(
-        limit(description = "paging limit, defaults to 20"),
+        limit(description = "paging limit, defaults to All"),
         offset(description = "Offset into items, defaults to 0"),
         dir(description = "Sort direction, defaults to asc")
     ))]
@@ -184,7 +184,7 @@ impl QueryRoot {
     }
 
     #[graphql(arguments(
-        limit(description = "paging limit, defaults to 20"),
+        limit(description = "paging limit, defaults to All"),
         offset(description = "Offset into items, defaults to 0"),
         dir(description = "Sort direction, defaults to asc"),
         fs_name(description = "Targets associated with the specified filesystem"),
@@ -293,7 +293,7 @@ impl QueryRoot {
     }
 
     #[graphql(arguments(
-        limit(description = "paging limit, defaults to 20"),
+        limit(description = "paging limit, defaults to All"),
         offset(description = "Offset into items, defaults to 0"),
         dir(description = "Sort direction, defaults to ASC"),
         fsname(description = "Filesystem the snapshot was taken from"),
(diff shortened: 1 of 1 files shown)


iml-api/src/graphql.rs
```diff
@@ -173,7 +173,7 @@ impl QueryRoot {
                 OFFSET $2 LIMIT $3"#,
             dir.deref(),
             offset.unwrap_or(0) as i64,
-            limit.unwrap_or(20) as i64
+            limit.map(|x| x.to_string()).unwrap_or("ALL".to_string()) as String
```
ip1981, line 176:
```suggestion
            limit.map(|x| x.to_string()).unwrap_or_else(|| "ALL".to_string()) as String
```

johnsonw: sure, makes sense. I'll update.

jgrund (changes requested):
Comments need updating as well, as they still say they default to 20.

iml-api/src/graphql.rs
```diff
@@ -173,7 +173,9 @@ impl QueryRoot {
                 OFFSET $2 LIMIT $3"#,
             dir.deref(),
             offset.unwrap_or(0) as i64,
-            limit.unwrap_or(20) as i64
+            limit
+                .map(|x| x.to_string())
+                .unwrap_or_else(|| "ALL".to_string()) as String
```
ip1981, line 178: `as String` really needed?

johnsonw: It required the cast.

iml-api/src/graphql.rs
```diff
@@ -136,7 +136,7 @@ pub(crate) struct QueryRoot;
 #[juniper::graphql_object(Context = Context)]
 impl QueryRoot {
     #[graphql(arguments(
-        limit(description = "paging limit, defaults to 20"),
+        limit(description = "paging limit, defaults to All"),
```
jgrund, line 139: Maybe: 

```suggestion
        limit(description = "optional paging limit, defaults to all rows"),
```

jgrund (approved):

ip1981 (approved):

Issue #2248, opened 2020-09-16
Snapshots should not disappear from the DB


ip1981:
Snapshots should not disappear from the DB due to errors.

For example, when listing cannot be performed due to other operations running simultaneously (creating/destroying), we send and empty list of snapshots and get all known snapshots removed to the DB.

Caveats:

1) some FSs may not have snapshots enabled at all, so listing for them will always fail.
2) mounted snapshots are seen as regular FSs, listing for them will fail as well.

jgrund:
Is there a different error code returned when the resource is busy?

ip1981:
No, it's always 1 AFAIK

jgrund:
so it's only the error string we can match on?

ip1981:
> so it's only the error string we can match on?

Yes.

jgrund:
yeah:

```
lctl snapshot_list -F zfsmo
Can't lock the snapshot config file /etc/ldev.conf (1): Resource temporarily unavailable
[root@mds1 ~]# echo $?
1
```

jgrund:
Sounds like that's what we should do then.

ip1981:
At least the "Resource temporarily unavailable" message must be stable enough :)

jgrund:
Replaced by #2364

PR #2282, opened 2020-09-29
Analyze lctl output for real snapshot listing failures

ip1981:
Closing #2248.

### DIFF
--- a/iml-agent/src/daemon_plugins/snapshot.rs
+++ b/iml-agent/src/daemon_plugins/snapshot.rs
@@ -20,6 +20,7 @@ use futures::{
     lock::Mutex,
     Future, FutureExt,
 };
+use iml_cmd::CmdError;
 use iml_wire_types::snapshot::{List, Snapshot};
 use std::collections::BTreeSet;
 use std::{pin::Pin, sync::Arc, time::Duration};
@@ -50,7 +51,7 @@ async fn list() -> Result<Vec<Snapshot>, ()> {
     let fss: Vec<String> = lctl(vec!["get_param", "-N", "mgs.MGS.live.*"])
         .await
         .map_err(|e| {
-            // XXX debug because of false positives
+            // XXX debug because of false positives. But this is still a failure.
             tracing::debug!("listing filesystems failed: {}", e);
         })
         .map(|o| {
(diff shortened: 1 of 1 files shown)


iml-agent/src/daemon_plugins/snapshot.rs
```diff
@@ -87,14 +88,37 @@ async fn list() -> Result<Vec<Snapshot>, ()> {
...
+                return false;
+            }
+
+            match &x.1 {
+                ImlAgentError::CmdError(CmdError::Output(o)) => {
+                    // XXX lctl returns 1 no matter what, so have to read its output:
+                    let stderr = String::from_utf8_lossy(&o.stderr);
+                    if stderr.find("Resource temporarily unavailable").is_some() {
+                        true
+                    } else if stderr.find("Miss MDT0 in the config file").is_some() {
```
jgrund, line 105: Clippy warning here

ip1981: Maybe the first time I am inclined to disagree with clippy, but there is a better way though.

ip1981:
It may be a good idea to catch "resource temporarily unavailable" at lctl call here:
https://github.com/whamcloud/integrated-manager-for-lustre/blob/2cd9ed8465aee05e85b4afebc199da34888469ec/iml-agent/src/lustre.rs#L13-L24

and then retry.

So all command would work more reliably.

iml-agent/src/lustre.rs
```diff
@@ -4,23 +4,38 @@
...
-        .err_into()
-        .await
-        .map(|o| String::from_utf8_lossy(&o.stdout).to_string())
+    loop {
+        let r = Command::new("/usr/sbin/lctl")
+            .args(args.clone())
+            .checked_output()
+            .await;
+
+        if let Err(CmdError::Output(o)) = &r {
```
jgrund, line 26: It seems like this should wrap the `lctl` call as a new fn instead of being within it.

ip1981: I did it first, but the function was private anyway.

ip1981: Done

ip1981:
I successfully created few snapshots from two terminals at the same time :)

iml-agent/src/lustre.rs
jgrund: Maybe call this fn `lctl_retry`, and keep the original `lctl` fn in place. the `lctl` fn is being used in a few other modules already, and it's behavior should stay the same I think.

ip1981: Actually, other modules will just become more sound out of the box. I don't think there is sense in failing for a temporarily unavailable resource.

jgrund: Have we seen this on any other lctl call (and do we know the behavior is the same)?

ip1981: Ah, that's valid question. As a precaution we indeed can have two version of lctl function. But the lctl executable is using perror() or strerror() consistently. And retrying temporary failures is still needed.

It is really simple: we would loop iff lctl prints "Resource temporarily unavailable" (to stderr + exit code, it could not be a snapshot comment :) )

beevans: Maybe add the call into IML's liblustreapi library?  Might be cleaner than doing a shell command.

ip1981: Make it `lctl_retry`

mkpankov (approved):

jgrund:
Replaced by #2364

PR #2285, opened 2020-09-30
Add shell completions for CLI

### LINKED ISSUE #2283: Generate and ship shell completion scripts
Generate and ship shell completion scripts.

See https://docs.rs/clap/2.33.3/clap/struct.App.html#method.gen_completions.

ip1981:
Closes #2283 
Screenshot_2020-09-30_18-11-02

Screenshot_2020-10-01_19-27-29

Not sure if it's worth or possible to hide "hidden" options.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2285)

### DIFF
--- a/iml-manager-cli/src/main.rs
+++ b/iml-manager-cli/src/main.rs
@@ -13,6 +13,7 @@ use iml_manager_cli::{
 };
 
 use std::process::exit;
+use structopt::clap::Shell;
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
@@ -54,6 +55,16 @@ pub enum App {
     #[structopt(name = "debugql", setting = structopt::clap::AppSettings::Hidden)]
     /// Direct GraphQL Access (for testing and debug)
     DebugQl(api::GraphQlCommand),
+
+    #[structopt(name = "shell-completion", setting = structopt::clap::AppSettings::Hidden)]
+    /// Generate shell completion script
+    Shell {
+        shell: Shell,
+        #[structopt(short = "e", long = "executable", default_value = "iml")]
+        exe: String,
+        #[structopt(short = "o", long = "output")]
+        output: Option<String>,
+    },
 }
 
 #[tokio::main]
@@ -64,7 +75,12 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     tracing::debug!("Matching args {:?}", matches);
 
-    dotenv::from_path("/var/lib/chroma/iml-settings.conf").expect("Could not load cli env");
+    match matches {
+        App::Shell { .. } => (),
+        _ => {
+            dotenv::from_path("/var/lib/chroma/iml-settings.conf").expect("Could not load cli env")
+        }
+    }
 
     let r = match matches {
         App::DebugApi(command) => api_cli(command).await,
@@ -74,6 +90,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
         App::Snapshot { command } => snapshot_cli(command).await,
         App::Stratagem { command } => stratagem_cli(command).await,
         App::UpdateRepoFile(config) => update_repo_file_cli(config).await,
+        App::Shell { shell, exe, output } => {
+            if let Some(out) = output {
+                let mut o = std::fs::File::create(out)?;
+                App::clap().gen_completions_to(exe, shell, &mut o);
+            } else {
+                App::clap().gen_completions_to(exe, shell, &mut std::io::stdout());
+            };
+            Ok(())
+        }
     };
 
     if let Err(e) = r {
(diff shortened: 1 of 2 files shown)


iml-manager-cli/src/main.rs
```diff
@@ -64,7 +75,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     tracing::debug!("Matching args {:?}", matches);
 
-    dotenv::from_path("/var/lib/chroma/iml-settings.conf").expect("Could not load cli env");
+    match matches {
+        App::Shell {
+            shell: _,
+            exe: _,
+            output: _,
```
ip1981, line 82: Is there a way to omit all the fields?

rust-iml.spec
```diff
@@ -74,6 +74,8 @@ mkdir -p %{buildroot}%{_presetdir}
 cp 00-rust-iml-agent.preset %{buildroot}%{_presetdir}
 mkdir -p %{buildroot}%{_sysconfdir}/iml/
 cp settings.conf %{buildroot}%{_sysconfdir}/iml/iml-agent.conf
+mkdir -p %{buildroot}%{_sysconfdir}/bash_completion.d
+%{buildroot}%{_bindir}/iml shell-completion bash -e iml -o %{buildroot}%{_sysconfdir}/bash_completion.d/iml
```
jgrund, line 78: Should we allow users to do this instead of doing it for them?

ip1981: AFAIK, completions are usually packaged alone with the binaries. So it could just work.

jgrund: Would probably be better as a standalone RPM with a dependency on this package.

ip1981: Zsh users are free to do it themselves :)

jgrund: I.E. 

```
%package cli-bash-completions
Requires: rust-iml-cli
...
```

ip1981: Done!

jgrund:
```suggestion
```

This should be done in the `%post` scriptlet of `cli-bash-completion` instead.

ip1981: But then the file will not removed on removal, will it?

ip1981: And if in post-scripts, then I don't see a point in a separate package at all.

jgrund: > And if in post-scripts, then I don't see a point in a separate package at all.

Why? it gives the user a way to decide whether they want completions installed or not.

jgrund: > But then the file will not removed on removal, will it?

`%postun`?

mkpankov (approved):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2285)* status: all files reviewed, 2 unresolved discussions (waiting on @ip1981)

rust-iml.spec
ip1981: If the file is created in scripts, it will be hard to figure out which package is responsible for it, right?
And then it will be almost the same as letting users execute `iml shell-completion ...` on their own, even in home dir,
In addition, they will have to install the `bash-completion` package anyway.

Also generating the completion file when building a package should be more reliable, because we can catch any issue earlier and there will be no risk when installing the package.

Note that bash completions are shipped along with the executables (as well as man pages, etc.), except Django :)
```
# rpm -qf /etc/bash_completion.d/
filesystem-3.2-25.el7.x86_64
yum-utils-1.1.31-52.el7.noarch
iprutils-2.4.17.1-2.el7.x86_64
createrepo-0.9.9-28.el7.noarch

# rpm -qf /usr/share/bash-completion
glib2-2.56.1-5.el7.x86_64
yum-3.4.3-163.el7.centos.noarch
python-django-bash-completion-1.11.27-1.el7.noarch
```

So there are four options in the UX descending order (IMHO):
1. Same package as the executable.
2. Separate package (+ one for zsh, etc.), this might be a good idea.
3. Separate package with scripts (which will have to be coordinated with each other).
4. Manual `iml shell-completion ...`.

Open question is Docker packaging, but at least number 4 should will do, and 3 seems the worst.

jgrund: I'm a bit weary of executing the binary during build time, but I suppose hitting issues earlier is a fair point. Let's go with #2 including separate packages for ZSH.

I think we will need to revisit how we do CLI in docker generally, but not something for this issue.

iml-manager-cli/src/main.rs
```diff
@@ -64,7 +75,16 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+        } => (),
```
jgrund, line 83:
```suggestion
        App::Shell { .. } => (),
```

Should work I think

jgrund (dismissed):

mkpankov (approved):
> Reviewed 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2285)* status: all files reviewed, 3 unresolved discussions (waiting on @ip1981)

jgrund (approved):

PR #2289, opened 2020-09-30
2288: remove "pull" as an option for filesync

beevans:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2289)

### DIFF
--- a/iml-agent/src/cli.rs
+++ b/iml-agent/src/cli.rs
@@ -545,6 +545,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 target_fs,
                 files,
             } => {
+                if action == action_filesync::ActionType::Pull {
+                    tracing::error!("Pull is not available as an option");
+                    exit(exitcode::USAGE);
+                }
                 let llapi =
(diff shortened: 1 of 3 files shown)


iml-agent/src/cli.rs
```diff
@@ -545,6 +545,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 target_fs,
                 files,
             } => {
+                if action == action_filesync::ActionType::Pull {
+                    tracing::error!("Pull is not available as an option");
```
jgrund, line 549: Is pull just not supported in the CLI, or not at all?

beevans: Not in V1, but there's a bunch of code in the back end (action_filesync.rs) that's already written, so I'd prefer to do some minor changes to disable rather than write and replace.

utopiabound (approved):

jgrund (approved):
nit: it would be better to remove the other enum type instead of failing in the code.

PR #2290, opened 2020-09-30
chroma: Add graphql helper function

utopiabound:
Use graphql for fs.get_server_groups()

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2290)


chroma_core/lib/graphql.py
```diff
@@ -0,0 +1,27 @@
...
+    s.headers.update({"AUTHORIZATION": "ApiKey {}:{}".format(API_USER, API_KEY)})
+    return s
+
+
+def graphql_query(query):
+    """
+    Issue a GraphQL query
+    """
+    s = get_api_session_request()
+    path = os.path.join(settings.IML_API_PROXY_PASS, "graphql")
```
johnsonw, line 23: Isn't this going to go to the API port? I thought a call to graphql would be something like this:
```
https://127.0.0.1:8443/graphql/
```
Is this wrong?

chroma_core/models/filesystem.py
```diff
@@ -109,6 +109,15 @@ def get_servers(self):
         # (http://stackoverflow.com/questions/4764110/django-template-cant-loop-defaultdict)
         return dict(servers)
 
+    def get_server_groups(self):
```
johnsonw, line 112: Is this needed in this patch?

chroma_core/lib/graphql.py
```diff
@@ -0,0 +1,27 @@
...
+    s = requests.Session()
+    s.headers.update({"AUTHORIZATION": "ApiKey {}:{}".format(API_USER, API_KEY)})
+    return s
+
+
+def graphql_query(query):
+    """
+    Issue a GraphQL query
+    """
+    s = get_api_session_request()
```
jgrund, line 22: This is going to add 1 more db connection / query per-request. This should instead by saved to a module-level variable that is first checked and if empty does this.

jgrund (changes requested):
This will add more db overhead than needed. use a module-level variable to cache the results.

chroma_core/models/filesystem.py
utopiabound: Helpful "example"?  (I've been trying to get this function in for a while) I need it for hotpools

chroma_core/lib/graphql.py
utopiabound: okay.

chroma_core/models/filesystem.py
johnsonw: Ah gotcha. Ok just wondering.

chroma_core/lib/graphql.py
johnsonw: I was wondering about this as well. We also have /var/lib/chroma/iml-settings.conf that we could read from and cache. I guess if it's cached it shouldn't make much of a difference though.

jgrund: `/var/lib/chroma/iml-settings.conf` will mean introducing env vars into the python services, which shouldn't be necessary.

chroma_core/lib/graphql.py
utopiabound: graphql is handled by `iml-api` and if I go to https, I get TLS errors.

johnsonw: Neat, I didn't realize that. I've always made my requests like this:
```
curl -k "https://127.0.0.1:443/graphql/"  
    -X POST
    -d "{ \"query\": \"{ targets { state name activeHostId hostIds filesystems uuid mountPath } }\" }"
    -H "Authorization: ApiKey API:<key>"
    -H "Content-Type: application/json"
```

jgrund: The advantage of going through nginx is that it will provide ApiKey auth (which won't be provided when going directly to the service).

johnsonw (dismissed):

chroma_core/lib/graphql.py
jgrund: I have a feeling this isn't going to work in docker.

chroma_core/lib/graphql.py
```diff
@@ -5,11 +5,16 @@
...
+API_USER = None
+API_KEY = None
+
+
 def get_api_session_request():
-    cursor = connection.cursor()
-    cursor.execute("SELECT * from api_key()")
-    (API_USER, API_KEY) = cursor.fetchone()
-    cursor.close()
+    if not API_USER or not API_KEY:
```
jgrund, line 13: pretty sure these need `global` keyword:

```
>>> FOO = None
>>> def bar():
...    if not FOO:
...        FOO='bar'
...
>>> bar()
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in bar
  UnboundLocalError: local variable 'FOO' referenced before assignment

```

jgrund (changes requested):
Current change will not work due to python scoping rules.

In addition, I think use of `IML_API_PROXY_PASS` is going to cause issues in docker.

chroma_core/lib/graphql.py
jgrund: Inside the job-scheduler container:

```
>>> settings.IML_API_PROXY_PASS
'http://iml-action-runner:8004'
```

johnsonw:
Could probably do something like:
```
result = requests.put("{}/graphql/".format(SERVER_HTTP_URL), json=data, verify=False)
```

chroma_core/models/filesystem.py
```diff
@@ -109,6 +109,15 @@ def get_servers(self):
...
+        """Return: Array(Array(ManagedHost)) """
+        from chroma_core.lib.graphql import graphql_query
+        from chroma_core.models.host import ManagedHost
```
jgrund, line 115:
```suggestion
```

utopiabound:
> In addition, I think use of `IML_API_PROXY_PASS` is going to cause issues in docker.

So `IML_API_PROXY_PASS` is wrong in settings.py in docker? should we override it like `PROXY_HOST`?

chroma_core/lib/graphql.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+API_USER = None
+API_KEY = None
+
+
+def get_api_session_request():
+    global API_USER
+    global API_KEY
+    if not API_USER or not API_KEY:
```
jgrund, line 14: nit: tuple of `API_CREDS = (user, key)` would be a bit easier to work with

johnsonw (approved):
I think this looks good. Has it been tested yet for both rpm and docker?

johnsonw:
Codefactor doesn't like the global statements. Is there a way to ignore that?

chroma_core/lib/graphql.py
utopiabound: > I think this looks good. Has it been tested yet for both rpm and docker?

I've manually tested the logic on RPM system.

utopiabound:
> I think this looks good. Has it been tested yet for both rpm and docker?

I've manually tested the logic on an RPM system.

johnsonw:
I installed the docker build on my onyx node and ran a quick test. Specifically, I ssh'd into the job scheduler container and opened a python REPL, imported `graphql.py` and ran the following:

```
>>> from chroma_core.lib.graphql import graphql_query
>>> graphql_query('{targets(fsName: "fs") {state,name,activeHostId,hostIds}}')
{u'targets': [{u'state': u'mounted', u'hostIds': [1, 2], u'name': u'MGS', u'activeHostId': 1}, {u'state': u'mounted', u'hostIds': [1, 2], u'name': u'fs-MDT0000', u'activeHostId': 1}, {u'state': u'mounted', u'hostIds': [2, 1], u'name': u'fs-MDT0001', u'activeHostId': 2}, {u'state': u'mounted', u'hostIds': [3, 4], u'name': u'fs-OST0000', u'activeHostId': 3}, {u'state': u'mounted', u'hostIds': [3, 4], u'name': u'fs-OST0001', u'activeHostId': 3}, {u'state': u'mounted', u'hostIds': [3, 4], u'name': u'fs-OST0002', u'activeHostId': 3}, {u'state': u'mounted', u'hostIds': [3, 4], u'name': u'fs-OST0003', u'activeHostId': 3}, {u'state': u'mounted', u'hostIds': [3, 4], u'name': u'fs-OST0004', u'activeHostId': 3}, {u'state': u'mounted', u'hostIds': [4, 3], u'name': u'fs-OST0005', u'activeHostId': 4}, {u'state': u'mounted', u'hostIds': [4, 3], u'name': u'fs-OST0006', u'activeHostId': 4}, {u'state': u'mounted', u'hostIds': [4, 3], u'name': u'fs-OST0007', u'activeHostId': 4}, {u'state': u'mounted', u'hostIds': [4, 3], u'name': u'fs-OST0008', u'activeHostId': 4}, {u'state': u'mounted', u'hostIds': [4, 3], u'name': u'fs-OST0009', u'activeHostId': 4}]}
```

johnsonw (approved):

mkpankov (approved):

chroma_core/lib/graphql.py
```diff
@@ -0,0 +1,31 @@
...
+    s.headers.update({"AUTHORIZATION": "ApiKey {}:{}".format(this.API_CRED[0], this.API_CRED[1])})
+    return s
+
+
+def graphql_query(query):
+    """
+    Issue a GraphQL query
+    """
+    s = get_api_session_request()
+    path = os.path.join(settings.IML_API_PROXY_PASS, "graphql")
```
jgrund, line 28: nit, since you talk directly to the api container (bypassing nginx), the api_key is not needed.

PR #2291, opened 2020-10-01
Create graphql_map scalar type

johnsonw:
Since graphql doesn't support a map type, we can create a scalar that
will use serde to serialize and deserialze map data.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2291)


nlinker (changes requested):
Is it correct behavior? I think it breaks the compatibility, no?
image

jgrund:
Closing this. I think any map variant we use will need to be able to be input by users and not just (de)serialized by rust.

PR #2304, opened 2020-10-02
EX-1850 - Implement Report query and deletion

johnsonw:
Implement Report query and deletion in graphql

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2304)

### DIFF
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -494,12 +494,15 @@ impl QueryRoot {
             .try_collect::<Vec<String>>()
             .await?
             .into_iter()
-            .map(|filename| fs::canonicalize(get_report_path().join(filename)));
+            .map(|filename| {
+                fs::canonicalize(get_report_path().join(filename.clone()))
+                    .map_ok(|file_path| (file_path, filename))
+            });
 
         let items = try_join_all(file_paths)
             .await?
             .into_iter()
-            .map(|x| x.to_string_lossy().to_string())
+            .map(|(file_path, filename)| (file_path.to_string_lossy().to_string(), filename))
             .map(get_stratagem_files);
 
         let items = try_join_all(items).await?;
@@ -1122,11 +1125,13 @@ async fn fqdn_by_host_id(pool: &PgPool, id: i32) -> Result<String, iml_postgres:
     Ok(fqdn)
 }
 
-async fn get_stratagem_files(file_path: String) -> juniper::FieldResult<StratagemReport> {
+async fn get_stratagem_files(
+    (file_path, filename): (String, String),
+) -> juniper::FieldResult<StratagemReport> {
     let attr = fs::metadata(file_path.to_string()).await?;
 
     Ok(StratagemReport {
-        filename: file_path.to_string(),
+        filename,
         modify_time: attr.modified()?.into(),
         size: attr.len() as i32,
     })
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -468,12 +468,13 @@ impl QueryRoot {
 
         Ok(xs)
     }
-    /// List Stratagem reports
+    /// List completed Stratagem reports that currently reside on the manager node.
+    /// Note: All report names must be valid unicode.
     async fn stratagem_reports(_context: &Context) -> juniper::FieldResult<Vec<String>> {
         let paths = tokio::fs::read_dir(get_report_path()).await?;
 
         let items: Vec<String> = paths
-            .map_ok(|x| x.file_name().into_string().expect("DirEntry to string"))
+            .map_ok(|x| x.file_name().to_string_lossy().to_string())
             .try_collect()
             .await?;
 
@@ -832,8 +833,12 @@ impl MutationRoot {
         _context: &Context,
         filename: String,
     ) -> juniper::FieldResult<bool> {
-        let mut path = get_report_path();
-        path.push(filename);
+        let report_base = get_report_path();
+        let path = tokio::fs::canonicalize(report_base.join(filename)).await?;
+
+        if !path.starts_with(report_base) {
+            return Err(FieldError::new("Invalid path", Value::null()));
+        }
 
         tokio::fs::remove_file(path).await?;
 
--- a/iml-api/src/graphql.rs
+++ b/iml-api/src/graphql.rs
@@ -9,6 +9,7 @@ use crate::{
 };
 use chrono::{DateTime, Utc};
 use futures::{future::join_all, TryFutureExt, TryStreamExt};
+use iml_manager_env::get_report_path;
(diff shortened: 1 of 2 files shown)


johnsonw:
image

image

```
[root@adm report]# touch file1.txt
[root@adm report]# ls
file1.txt
[root@adm report]# ls
[root@adm report]#
```

iml-api/src/graphql.rs
```diff
@@ -813,6 +828,20 @@ impl MutationRoot {
...
     }
+
+    /// Delete a stratagem report
+    #[graphql(arguments(filename(description = "The report filename to delete")))]
+    async fn delete_stratagem_report(
+        _context: &Context,
+        filename: String,
+    ) -> juniper::FieldResult<bool> {
+        let mut path = get_report_path();
+        path.push(filename);
```
jgrund, line 839: What happens if filename is `../../*`?

iml-api/src/graphql.rs
```diff
@@ -467,6 +469,19 @@ impl QueryRoot {
...
         Ok(xs)
     }
+    /// List Stratagem reports
+    async fn stratagem_reports(
+        _context: &Context,
+    ) -> juniper::FieldResult<Vec<String>> {
+        let paths = tokio::fs::read_dir(get_report_path()).await?;
+
+        let items: Vec<String> = paths
+            .map_ok(|x| x.file_name().into_string().expect("DirEntry to string"))
```
jgrund, line 479: Seems like we should return the error instead of panicking here.

iml-api/src/graphql.rs
johnsonw: right, I'm currently thinking about that but wanted to push a first draft. Do you have any initial thoughts?

iml-api/src/graphql.rs
johnsonw: I'm trying to think of the best way to do this. So far I came up with:
```
let items: Vec<String> = paths
            .map_ok(|x| {
                x.file_name().into_string().map_or_else(
                    |e| format!("Error getting stratagem report file name: {:?}", e),
                    |x| x,
                )
            })
            .try_collect()
            .await?;
```
However, I feel like I'm missing something..

jgrund: Pushed an update.

iml-api/src/graphql.rs
jgrund: Pushed an update.

johnsonw:
Docker test:

Displaying files:
image

[root@iml-api report]# ls
[root@iml-api report]# ls
expiring_fids-fs-32621d4c-9b38-4d22-8d4a-9a45d96f5667.txt

Deleting a file:
image
```
[root@iml-api report]# ls
expiring_fids-fs-52ee437c-646a-402a-9df3-e90810a77a68.txt
[root@iml-api report]# ls
[root@iml-api report]#
```

johnsonw:
Trying to delete a file maliciously:

image

mkpankov (approved):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2304)* status: all files reviewed, 2 unresolved discussions (waiting on @johnsonw)

johnsonw:
Integration tests passed here: https://build.whamcloud.com/view/IML/job/manager-for-lustre/2489/

johnsonw:
Also, I've updated the query to return both the filesize and modified time. The data returned now looks like this:
```
{
  "data": {
    "stratagemReports": [
      {
        "filename": "/var/spool/iml/report/test.txt",
        "modifyTime": "2020-10-05T21:51:57.887528595+00:00",
        "size": 0
      }
    ]
  }
}
```

johnsonw:
The CodeQL failure is unrelated and I see it in other patches as well.

jgrund:
Does this now print the whole file path or just the name?

johnsonw:
It's printing the whole file path. I thought it might be useful to show the whole path in case they weren't aware of the report being under /var/spool/iml. However, we could always print a note above indicating where the file live and then just show the filename. What are your thoughts?

jgrund:
Probably just the filename. The reports can be downloaded at /report/<filename>, so it's probably not important to know the full path. In addition, if we are running in containers, that path will exist inside a container only and will be confusing to users that it's not on the host.

johnsonw:
Sure, makes sense. I will update it to be the filename only.

johnsonw:
Here is an updated screenshot. I created three files under /var/spool/iml/report. Notice that only the filename is now displayed:
test1.txt = 1Kib
test2.txt = 1Mib
test3.txt = 1Gib

image

iml-api/src/graphql.rs
```diff
@@ -813,6 +855,24 @@ impl MutationRoot {
...
+    #[graphql(arguments(filename(description = "The report filename to delete")))]
+    async fn delete_stratagem_report(
+        _context: &Context,
+        filename: String,
+    ) -> juniper::FieldResult<bool> {
+        let report_base = get_report_path();
+        let path = tokio::fs::canonicalize(report_base.join(filename)).await?;
+
+        if !path.starts_with(report_base) {
+            return Err(FieldError::new("Invalid path", Value::null()));
```
ip1981, line 869: nit:
```suggestion
            return Err(FieldError::new("Invalid filename", Value::null()));
```

jgrund (approved):

iml-api/src/graphql.rs
```diff
@@ -467,6 +483,32 @@ impl QueryRoot {
...
+    async fn stratagem_reports(_context: &Context) -> juniper::FieldResult<Vec<StratagemReport>> {
+        let paths = tokio::fs::read_dir(get_report_path()).await?;
+
+        let file_paths = paths
+            .map_ok(|x| x.file_name().to_string_lossy().to_string())
+            .try_collect::<Vec<String>>()
+            .await?
+            .into_iter()
+            .map(|filename| {
+                fs::canonicalize(get_report_path().join(filename.clone()))
```
ip1981, line 498: Could call `get_report_path()` juts once at the beginning?

iml-api/src/graphql.rs
```diff
@@ -467,6 +483,32 @@ impl QueryRoot {
 
         Ok(xs)
     }
+
+    /// List completed Stratagem reports that currently reside on the manager node.
+    /// Note: All report names must be valid unicode.
+    async fn stratagem_reports(_context: &Context) -> juniper::FieldResult<Vec<StratagemReport>> {
+        let paths = tokio::fs::read_dir(get_report_path()).await?;
```
ip1981, line 490: nit:
```suggestion
        let paths = fs::read_dir(get_report_path()).await?;
```

iml-api/src/graphql.rs
ip1981, line 498: Is clone() needed?

PR #2305, opened 2020-10-05
Allow different names for IML CLI

### LINKED ISSUE #2295: Make it possible to have IML CLI under different names
1. Make help text more neutral (avoid "IML")
2. Make generic description configurable / automatic:
```
foo 0.3.0
The Foo Bar for All CLI
```
3. A symbolic link or a shell wrapper might work just fine, shipped in a separate package:

Screenshot_2020-10-01_19-03-22

ip1981:
- Determine executable name.

- Remove "IML" where possbile.

Screenshot_2020-10-06_15-21-03

Separate packages would create a hard link for the executable and symlinks for shell completions.

Closes #2295.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2305)

### DIFF
--- a/iml-manager-cli/src/lib.rs
+++ b/iml-manager-cli/src/lib.rs
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a MIT-style
 // license that can be found in the LICENSE file.
 
+use std::{collections::BTreeSet, env};
 pub mod api;
 pub mod api_utils;
 pub mod display_utils;
@@ -14,3 +15,33 @@ pub mod server;
 pub mod snapshot;
 pub mod stratagem;
 pub mod update_repo_file;
+
+pub fn parse_hosts(hosts: &[String]) -> Result<BTreeSet<String>, error::ImlManagerCliError> {
+    let parsed: Vec<BTreeSet<String>> = hosts
+        .iter()
+        .map(|x| hostlist_parser::parse(x))
+        .collect::<Result<_, _>>()?;
+
+    let union = parsed
+        .into_iter()
+        .fold(BTreeSet::new(), |acc, h| acc.union(&h).cloned().collect());
+
+    Ok(union)
+}
+
+fn exe_name() -> Option<String> {
+    Some(
+        std::env::current_exe()
+            .ok()?
+            .file_stem()?
+            .to_str()?
+            .to_string(),
+    )
+}
+
+pub fn selfname(suffix: Option<&str>) -> Option<String> {
+    match env::var("CLI_NAME") {
+        Ok(n) => suffix.map(|s| format!("{}-{}", n, s)).or_else(|| Some(n)),
+        Err(_) => exe_name(),
+    }
+}
--- a/iml-manager-cli/src/main.rs
+++ b/iml-manager-cli/src/main.rs
@@ -6,6 +6,7 @@ use iml_manager_cli::{
     api::{self, api_cli, graphql_cli},
     display_utils::display_error,
     filesystem::{self, filesystem_cli},
+    selfname,
     server::{self, server_cli},
     snapshot::{self, snapshot_cli},
     stratagem::{self, stratagem_cli},
@@ -17,8 +18,7 @@ use structopt::clap::Shell;
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
-#[structopt(name = "iml", setting = structopt::clap::AppSettings::ColoredHelp)]
-/// The Integrated Manager for Lustre CLI
+#[structopt(setting = structopt::clap::AppSettings::ColoredHelp)]
 pub enum App {
     #[structopt(name = "stratagem")]
     /// Work with Stratagem server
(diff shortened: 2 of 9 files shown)


iml-manager-cli/src/main.rs
```diff
@@ -16,9 +16,18 @@ use std::process::exit;
...
+            .ok()?
+            .file_stem()?
+            .to_str()?
+            .to_string(),
+    )
+}
+
 #[derive(Debug, StructOpt)]
-#[structopt(name = "iml", setting = structopt::clap::AppSettings::ColoredHelp)]
-/// The Integrated Manager for Lustre CLI
```
ip1981, line 21: Could not come up with a generic non-trivial description :)

iml-manager-cli/src/main.rs
```diff
@@ -16,9 +16,18 @@ use std::process::exit;
...
+#[structopt(name = basename().unwrap_or_else(|| "iml".to_string()), setting = structopt::clap::AppSettings::ColoredHelp)]
```
ip1981, line 30: If needed can use something more sophisticated like env variables, etc.

jgrund:
General question: what happens if the structopt name does not match the binary name?

I.E. if we had a binary called `iml` still but did  something like `#[structopt(name = get_env_name())...` within the code, where `get_env_name()` resolved to `foo`

ip1981:
> General question: what happens if the structopt name does not match the binary name?

I think nothing. The binary name is totally unrelated to the actually running code.
Under the hood structopt just calls some Clap functions. Previously it was using a static string, but actually it can execute any code which deliver a string, for example reading env var or use OS-specific hacks to get the binary's name.

jgrund:
> I think nothing. The binary name is totally unrelated to the actually running code.

So in the case above it would print `foo` for the help text?

ip1981:
> > I think nothing. The binary name is totally unrelated to the actually running code.
> 
> So in the case above it would print `foo` for the help text?

Yes.

jgrund:
> > > I think nothing. The binary name is totally unrelated to the actually running code.
> > 
> > 
> > So in the case above it would print `foo` for the help text?
> 
> Yes.

Ok, so in that case we will want to have an environment variable to name the cli. It should default to `iml` if it's not present and can be used for naming both `iml` and `iml-config` cli (with the iml prefix being the common part).

iml-manager-cli/src/lib.rs
```diff
@@ -14,3 +15,26 @@ pub mod server;
...
+        .collect::<Result<_, _>>()?;
+
+    let union = parsed
+        .into_iter()
+        .fold(BTreeSet::new(), |acc, h| acc.union(&h).cloned().collect());
+
+    Ok(union)
+}
+
+pub fn selfname() -> Option<String> {
```
ip1981, line 32: What about this:
```rust
pub fn selfname(suffix: Option<&str>) -> Option<String> {
// get var
// build with optional suffix (specific to the executable)
// default to executable name
}
```

ip1981:
So it seems impossible without actually renaming the executable:

Screenshot_2020-10-05_22-14-42

jgrund:
These parts look interesting: https://github.com/clap-rs/clap/blob/3e9ee86713b5c407b50ba76f30cffaed25952063/src/output/usage.rs#L48-L53

jgrund:
Can we reach into clap and set `.bin_name(...)`?

ip1981:
> Can we reach into clap and set `.bin_name(...)`?

This will be
```rust
    let matches = App::from_clap(&App::clap().bin_name("fffffffffff").get_matches());
```

mkpankov (dismissed):
> Reviewed 1 of 5 files at r1, 8 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2305)* status: all files reviewed, 3 unresolved discussions (waiting on @ip1981)

___
*[iml-manager-cli/src/config-main.rs, line 28 at r3](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2305#-MIy4mGU5HbygSz_cLlk:-MIy4mGV9DrKU1spxIyQ:bx3fhy0) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/95d693a48bf9ad2e8eeaf9910ae906c847c3aacd/iml-manager-cli/src/config-main.rs#L28)):*
> ```Rust
>     iml_tracing::init();
> 
>     let name = selfname(Some("config")).unwrap_or_else(|| "iml-config".to_string());
> ```

Would it make sense to move the default into `selfname` so that it concatenates `iml-` to whatever comes in `Option`?
___
*[iml-manager-cli/src/lib.rs, line 32 at r2](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2305#-MItRg1uvfVWe4icUY8C-r2-32:-MIy5t4p2SfYWG5TxfPV:btdo28j) ([raw file](https://github.com/whamcloud/integrated-manager-for-lustre/blob/30e218257c96a2e7b0ffe24a06229db25ae5f4da/iml-manager-cli/src/lib.rs#L32)):*

What about this:
```rust
pub fn selfname(suffix: Option<&str>) -> Option<String> {
// get var
// build with optional suffix (specific to the executable)
// default to executable name
}
```
</blockquote></details>

I like

jgrund (dismissed):

jgrund (approved):

PR #2307, opened 2020-10-06
Do not require chronyd to be installed on storage servers

johnsonw:
When installing a minimum build of centos it will not include chronyd in
the installation. This currently prevents users from being able to add
storage servers. Update the disable chronyd step to be smarter such that
the command does not error if it is not installed.

### DIFF
--- a/chroma_core/models/ntp.py
+++ b/chroma_core/models/ntp.py
@@ -19,6 +19,7 @@ from chroma_core.lib.job import DependOn, DependAll, Step
 from chroma_help.help import help_text
 
 import settings
+import sys
 
 
 class NTPConfiguration(DeletableStatefulObject):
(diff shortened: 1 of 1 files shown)


johnsonw:
This show the failure to stop the chronyd service if it is not installed:
image

And this shows that after the patch is applied, the exception is swallowed and the process will continue:
image

chroma_core/models/ntp.py
```diff
@@ -56,14 +57,34 @@ class StopChronyStep(Step):
...
     def run(self, kwargs):
-        return self.invoke_rust_agent_expect_result(kwargs["fqdn"], "stop_unit", "chronyd.service")
+        from chroma_core.services.job_scheduler.agent_rpc import AgentException
+
+        try:
+            return self.invoke_rust_agent_expect_result(kwargs["fqdn"], "stop_unit", "chronyd.service")
+        except AgentException as e:
+            t, v, tb = sys.exc_info()
+
+            if "Unknown busctl" in str(e):
```
ip1981, line 67: Is there a more clear way to say no such unit? 🙂

ip1981 (approved):

chroma_core/models/ntp.py
johnsonw: Yeah, I agree this is frustrating and ugly. If you know of a better way to check please let me know.

chroma_core/models/ntp.py
```diff
@@ -56,14 +57,34 @@ class StopChronyStep(Step):
...
+                return ""
+            else:
+                raise t, v, tb
```
mkpankov, line 70: Why explicitly raise with values? I thought just `raise` would re-raise last exception.

chroma_core/models/ntp.py
jgrund: Instead of trying to stop and handling the error. We should just check the state first and then stop iff it's already installed / running

jgrund:
This patch doesn't address all the places we are assuming chrony is installed. Specifically in the rust ntp daemon plugin.

jgrund (changes requested):
This patch doesn't address all the places we assume chrony is installed. Specifically the rust ntp daemon plugin.

PR #2320, opened 2020-10-09
Disable xattr copy in rsync

beevans:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2320)

### DIFF
--- a/iml-agent/src/action_plugins/stratagem/action_filesync.rs
+++ b/iml-agent/src/action_plugins/stratagem/action_filesync.rs
@@ -26,7 +26,6 @@ fn do_rsync<'a>(
             let output = Command::new("rsync")
                 .arg("-a")
                 .arg("-t")
-                .arg("-X")
                 .arg(&work.src_file)
                 .arg(&work.dest_file)
                 .output()


iml-agent/src/action_plugins/stratagem/action_filesync.rs
```diff
@@ -26,7 +26,6 @@ fn do_rsync<'a>(
             let output = Command::new("rsync")
                 .arg("-a")
                 .arg("-t")
-                .arg("-X")
```
utopiabound, line 29: Don't we WANT xattrs? otherwise we lose striping info, right?

beevans: Right, but if you copy it to NFS the whole thing fails.  We need to create an enhancement for this to duplicate them somehow on the side like other systems do, or only support them for mountpoints that support xattrs.

utopiabound: okay.  We should probably check at some point, but we would want to cache answer locally, so it can wait.

utopiabound (approved):
A ticket should be opened to sync xattr's when possible

jgrund (approved):

PR #2321, opened 2020-10-10
Begin converting the linux plugin over to rust

johnsonw:
Begin converting the linux plugin over to rust

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2321)


iml-agent/src/network.rs
```diff
@@ -0,0 +1,449 @@
...
+            InterfaceProperties::Inet6AddressAndPrefix((address, prefix)) => {
+                self.inet6_address.push(Address { address, prefix });
+            }
+        }
+
+        self
+    }
+}
+
+impl Default for NetworkInterface {
```
jgrund, line 74: Why not just derive this?

iml-agent/src/network.rs
```diff
@@ -0,0 +1,449 @@
...
+            interface_type: None,
+            inet4_address: vec![],
+            inet6_address: vec![],
+            is_up: false,
+            is_slave: false,
+        }
+    }
+}
+
+fn interface_start<I>() -> impl Parser<I, Output = u32>
```
jgrund, line 88: Clippy

iml-agent/src/network.rs
```diff
@@ -0,0 +1,449 @@
...
+    I: Stream<Token = char>,
+    I::Error: ParseError<I::Token, I::Range, I::Position>,
+{
+    spaces()
+        .and(many1::<Vec<_>, _, _>(digit()))
+        .map(|x| x.1.into_iter().collect::<String>().parse::<u32>().unwrap())
+        .skip(char(':'))
+}
+
+fn interface<I>() -> impl Parser<I, Output = String>
```
jgrund, line 99: Clippy

iml-agent/src/network.rs
```diff
@@ -0,0 +1,449 @@
...
+#[derive(Debug, PartialEq, Eq)]
+pub enum InterfaceProperties {
+    InterfaceFlagsAndAttributes((String, Vec<String>, HashMap<String, String>)),
+    InterfaceTypeAndMacAddress((Option<String>, Option<String>)),
+    Inet4AddressAndPrefix((String, u32)),
+    Inet6AddressAndPrefix((String, u32)),
+}
+
+#[derive(Debug)]
+pub struct Address {
```
jgrund, line 28: You should probably use https://docs.rs/ipnetwork/ here instead.

johnsonw: Makes sense to me. I'll update.

johnsonw: Updated.

iml-agent/src/network.rs
johnsonw: By derive, are you suggesting that I get rid of the default implementation and just create the struct since I'm only using it in one place?

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,459 @@
...
+    Inet6AddressAndPrefix((String, u32)),
+}
+
+#[derive(Debug)]
+pub struct Address {
+    address: String,
+    prefix: u32,
+}
+
+#[derive(Debug)]
```
mkpankov, line 34:
```suggestion
#[derive(Debug, Default)]
```

iml-agent/src/network.rs
jgrund: No, I'm suggesting you use `#[derive(Default)]` instead of writing it by hand.

mkpankov (commented):
Looks like there are some test failures

iml-agent/src/network_interface.rs
johnsonw: Ah, right

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,454 @@
...
+                if flags.iter().any(|x| x == &"SLAVE".to_string()) {
+                    self.is_slave = true;
+                }
+            }
+            InterfaceProperties::InterfaceTypeAndMacAddress((interface_type, address)) => {
+                self.interface_type = interface_type;
+                self.mac_address = address;
+            }
+            InterfaceProperties::Inet4AddressAndPrefix((address, prefix)) => {
+                let address: Ipv4Addr = address.parse().expect("parse address to Ipv4Addr");
```
jgrund, line 63: Please remove expect here and propagate Errors instead.

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,454 @@
...
+                self.inet4_address
+                    .push(Ipv4Network::new(address, prefix).expect("create Ipv4Network"));
```
jgrund, line 65: Please remove expect here and propagate Errors instead.

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,454 @@
...
+            }
+            InterfaceProperties::Inet6AddressAndPrefix((address, prefix)) => {
+                let address: Ipv6Addr = address.parse().expect("parse address to Ipv6Addr");
```
jgrund, line 68: Please remove expect here and propagate Errors instead.

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,454 @@
...
+                self.inet6_address
+                    .push(Ipv6Network::new(address, prefix).expect("create Ipv6Network"));
```
jgrund, line 70: Please remove expect here and propagate Errors instead.

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,454 @@
...
+}
+
+fn interface_start<I>() -> impl Parser<I, Output = u32>
+where
+    I: Stream<Token = char>,
+    I::Error: ParseError<I::Token, I::Range, I::Position>,
+{
+    spaces()
+        .and(many1::<Vec<_>, _, _>(digit()))
+        .map(|x| x.1.into_iter().collect::<String>().parse::<u32>().unwrap())
```
jgrund, line 85: Can we propagate the error here instead of unwrapping?

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,454 @@
...
+        .map(|x| x.1)
+}
+
+fn parse_inet4_prefix<I>() -> impl Parser<I, Output = u8>
+where
+    I: Stream<Token = char>,
+    I::Error: ParseError<I::Token, I::Range, I::Position>,
+{
+    char('/')
+        .and(many1(digit()).map(|x: String| x.parse::<u8>().unwrap()))
```
jgrund, line 149: Can we propagate the error here instead of unwrapping?

iml-agent/src/network_interface.rs
```diff
@@ -0,0 +1,454 @@
...
+        .map(|x| x.1)
+}
+
+fn parse_inet6_prefix<I>() -> impl Parser<I, Output = u8>
+where
+    I: Stream<Token = char>,
+    I::Error: ParseError<I::Token, I::Range, I::Position>,
+{
+    char('/')
+        .and(many1(digit()).map(|x: String| x.parse::<u8>().unwrap()))
```
jgrund, line 171: Can we propagate the error here instead of unwrapping?

iml-agent/src/network_interface_stats.rs
```diff
@@ -0,0 +1,207 @@
...
+        .skip(char(':'))
+}
+
+fn parse_stat<I>() -> impl Parser<I, Output = u64>
+where
+    I: Stream<Token = char>,
+    I::Error: ParseError<I::Token, I::Range, I::Position>,
+{
+    spaces()
+        .and(many1(digit()).map(|x: String| x.parse::<u64>().unwrap()))
```
jgrund, line 97: Can we propagate the error here instead of unwrapping?

iml-agent/src/network.rs
johnsonw: I think this is outdated.

iml-agent/src/network_interface_stats.rs
```diff
@@ -5,7 +5,9 @@ use combine::{
     stream::Stream,
     token, Parser,
 };
-use std::collections::HashMap;
+use std::{collections::HashMap, convert::TryFrom, num::ParseIntError};
+
+type MaybeStat = Result<u64, ParseIntError>;
```
jgrund, line 10: nit: `MaybeStat` makes me think of an option. `StatResult` would probably be a good name.

jgrund (dismissed):

iml-agent/src/network_interface_stats.rs
johnsonw: Sure, I'll update.

iml-agent/src/agent_error.rs
```diff
@@ -71,11 +71,28 @@ impl std::error::Error for CibError {
     }
 }
 
+#[derive(Debug)]
+pub struct InterfaceError(pub String);
```
jgrund, line 75: Why do we need a new Error struct here when the contents is just a String?

johnsonw: I thought it looked a little cleaner but I can just make it a string if you prefer.

jgrund: I don't think we need the extra level of indirection since it's all in the same crate.

iml-wire-types/src/lib.rs
```diff
@@ -2079,3 +2081,217 @@ impl PartialEq for LdevEntry {
...
+pub enum LndType {
+    Tcp,
+    O2ib,
+}
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum LinuxType {
+    Ethernet,
+    Ether,
```
jgrund, line 2096: What is the difference between `Ethernet` and `Ether`?

johnsonw: Both types were included in the original source and both are mapped to "tcp".

jgrund: Why do we need to have both of them?

johnsonw: I believe it depends on the output. I don't think we are guaranteed that it will always be `ether` or always be `ethernet`.

johnsonw: From my experience, however, i've only seen `ether`.

jgrund: Don't we always convert both to tcp in the output though?

johnsonw: Yes, both are converted to "tcp" for the `network_interface` table.

johnsonw: I can take the `Ethernet` variant out but I don't think it hurts to have it there to cover that case in the event that it is used in the output.

jgrund: I think it could be confusing to have two variants that mean the same thing.

jgrund (commented):
nit: network_interface.rs, network_interface_stats.rs, and network_interfaces.rs  could all have been in the same file.

mkpankov (approved):

PR #2325, opened 2020-10-12
EX-1932 handle softlinks for llapi queries

beevans:
When constructing an llapi instance from a random path
liblustreapi first resolves that path fully, including
symbolic links.  This causes a failure when trying to
get an llapi instance from a symbolic link that lives on
a lustre filesystem (ex /mnt/lustre/symbolic_link)

If the path passed into search_fsname is a link, use
the parent dir instead.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2325)

### DIFF
--- a/liblustreapi/src/lib.rs
+++ b/liblustreapi/src/lib.rs
@@ -312,7 +312,7 @@ impl Llapi {
             path.pop();
         }
 
-        let fsc = CString::new(path.clone().into_os_string().as_bytes())?;
+        let fsc = CString::new(path.into_os_string().as_bytes())?;
         let mut fsname: Vec<u8> = vec![0; std::mem::size_of::<u8>() * MAXFSNAME + 1];
         let ptr = fsname.as_mut_ptr() as *mut libc::c_char;
 
--- a/liblustreapi/src/lib.rs
+++ b/liblustreapi/src/lib.rs
@@ -305,7 +305,19 @@ impl Llapi {
     }
 
     pub fn search_fsname(&self, pathname: &str) -> Result<String, LiblustreError> {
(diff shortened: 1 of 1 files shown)


liblustreapi/src/lib.rs
```diff
@@ -305,7 +305,19 @@ impl Llapi {
...
+            path.pop();
+        }
+
+        let fsc = CString::new(
+            path.clone()
+                .into_os_string()
+                .into_string()
+                .unwrap()
+                .as_bytes(),
+        )?;
```
jgrund, line 320:
```suggestion
        let fsc = CString::new(path.into_os_string().as_bytes())?;
```

(Import `std::os::unix::ffi::OsStrExt` at the top of the file)

beevans: Done

liblustreapi/src/lib.rs
```diff
@@ -305,7 +306,13 @@ impl Llapi {
...
 
     pub fn search_fsname(&self, pathname: &str) -> Result<String, LiblustreError> {
-        let fsc = CString::new(pathname.as_bytes())?;
+        let mut path = PathBuf::from(pathname);
+        let md = std::fs::symlink_metadata(&path)?;
+        if md.file_type().is_symlink() {
+            path.pop();
+        }
+
+        let fsc = CString::new(path.clone().into_os_string().as_bytes())?;
```
jgrund, line 315: Why do we need to clone here?

beevans: Turns out we don't!

Done

PR #2326, opened 2020-10-12
filesync should only return fids that errored

beevans:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2326)

### DIFF
--- a/iml-agent/src/action_plugins/stratagem/action_filesync.rs
+++ b/iml-agent/src/action_plugins/stratagem/action_filesync.rs
@@ -40,8 +40,9 @@ fn do_rsync<'a>(
         .chunks(10)
         .map(Ok)
         .try_fold(vec![], |mut acc, xs| async {
-            let xs = try_join_all(xs).await?;
+            let mut xs = try_join_all(xs).await?;
 
+            xs.retain(|x| x.errno != 0);
             acc.extend(xs);
 
             Ok::<_, ImlAgentError>(acc)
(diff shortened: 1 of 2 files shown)


iml-agent/src/action_plugins/stratagem/action_filesync.rs
```diff
@@ -40,8 +40,9 @@ fn do_rsync<'a>(
         .chunks(10)
         .map(Ok)
         .try_fold(vec![], |mut acc, xs| async {
-            let xs = try_join_all(xs).await?;
+            let mut xs = try_join_all(xs).await?;
 
+            xs.retain(|x| x.errno != 0);
             acc.extend(xs);
```
utopiabound, line 46: @jgrund 
Would this be better as `acc.extend(xs.filter(|x| x.errno != 0))`?

beevans: I think this is simpler

jgrund: I'd probably do something like:

```rust
acc.into_iter().chain(xs.into_iter().filter(|x| x.errno != 0)).collect()
```

Which would avoid both `acc` and `xs` from being `mut`.

That said, meh.

jgrund (approved):

jgrund (approved):

PR #2328, opened 2020-10-13
Convert stratagem scanning over to graphql

jgrund:
Running a scan currently consists of at least three different areas on
the manager:

1. The python API which performs validation and resource -> db id
resolution
2. The job_scheduler, which assigns tasks and builds out the command
plan.
3. The RunStratagem step, which will build the actual stratagem config and ship it
off to each MDT.

Each area only owns part of the whole solution (creating /
destroying tasks and shipping a config off to an MDT). It is overly
complex to split responsibility in this way, and also brittle for future
changes.

Instead, we can add a single mutation that takes in the fsname to scan + tasks and rule groups and does all the work necessary to
generate the command. 

This cuts the job_scheduler out of the picture and
the step to run the config becomes very simple as it no longer is
responsible for building the config.

- Create a graphql mutation to run stratagem scans given a config and
some tasks.
- Extract stratagem api code to a separate module.
- Implement a simple step to scan a MDT.
- Refactor GUI to remove reliance on certain profiles `USE_STRATAGEM=true` is now the only requirement.
- Refactor CLI to use graphql
- Refactor GUI to use graphql
- Refactor integration testing to share more of the path

Screen Shot 2020-10-14 at 9 31 40 AM

This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2328)


iml-manager-cli/src/stratagem.rs
```diff
@@ -98,12 +98,14 @@ pub struct StratagemRemoveData {
...
-    /// The report duration
-    #[structopt(short = "r", long = "report", parse(try_from_str = parse_duration))]
-    report_duration: Option<u64>,
-    /// The purge duration
-    #[structopt(short = "p", long = "purge", parse(try_from_str = parse_duration))]
-    purge_duration: Option<u64>,
+    /// The report duration, specified as a humantime string
+    /// EX: 1hour
+    #[structopt(short = "r", long = "report", min_values = 1)]
+    report_duration: Option<Vec<String>>,
```
ip1981, line 104: Does it need to be optional with min_values = 1?

iml-manager-cli/src/stratagem.rs
```diff
@@ -125,38 +127,28 @@ arg_enum! {
 
 #[derive(serde::Serialize, StructOpt, Debug)]
 pub struct StratagemFilesyncData {
-    /// action, 'push' is supported
-    #[structopt()]
-    action: FilesyncAction,
     /// The name of the filesystem to scan
-    #[structopt(short = "f", long = "filesystem")]
```
ip1981, line 132: :+1:

iml-manager-cli/src/stratagem.rs
jgrund: probably semantics, but no report duration has a different meaning than the flag with nothing there.

iml-graphql-queries/src/stratagem.rs
```diff
@@ -27,18 +34,22 @@ pub mod list_reports {
     }
 
     #[derive(Debug, Clone, serde::Deserialize)]
-    pub struct Resp {
+    pub struct StratagemReports {
         #[serde(rename(deserialize = "stratagemReports"))]
         pub stratagem_reports: Vec<StratagemReport>,
     }
+
+    pub type Resp = super::Resp<StratagemReports>;
```
ip1981, line 42: :+1:

ip1981 (approved):

chroma_core/models/stratagem.py
```diff
@@ -317,153 +318,33 @@ def get_deps(self):
...
-        def calc_warn_duration(report_duration, purge_duration):
-            if report_duration is not None and purge_duration is not None:
-                return "&& != type S_IFDIR && < atime - sys_time {} > atime - sys_time {}".format(
-                    report_duration, purge_duration
-                )
+        pass
 
-            return "&& != type S_IFDIR < atime - sys_time {}".format(report_duration or 0)
-
-        def get_body(mount_point, action, report_duration, purge_duration, search_expression):
```
beevans, line 336: I may be missing it, but I'm not seeing this getting generated anywhere by this patch.  Specifically LAT_COUNTER_INC (etc.) are not being generated by the serialization code in iml-wire-types (where I assume the conf file is getting generated)

jgrund: That is being done here (the file is large so github hid it's contents by default): 

https://github.com/whamcloud/integrated-manager-for-lustre/pull/2328/files/81ded592c199c59bd0c92f0a974bd65e47cbc2dd#diff-f138cff3f473efc07fd27907a450b2c0e5583cd133dcecec863e2a7d1d881d51R350-R378

jgrund:
Current available mutations:

Screen Shot 2020-10-21 at 10 01 57 AM

iml-api/src/graphql/stratagem.rs
```diff
@@ -0,0 +1,680 @@
...
+            Some(kwargs),
+        )
+        .map_err(ImlApiError::ImlJobSchedulerRpcError)
+        .await?;
+
+        let command = get_command(&context.pg_pool, command_id).await?;
+
+        Ok(command)
+    }
+    async fn run_fast_file_scan(
```
beevans, line 253: Rename: run_report_and_purge

iml-api/src/graphql/stratagem.rs
```diff
@@ -0,0 +1,680 @@
...
+
+            let mut cfg = stratagem::StratagemConfig {
+                flist_type: "none".into(),
+                summarize_size: true,
+                device: stratagem::StratagemDevice {
+                    path,
+                    groups: groups.clone(),
+                },
+                groups: vec![
+                    stratagem::StratagemGroup {
```
beevans, line 351: In the old version, generating the size distribution happened every time we did a scan.  Now it's only when report/purge are called.

iml-api/src/graphql/stratagem.rs
jgrund: `fast_file_scan` is the blessed name for this feature (I agree report_and_purge has more meaning).

iml-api/src/graphql/stratagem.rs
jgrund: Yes, and I think this is the correct behavior.

The UI for size_distribution and inode users (the only place where this data is visible) is currently tightly coupled to running this particular scan with these particular rules. I don't see a benefit to duplicating these groups in the other mutations as well.

What we should do in a future patch is allow custom stats ingestion and reporting; I think the first step of that is to limit hardcoding groups like these to only the scans where they are truly needed.

beevans (approved):

iml-system-test-utils/src/lib.rs
```diff
@@ -732,20 +711,27 @@ pub async fn test_stratagem_taskqueue(config: Config) -> Result<Config, TestErro
...
+        },
+    );
 
-    let ssh_stdin = ssh_child.stdin.as_mut().unwrap();
-    ssh_stdin.write_all(task).await?;
-    ssh_child.wait_with_checked_output().await?;
+    ssh::graphql_call::<_, iml_graphql_queries::Response<task::create::Resp>>(&config, &q).await?;
 
-    // send fid
+    // TODO wait for command to complete
```
johnsonw, line 731: I thought we had a function to wait for a command to complete. Will that not work here?

johnsonw (approved):

iml-system-test-utils/src/lib.rs
jgrund: Not without some work. It currently uses http api calls internally and the current approach here is using SSH.

That said, we should probably just do our calls over HTTP anyway, we already are exposing the port in vagrant.

PR #2335, opened 2020-10-14
Remove swap-emitter.timer

### LINKED ISSUE #2232: Start(ed|ing) IML Swap Emitter...
Seeing tons of Started/Starting IML Swap Emitter messages in the GUI logs.

Maybe only start it once?  Maybe not tell me when it succeeds, only fails?

Maybe filter the GUI logs for Warning/Error messages by default?

jgrund:
Due to systemd/systemd#9328, there is no way to silence a systemd timer.

To workaround this limitation, create a new daemon that will list swap
devices one per minute, but without logging each tick to the journal.

In addition, update mount-emitter to use latest futures and tokio.

Fixes #2232

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2335)


device-scanner/mount-emitter/src/lib.rs
```diff
@@ -8,67 +8,94 @@
...
-        FsType(x.remove("FSTYPE").expect("missing FSTYPE")),
-        MountOpts(x.remove("OPTIONS").expect("missing OPTIONS")),
+        MountPoint(
+            x.remove("TARGET")
+                .ok_or_else(|| Error::Missing("TARGET"))?
+                .into(),
+        ),
+        DevicePath(
+            x.remove("SOURCE")
+                .ok_or_else(|| Error::Missing("SOURCE"))?
```
ip1981, line 47: Doesn't need to be `_else`?

ip1981 (approved):

device-scanner/mount-emitter/src/lib.rs
jgrund: Wasn't sure if clippy would complain about the allocation

PR #2337, opened 2020-10-15
Do stats accounting for tasks

utopiabound:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2337)

### DIFF
--- a/iml-task-runner/src/stats.rs
+++ b/iml-task-runner/src/stats.rs
@@ -1,0 +1,89 @@
+// Copyright (c) 2020 DDN. All rights reserved.
+// Use of this source code is governed by a MIT-style
+// license that can be found in the LICENSE file.
+
+use crate::error::ImlTaskRunnerError;
+use bigdecimal::{BigDecimal, ToPrimitive};
+use iml_influx::{Client, Point, Points, Precision, Value};
+use iml_manager_client::Url;
+use iml_manager_env::{get_influxdb_addr, get_influxdb_metrics_db};
+use iml_postgres::sqlx::{self, PgPool};
+use iml_tracing::tracing;
+use std::time::Duration;
+use tokio::time;
+
+struct TaskStats {
+    actions: Vec<String>,
+    fs_name: String,
+    fids_total: BigDecimal,
+    fids_completed: BigDecimal,
+    fids_failed: BigDecimal,
+}
+
+const DELAY: Duration = Duration::from_secs(60);
+
+pub(crate) async fn collector(pool: PgPool) -> Result<(), ImlTaskRunnerError> {
+    let mut interval = time::interval(DELAY);
+
+    let influx_url: String = format!("http://{}", get_influxdb_addr());
+    tracing::debug!("influx_url: {}", &influx_url);
+
+    loop {
+        interval.tick().await;
+
+        let stats: Vec<TaskStats> = sqlx::query_as!(
+            TaskStats,
+            r#"SELECT COALESCE(SUM(fids_total), 0) AS "fids_total!",
+COALESCE(SUM(fids_completed), 0) AS "fids_completed!",
+COALESCE(SUM(fids_failed), 0) AS "fids_failed!",
+actions,
+fs_name FROM chroma_core_task GROUP BY actions,fs_name"#
+        )
+        .fetch_all(&pool)
+        .await?;
+
+        if stats.is_empty() {
+            continue;
+        }
+
+        let client = Client::new(
+            Url::parse(&influx_url).expect("Influx URL is invalid."),
+            get_influxdb_metrics_db(),
+        );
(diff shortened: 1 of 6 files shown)


iml-task-runner/src/stats.rs
```diff
@@ -0,0 +1,108 @@
...
+            get_influxdb_metrics_db(),
+        );
+
+        let xs = stats
+            .iter()
+            .filter_map(|stat| {
+                if let Some(action) = stat.actions.first() {
+                    Some(
+                        Point::new("task")
+                            .add_tag("action", Value::String(action.clone()))
```
jgrund, line 61: Shouldn't need to clone

iml-task-runner/src/stats.rs
```diff
@@ -0,0 +1,108 @@
...
+                            .add_tag("filesystem", Value::String(stat.filesystem.clone()))
```
jgrund, line 62: Shouldn't need to clone

iml-task-runner/src/stats.rs
```diff
@@ -0,0 +1,108 @@
...
+
+struct TaskStats {
+    actions: Vec<String>,
+    filesystem: String,
+    fids_total: Option<BigDecimal>,
+    fids_completed: Option<BigDecimal>,
+    fids_failed: Option<BigDecimal>,
+}
+
+const DELAY: Duration = Duration::from_secs(300);
```
jgrund, line 23: I think 5 minutes is too long to get a good reading over time

iml-task-runner/src/stats.rs
```diff
@@ -0,0 +1,108 @@
...
+SUM(fids_completed) AS fids_completed,
+SUM(fids_failed) AS fids_failed,
+actions,
+fs.name AS filesystem FROM chroma_core_task AS t
+JOIN chroma_core_managedfilesystem AS fs ON t.filesystem_id = fs.id GROUP BY actions,fs.name"#
+        )
+        .fetch_all(&pool)
+        .await?;
+
+        if stats.len() == 0 {
```
jgrund, line 46: clippy warning

iml-task-runner/src/stats.rs
```diff
@@ -0,0 +1,108 @@
...
+        interval.tick().await;
+
+        let stats: Vec<TaskStats> = sqlx::query_as!(
+            TaskStats,
+            r#"SELECT SUM(fids_total) AS fids_total,
+SUM(fids_completed) AS fids_completed,
+SUM(fids_failed) AS fids_failed,
+actions,
+fs.name AS filesystem FROM chroma_core_task AS t
+JOIN chroma_core_managedfilesystem AS fs ON t.filesystem_id = fs.id GROUP BY actions,fs.name"#
```
jgrund, line 41: I'm a bit concerned about using the fs id instead of the name, can we update the task table to use the fsname instead of the id (in a separate patch)?

jgrund, line 41:
```suggestion
            r#"SELECT COALESCE(SUM(fids_total), 0) AS "fids_total!",
COALESCE(SUM(fids_completed)) AS "fids_completed!",
COALESCE(SUM(fids_failed)) AS "fids_failed!",
actions,
fs.name AS filesystem FROM chroma_core_task AS t
JOIN chroma_core_managedfilesystem AS fs ON t.filesystem_id = fs.id GROUP BY actions,fs.name"#
```

This forces null `SUM` fields to be 0.  `TaskStats` can be updated to remove all the sums.

jgrund: https://github.com/whamcloud/integrated-manager-for-lustre/issues/2358

jgrund (changes requested):
The query can be updated, and the timer dropped down a bit to get a better reading on scans.

In addition, I've opened #2358 because I don't think we should rely on fs id in the task table.

iml-task-runner/src/stats.rs
utopiabound:
```
error[E0507]: cannot move out of `stat.filesystem` which is behind a shared reference
  --> iml-task-runner/src/stats.rs:62:66
   |
62 | ...                   .add_tag("filesystem", Value::String(stat.filesystem))
   |                                                            ^^^^^^^^^^^^^^^ move occurs because `stat.filesystem` has type `std::string::String`, which does not implement the `Copy` trait
```

iml-task-runner/src/stats.rs
utopiabound: SQLX didn't seem to care about the COALESCE when I tried that.

jgrund: It's not the `COALESCE` part, it's the cast I added afterwards that allows this to compile.

PR #2341, opened 2020-10-21
Reconfigure zfs-zed before start zed-enhancer

### LINKED ISSUE #2201: zed-enhancer:  send hup signal to zfs-zed
`device-scanner` installs a number of zedlets to get near-realime changes from `zed`. However, these zedlets won't be picked up by default by `zed` until a restart. 

We should make sure to send a `HUP` to `zed` before starting `zed-enhancer` so that the zedlets are picked up and processed as expected.

ip1981:
To pick up installed ZEDLETs

Closes #2201.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2341)

### DIFF
--- a/device-scanner/zed-enhancer/systemd-units/zed-enhancer.service
+++ b/device-scanner/zed-enhancer/systemd-units/zed-enhancer.service
@@ -9,6 +9,8 @@ After=device-scanner.socket
 [Service]
 Restart=on-failure
 Environment=RUST_LOG=info
+# Reconfigure the daemon and rescan the directory for enabled ZEDLETs:
+ExecStartPre=/bin/systemctl kill --signal HUP zfs-zed.service
 ExecStart=/usr/bin/zed-enhancer
 StandardOutput=journal
 StandardError=journal


device-scanner/zed-enhancer/systemd-units/zed-enhancer.service
```diff
@@ -9,6 +9,8 @@ After=device-scanner.socket
 [Service]
 Restart=on-failure
 Environment=RUST_LOG=info
+# Reconfigure the daemon and rescan the directory for enabled ZEDLETs:
+ExecStartPre=/bin/bash -c '! systemctl --quiet is-active zfs-zed.service || systemctl kill --signal HUP zfs-zed.service'
```
jgrund, line 13: Hrrm, if this was not active, wouldn't we still have the same issue we do now?

ip1981: It's active if and only if it's running. If it's not running, it will read zedlets on start. It could also be uninstalled at all, if zfs is not used.

So this is simple: hup if running, otherwise do nothing.

jgrund: Well, it *must* be running because we have a `Requires` and `After` on it in this unit.

I wonder if we gain anything by having this check here... Is it ever possible to have a race?

ip1981: I don't think race is possible.
A typical case would be upgrading zed-enhancer which would need to kick zed itself.

Worse case: zed will reread zedlets shortly after start.

I guess just sending the signal will do, I was mistakenly assuming that zed is optional.

jgrund (approved):

mkpankov (approved):

PR #2344, opened 2020-10-23
Postoffice: Use atomic conf file write

utopiabound:
Write a new file and rename over old.
The rename is atomic.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2344)


jgrund:
I'm not quite sure what problem this is solving, but why not use an actual mutex singleton? That would ensure a single writer in an idiomatic way.

iml-agent/src/action_plugins/postoffice.rs
```diff
@@ -2,43 +2,118 @@
...
         .lines()
         .map(|s| s.to_string())
         .collect())
 }
 
-pub async fn route_add(mailbox: String) -> Result<(), ImlAgentError> {
+async fn locked_route(
+    file: fs::File,
+    mailbox: String,
+    insert: bool,
```
mkpankov:
```
enum Operation {
    Insert,
    Remove
}
```
is probably easier to understand

mkpankov (dismissed):

johnsonw:
@jgrund, would using a combination of lazy static with a mutex work?

mkpankov (dismissed):

utopiabound:
The provides Reader/Writer protection without a mutex, and without worrying about ensuring all threads have access to a shared mutex.  A reader will ALWAYS get a well formatted complete list regardless of when they read, and will never have to block on a writer.  A writer will never leave a file half updated (since the tmp file is deleted at startup).

johnsonw (dismissed):

PR #2347, opened 2020-10-27
Return all resources from iml-agent::high-availability::get_crm_mon

utopiabound:
Return all resources in Cluster.resources

Blocks #2178 

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2347)


jgrund:
Put this on a cluster:

Screen Shot 2020-10-29 at 9 56 08 AM


I guess we can just rename the table and expose a new query and we should be good to go I think.

jgrund (approved):

PR #2356, opened 2020-10-30
Update action-client

jgrund:
Update action-client once again to optionally take a uuid that can be
used to bookkeep the request in flight.

In addition, remove `invoke_rust_agent_cancelable` as it adds an extra
layer of indirection that is more useful in tools calling the client.

Replace it with a `cancel_request` fn that will take a `Uuid` ref and
attempt to cancel a remote rpc matching the uuid.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2356)

### DIFF
--- a/iml-action-client/src/lib.rs
+++ b/iml-action-client/src/lib.rs
@@ -3,14 +3,10 @@
 // license that can be found in the LICENSE file.
 
 use bytes::buf::BufExt as _;
-use futures::{
-    future::{abortable, AbortHandle, Aborted},
-    Future,
-};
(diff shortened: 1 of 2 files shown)


johnsonw (commented):
Looks like there is a clippy error in main.rs

jgrund:
> Looks like there is a clippy error in main.rs

Yeah, but not from this patch. I actually think there is a clippy issue inside that macro:

Screen Shot 2020-10-30 at 3 18 40 PM

johnsonw (approved):

PR #2361, opened 2020-11-03
Job and steps migrations

nlinker:
Complementary changes in `iml-api` are in the PR https://github.com/whamcloud/integrated-manager-for-lustre/pull/2404 (after the current PR)

Signed-off-by: Nick Linker <nlinker@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2361)


nlinker:
Seems to work
image

chroma_core/models/step_result.py
```diff
@@ -47,6 +52,14 @@ class StepResult(models.Model):
 
     _step_types = {}
 
+    # Assign the dependent fields, we cannot assign these fields in the constructor,
+    # therefore we intercept the save() method
+    def save(self, *args, **kwargs):
```
utopiabound, line 57: should these check to see if they've already been set before overwriting?

chroma_core/models/jobs.py
```diff
@@ -350,6 +352,9 @@ def __hash__(self):
 
     wait_for_json = models.TextField()
     locks_json = models.TextField()
+    class_name = models.TextField(default="")
+    description_out = models.TextField(default="")
+    cancellable_out = models.BooleanField(default=True)
```
utopiabound, line 357: Why can't we just convert cancellable to a model field instead of adding a copied field?  Does that not work somehow?

nlinker: There are original method `description` and field `cancellable` in the `Job` (and all the descendants). I've made sure not to interfere with them.

chroma_core/models/step_result.py
nlinker: Could you reveal more the undesired situation that might happen?

utopiabound (approved):

jgrund (changes requested):
This PR needs to be moved.

nlinker:
Moving to emf

PR #2373, opened 2020-11-10
Fix Stale Influx Entries

johnsonw:
There is currently a case in which a stale entries are being left in the
influx db. This takes place when a filesystem is mounted, data is pulled
into influx, and then another filesystem is mounted, and both
filesystems use the same mgs. For example:

time n:
mgs_fs: "fs"

time n+1:
mgs_fs: "fs2"

time n+2:
mgs_fs: "fs,fs2"

Currently, only items matching "fs,fs2" will be removed, since that's
the current input. This results in entries in the target table showing
that the filesystems are: "fs, fs2, fs,fs2".

This patch ensures that all subsets of the input filesystems are deleted
and all but the last of the current subset is deleted.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2373)

### DIFF
--- a/iml-services/iml-stats/src/main.rs
+++ b/iml-services/iml-stats/src/main.rs
@@ -547,9 +547,9 @@ async fn delete_existing_mgs_fs_records(
         .query_into(
             format!(
                 r#"
-                SELECT mgs_fs, mgs_fs_count FROM 
+                SELECT mgs_fs, mgs_fs_count FROM
                     (SELECT count(is_mgs_fs) AS mgs_fs_count FROM target WHERE is_mgs_fs=true GROUP BY mgs_fs),
-                    (SELECT LAST(is_mgs_fs),time as last_time FROM target GROUP BY mgs_fs) 
+                    (SELECT LAST(is_mgs_fs),time as last_time FROM target GROUP BY mgs_fs)
                 GROUP BY mgs_fs
                 "#
             )
@@ -560,7 +560,7 @@ async fn delete_existing_mgs_fs_records(
         .unwrap_or_default();
 
     tracing::debug!(
-        "delete existing mgs_fs records - fs_names: {:?}; query results: {:?}",
+        "delete stale mgs_fs records - fs_names: {:?}; query results: {:?}",
         fs_names,
         xs
     );
@@ -570,15 +570,12 @@ async fn delete_existing_mgs_fs_records(
         .filter(|x| {
             let mgs_fs_set = fs_names_to_hash_set(&x.mgs_fs);
 
-            let intersection = mgs_fs_set
-                .intersection(&fs_names)
-                .map(|x| x.to_string())
-                .collect::<Vec<String>>();
+            let intersection = mgs_fs_set.intersection(&fs_names).collect::<Vec<_>>();
 
-            return !intersection.is_empty();
+            !intersection.is_empty()
         })
         .fold(HashMap::new(), |mut acc, record| {
-            let mgs_fs = record.mgs_fs.to_string();
+            let mgs_fs = record.mgs_fs;
             let mut entry = acc.entry(mgs_fs).or_insert((0, 0));
 
             if record.time > 0 {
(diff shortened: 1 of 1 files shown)


iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,93 @@ async fn main() -> Result<(), ImlStatsError> {
...
-            SELECT mgs_fs,is_mgs_fs
-            FROM target
-            WHERE mgs_fs='{}' AND is_mgs_fs=true
-            ORDER BY time ASC
-            LIMIT 2"#,
-                fs_names
+                SELECT mgs_fs, mgs_fs_count FROM 
+                    (SELECT count(is_mgs_fs) AS mgs_fs_count FROM target WHERE is_mgs_fs=true GROUP BY mgs_fs),
+                    (SELECT LAST(is_mgs_fs),time as last_time FROM target GROUP BY mgs_fs) 
+                GROUP BY mgs_fs
```
jgrund, line 553:
```suggestion
                SELECT mgs_fs, mgs_fs_count FROM
                    (SELECT count(is_mgs_fs) AS mgs_fs_count FROM target WHERE is_mgs_fs=true GROUP BY mgs_fs),
                    (SELECT LAST(is_mgs_fs),time as last_time FROM target GROUP BY mgs_fs)
                GROUP BY mgs_fs
```

iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,93 @@ async fn main() -> Result<(), ImlStatsError> {
...
+            Some(Precision::Nanoseconds)
         )
         .await?
         .unwrap_or_default();
 
-    if xs.len() < 2 {
-        return Ok(());
-    }
+    tracing::debug!(
+        "delete existing mgs_fs records - fs_names: {:?}; query results: {:?}",
```
jgrund, line 563:
```suggestion
        "delete stale mgs_fs records - fs_names: {:?}; query results: {:?}",
```

iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,93 @@ async fn main() -> Result<(), ImlStatsError> {
...
+        .into_iter()
+        .filter(|x| {
+            let mgs_fs_set = fs_names_to_hash_set(&x.mgs_fs);
+
+            let intersection = mgs_fs_set
+                .intersection(&fs_names)
+                .map(|x| x.to_string())
+                .collect::<Vec<String>>();
+
+            return !intersection.is_empty();
```
jgrund, line 578:
```suggestion
            let mgs_fs_set = fs_names_to_hash_set(&x.mgs_fs);

            let intersection = mgs_fs_set.intersection(&fs_names).collect::<Vec<_>>();

            !intersection.is_empty()
```

Don't need to convert the borrow into an owned String.

iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,93 @@ async fn main() -> Result<(), ImlStatsError> {
...
+        })
+        .fold(HashMap::new(), |mut acc, record| {
+            let mgs_fs = record.mgs_fs.to_string();
```
jgrund, line 581:
```suggestion
            let mgs_fs = record.mgs_fs;
```

This is already owned

iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,93 @@ async fn main() -> Result<(), ImlStatsError> {
...
+            let mut entry = acc.entry(mgs_fs).or_insert((0, 0));
+
+            if record.time > 0 {
```
jgrund, line 584: In what case will this be <= 0?

iml-services/iml-stats/src/main.rs
```diff
@@ -592,3 +635,11 @@ fn join_fs_names(xs: Vec<FsName>) -> String {
         .collect::<Vec<String>>()
         .join(",")
 }
+
+fn fs_names_vec_to_hash_set(xs: &Vec<FsName>) -> HashSet<String> {
```
jgrund, line 639:
```suggestion
fn fs_names_vec_to_hash_set(xs: &[FsName]) -> HashSet<String> {
```

iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,93 @@ async fn main() -> Result<(), ImlStatsError> {
...
+            }
+
+            acc
+        });
+
+    let delete_subset_futures = subsets.into_iter().map(|(mgs_fs, (time, count))| {
+        let mds_fs_set = fs_names_to_hash_set(&mgs_fs);
 
-        tracing::debug!("Running query: {}", query);
+        let query = if mds_fs_set == fs_names && count > 1 {
```
jgrund, line 598: I think this is too general. What happens in the case where the filesystem names match and the count is == 1?

In that case, it looks like the record will be deleted right? I think you want a filter step before this that removes any records where `mds_fs_set == fs_names && count <= 1`

iml-services/iml-stats/src/main.rs
johnsonw:
```
SELECT mgs_fs, mgs_fs_count FROM (SELECT count(is_mgs_fs) AS mgs_fs_count FROM target WHERE is_mgs_fs=true GROUP BY mgs_fs),(SELECT LAST(is_mgs_fs) as last_time FROM target GROUP BY mgs_fs) group by mgs_fs
name: target
tags: mgs_fs=6463788d,fs,fs2
time                mgs_fs          mgs_fs_count
----                ------          ------------
0                   6463788d,fs,fs2 408
1604958244843905323 6463788d,fs,fs2
name: target
tags: mgs_fs=fs
time                mgs_fs mgs_fs_count
----                ------ ------------
0                   fs     105
1604970477758642988 fs
name: target
tags: mgs_fs=fs,fs2
time                mgs_fs mgs_fs_count
----                ------ ------------
0                   fs,fs2 417
1604975047786615296 fs,fs2
```

iml-services/iml-stats/src/main.rs
johnsonw: You're right. Filter added.

iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,97 @@ async fn main() -> Result<(), ImlStatsError> {
...
+
+            acc
+        });
+
+    let delete_subset_futures = subsets
+        .into_iter()
+        .filter(|(mgs_fs, (_time, count))| {
+            let mgs_fs_set = fs_names_to_hash_set(&mgs_fs);
+
+            mgs_fs_set != fs_names || *count > 1
```
jgrund, line 597: Why is this an or?

johnsonw: The filter will continue for anything expression that is true and since I want to filter out anything the following:
```
mgs_fs_set == fs_names && count <= 1
```
The inverse of this expression will accept any case that does not match the above. Here is sample data that is run against expression `mgs_fs_set != fs_names || *count > 1`:
```
mgs_fs_set: {"fs"} fs_names: {"fs2", "fs"}, count: 1, passes filter: true
mgs_fs_set: {"fs"} fs_names: {"fs2", "fs"}, count: 2, passes filter: true
mgs_fs_set: {"fs2"} fs_names: {"fs2", "fs"}, count: 1, passes filter: true
mgs_fs_set: {"fs2"} fs_names: {"fs2", "fs"}, count: 2, passes filter: true
mgs_fs_set: {"fs2", "fs"} fs_names: {"fs2", "fs"}, count: 1, passes filter: false
mgs_fs_set: {"fs2", "fs"} fs_names: {"fs2", "fs"}, count: 2, passes filter: true
mgs_fs_set: {"fs3", "fs", "fs2"} fs_names: {"fs2", "fs"}, count: 1, passes filter: true
mgs_fs_set: {"fs3", "fs", "fs2"} fs_names: {"fs2", "fs"}, count: 2, passes filter: true
```

jgrund (approved):

iml-services/iml-stats/src/main.rs
```diff
@@ -531,51 +532,97 @@ async fn main() -> Result<(), ImlStatsError> {
...
+                    time,
+                    mgs_fs,
+                    TargetVariant::MGT
+                )
+            } else {
+                format!(
+                    "DELETE FROM target WHERE time <= {} AND mgs_fs = '{}' AND kind = '{}'",
+                    time,
+                    mgs_fs,
+                    TargetVariant::MGT
```
ip1981, line 614: The queries look the same except `time <= {}`.

johnsonw: Yes, in the case where the sets are equal we want to delete all entries except for the latest entry. In the case where we are dealing with a subset, we want to delete all entries <= the latest time.

PR #2381, opened 2020-11-12
 Duplicate target insert during attempted target mount

### LINKED ISSUE #2380: Duplicate target insert during attempted target mount
When targets are mounted quickly in succession (i.e after a failed mount), we can get data from the agents showing the target mounted on both hosts. When this occurs we will try to insert the target twice into the db.

Instead we should filter out any items that show as mounted but are unknown to the current host.

jgrund:
When targets are mounted quickly in succession (i.e after a failed mount), we can get data
from the agents showing the target mounted on both hosts.
When this occurs we will try to insert the target twice into the db.

Instead we should filter out any items that show as mounted but are unknown to the current host.

Fixes #2380

### DIFF
--- a/iml-services/iml-device/src/lib.rs
+++ b/iml-services/iml-device/src/lib.rs
@@ -407,7 +407,7 @@ pub fn find_targets<'a>(
                 .iter()
                 .filter(|(k, _)| *k != &fqdn)
                 .filter_map(|(k, x)| {
-                    tracing::debug!("Searching for device {:?} on {}", &x, &k);
+                    tracing::debug!("Searching for device {:?} on {}", dev_id, &k);
                     x.0.get(&dev_id)?;
 
                     tracing::debug!("found device on {}!", &k);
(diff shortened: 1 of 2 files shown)


iml-services/iml-device/src/lib.rs
```diff
@@ -457,6 +458,30 @@ pub fn find_targets<'a>(
             uuid: fs_uuid.into(),
             mount_path: Some(mntpnt.0.to_string_lossy().to_string()),
         })
+        .collect();
```
johnsonw, line 461: Is the collect necessary? Couldn't it go straight into a fold?

johnsonw (approved):

iml-services/iml-device/src/lib.rs
jgrund: It could, but up to this point there seemed to be one concept per iterator so I decided to keep it that way.

ip1981 (approved):

PR #2387, opened 2020-11-16
Search for config in different places

mkpankov:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2387)

### DIFF
--- a/iml-manager-cli/src/main.rs
+++ b/iml-manager-cli/src/main.rs
@@ -86,9 +86,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     match matches {
         App::Shell { .. } => (),
-        _ => {
-            dotenv::from_path("/var/lib/chroma/iml-settings.conf").expect("Could not load cli env")
-        }
+        _ => dotenv::from_path("/etc/emf/emf-settings.conf")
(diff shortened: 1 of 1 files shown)


iml-manager-cli/src/main.rs
```diff
@@ -86,9 +86,10 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     match matches {
         App::Shell { .. } => (),
-        _ => {
-            dotenv::from_path("/var/lib/chroma/iml-settings.conf").expect("Could not load cli env")
-        }
+        _ => dotenv::from_path("/etc/emf/emf-settings.conf")
+            .or_else(|_| dotenv::from_path("/etc/iml/iml-settings.conf"))
+            .or_else(|_| dotenv::from_path("/var/lib/chroma/iml-settings.conf"))
```
ip1981, line 91: I understand emf/iml difference, but why the third variant?

Shouldn't the name of the configuration file depend on selfname?
Shouldn't be the files read in order, overriding any previous options, at least the one under /var?

mkpankov: Third is for docker setup.

I don't think we should override anything.

ip1981 (approved):

nlinker (approved):

PR #2388, opened 2020-11-17
Add writing out of config

mkpankov:
As of this patch, IML runs on host successfully.
```
[root@adm vagrant]# which iml
/bin/iml
[root@adm vagrant]# file `which iml`
/bin/iml: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.32, BuildID[sha1]=aa6069cb937467fb51107fab9a6aec7de9319d09, not stripped
[root@adm vagrant]# iml server

[root@adm vagrant]# echo $?
0
[root@adm vagrant]# 
```

image

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2388)


ip1981:
Docker cp could be more reliable (no assumption regarding docker volumes) or even bind mount of the file.

mkpankov:
@ip1981 it requires a running container. Ours is one-shot so synchronization would be more effort than it's worth IMO.

What do you mean by bind mount?

ip1981:
But you need a running container anyway, no?

By bind mount I mean mounting /etc/iml or one file to be directly available from the container

iml-docker.service
```diff
@@ -13,6 +13,7 @@ WorkingDirectory=/etc/iml-docker
 ExecStartPre=-/usr/bin/docker load -i /var/lib/iml-images.tgz
 ExecStart=/usr/bin/docker stack deploy -c docker-compose.yml -c docker-compose.overrides.yml iml --resolve-image=never
 ExecStart=/bin/bash -c 'until /usr/bin/iml server list > /dev/null 2>&1; do sleep 1; done'
```
jgrund, line 15: A bit of trickyness here. We are going to need the iml command to have it's setting in place before we can actually run this check.

mkpankov:
@ip1981 no. Currently we copy the file from volume after `iml-setup` container has run and exited.

iml-docker.service
mkpankov: I've moved copying to `ExecStart` before `iml server list`.

> If more than one command is specified, the commands are invoked sequentially in the order they appear in the unit file. If one of the commands fails (and is not prefixed with "-"), other lines are not executed, and the unit is considered failed.

https://www.freedesktop.org/software/systemd/man/systemd.service.html

So this should be fine.

jgrund: What happens if that file does not exist yet? It gets generated early on during startup.

jgrund: Also, we can't depend on the docker/volume path being static, for example it's a different path in exascaler.

mkpankov: It can't not exist yet because next command will be started only after current command finishes.

I'll investigate getting the volume path.

docker/iml-copy-settings.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
jgrund, line 1: Can we rename this to `copy-embedded-settings`?

mkpankov: Sure

.copr/Makefile
```diff
@@ -73,7 +73,7 @@ iml-docker-srpm:
 	mkdir -p ${TMPDIR}/_topdir/{SOURCES,SPECS}
 	mkdir -p ${TMPDIR}/scratch/iml-docker
 
-	cp -r docker/{docker-compose.yml,iml-images.tgz,iml-cli-proxy.sh,update-embedded.sh} ${TMPDIR}/scratch/iml-docker/
+	cp -r docker/{docker-compose.yml,iml-images.tgz,iml-cli-proxy.sh,update-embedded.sh,copy-embedded-settings.sh} ${TMPDIR}/scratch/iml-docker/
```
jgrund, line 76:
```suggestion
	cp -r docker/{docker-compose.yml,iml-images.tgz,iml-cli-proxy.sh,update-embedded.sh,copy-embedded-settings} ${TMPDIR}/scratch/iml-docker/
```

docker/copy-embedded-settings.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
jgrund, line 1: Let's call this `copy-embedded-settings` instead.

iml-docker.spec
```diff
@@ -40,6 +41,7 @@ mv iml-docker.service %{buildroot}%{_unitdir}
 %attr(0640, root, root) %{_sharedstatedir}/iml-images.tgz
 %attr(0755, root, root) %{_bindir}/iml
 %attr(754, root, root) %{_bindir}/update-embedded
+%attr(754, root, root) %{_bindir}/copy-embedded-settings.sh
```
jgrund, line 44:
```suggestion
%attr(754, root, root) %{_bindir}/copy-embedded-settings
```

iml-docker.spec
```diff
@@ -31,6 +31,7 @@ mkdir -p %{buildroot}%{_unitdir}
 cp docker-compose.yml %{buildroot}%{_sysconfdir}/iml-docker
 mv iml-images.tgz %{buildroot}%{_sharedstatedir}
 mv iml-cli-proxy.sh %{buildroot}%{_bindir}/iml
+mv copy-embedded-settings.sh %{buildroot}%{_bindir}/
```
jgrund, line 34:
```suggestion
mv copy-embedded-settings %{buildroot}%{_bindir}/
```

jgrund:
Make sure to add a `Requires: rust-iml-cli` to the `iml-docker.spec`

iml-docker.service
```diff
@@ -12,6 +12,7 @@ EnvironmentFile=-/etc/iml-docker/setup/config
 WorkingDirectory=/etc/iml-docker
 ExecStartPre=-/usr/bin/docker load -i /var/lib/iml-images.tgz
 ExecStart=/usr/bin/docker stack deploy -c docker-compose.yml -c docker-compose.overrides.yml iml --resolve-image=never
+ExecStart=/bin/bash /usr/bin/copy-embedded-settings.sh
```
jgrund, line 15:
```suggestion
ExecStart=/bin/bash /usr/bin/copy-embedded-settings
```

docker/copy-embedded-settings.sh
mkpankov: Done

docker/copy-embedded-settings
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -eux
+/bin/mkdir -p /etc/iml
+service=iml_influxdb
+trailer=$(/usr/bin/docker service ps -f "name=$service.1" $service -q --no-trunc | head -n1)
+until /usr/bin/docker exec $service.1.$trailer /bin/bash -c 'until [ -f /var/lib/chroma/iml-settings.conf ]; do echo "Waiting for settings."; sleep 1; done'; do
```
ip1981, line 6: Paranoia: is writing the config file atomic? 🤔

mkpankov: Yes https://github.com/whamcloud/integrated-manager-for-lustre/blob/docker-config/chroma_core/lib/service_config.py#L782-L784

ip1981 (approved):

mkpankov:
As of this patch, IML runs on host successfully.
```
[root@adm vagrant]# which iml
/bin/iml
[root@adm vagrant]# file `which iml`
/bin/iml: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.32, BuildID[sha1]=aa6069cb937467fb51107fab9a6aec7de9319d09, not stripped
[root@adm vagrant]# iml server

[root@adm vagrant]# echo $?
0
[root@adm vagrant]# 
```

jgrund:
Looks like the latest Jenkins test failed with:

```
02:09:13.719  Error: Assert("Command Command { std: \"ssh\" \"-o\" \"StrictHostKeyChecking=no\" \"-o\" \"UserKnownHostsFile=/dev/null\" \"-i\" \"id_rsa\" \"10.73.10.10\" \"systemctl status iml-docker.service\", kill_on_drop: false } failed to succeed after 50 attempts.")
```

jgrund:
Looking at the sosreport for this run, `/etc/hosts` does not have the nginx entry in it yet, which is the cause of the docker failure:

Screen Shot 2020-12-03 at 5 50 00 PM

jgrund:
Interesting side-effect of this is that we could now run `iml` commands directly on the host (not for this patch).

iml-system-test-utils/src/lib.rs
```diff
@@ -516,8 +519,7 @@ pub async fn deploy_servers(config: Config) -> Result<Config, TestError> {
     Ok(config)
 }
 
-pub async fn configure_docker_network(config: &Config) -> Result<(), TestError> {
-    let host_list = config.profile_map.to_server_list();
```
jgrund, line 520: Why change this?

jgrund (approved):

iml-system-test-utils/src/lib.rs
johnsonw: I'm curious as to what the value is here. It doesn't look like it's causing an issue though.

johnsonw (approved):

iml-system-test-utils/src/lib.rs
jgrund: It's equivalent to what it was before.

PR #2389, opened 2020-11-17
Use fqdn and fsname in Task table

### LINKED ISSUE #2358: Update chroma_core_task to not rely on filesystem_id
The current `chroma_core_task` table schema looks like:

```
     Column      |           Type           |                           Modifiers
-----------------+--------------------------+---------------------------------------------------------------
 id              | integer                  | not null default nextval('chroma_core_task_id_seq'::regclass)
 name            | character varying(128)   | not null
 start           | timestamp with time zone | not null
 finish          | timestamp with time zone |
 state           | character varying(16)    | not null
 fids_total      | bigint                   | not null
 fids_completed  | bigint                   | not null
 fids_failed     | bigint                   | not null
 data_transfered | bigint                   | not null
 single_runner   | boolean                  | not null
 keep_failed     | boolean                  | not null
 actions         | text[]                   | not null
 args            | jsonb                    | not null
 filesystem_id   | integer                  | not null
 running_on_id   | integer                  |
```

Instead of `filesystem_id` we should have `fs_name` which we've already determined is unique per-install and will allow us to not rely on filesystem detection for tasks.

utopiabound:
Use running_on_fqdn instead of linking to ManagedHost
Use fsname instead of linking to ManagedFilesystem

Fixes #2358

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2389)


johnsonw (approved):

johnsonw:
Looks good, when looking over the files, but two things I notice:
1. sqlx-data.json needs to be updated
2. Integration tests are failing

Going to +1 pending these two things are fixed

johnsonw (dismissed):

PR #2397, opened 2020-11-23
Try to use line buffering in postoffice

jgrund:
Try to use line buffering in the postoffice UDS connection.

In addition, reuse the same client for all connections instead of
streaming one connection at a time.

### DIFF
--- a/iml-agent/src/daemon_plugins/postoffice.rs
+++ b/iml-agent/src/daemon_plugins/postoffice.rs
@@ -6,13 +6,11 @@ use crate::{
     agent_error::ImlAgentError,
     daemon_plugins::{DaemonPlugin, Output},
     env,
-    http_comms::streaming_client::send,
+    http_comms::crypto_client,
 };
 use async_trait::async_trait;
-use futures::{
-    stream::{StreamExt as _, TryStreamExt},
-    Future, FutureExt,
-};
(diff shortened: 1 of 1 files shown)


iml-agent/src/daemon_plugins/postoffice.rs
```diff
@@ -52,36 +50,61 @@ fn start_route(mailbox: String) -> Trigger {
         let _ = fs::remove_file(&addr).await.map_err(|e| {
             tracing::debug!("Failed to remove file {}: {}", &addr, &e);
         });
-        let mut listener = UnixListener::bind(addr.clone()).map_err(|e| {
-            tracing::error!("Failed to open unix socket {}: {}", &addr, &e);
-            e
-        })?;
-        let mut incoming = listener.incoming().take_until(tripwire);
+
+        let conn = UnixStream::connect(&addr)
```
utopiabound, line 54: Why is this changing?  This side should be the server listening for connections right?

utopiabound (changes requested):

iml-agent/src/daemon_plugins/postoffice.rs
jgrund: How many writers will there be on a socket at once?

utopiabound: current usage between 0 and number of mounted targets.

PR #2417, opened 2020-12-04
Add Basic formatting pre-commit check

utopiabound:
Formatting check supports both rust and python code.
Doesn't add the weight of doing a cargo build on a system
that doesn't support it.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2417)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,9 +49,3 @@ Precommit checks are run by [rusty-hook](https://github.com/swellaby/rusty-hook)
 ```sh
 cargo install --git https://github.com/swellaby/rusty-hook.git
 ```
-
-Basic formatting pre-commit hook
-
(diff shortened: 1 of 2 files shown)


jgrund (changes requested):
This will conflict with rusty-hook, which also installs a pre-commit hook for rust.

utopiabound:
It's an alternative, since I don't commit from a build machine in my work-flow.

jgrund:
> It's an alternative, since I don't commit from a build machine in my work-flow. 

Let's not document it then. I don't want to have competing pre-commit hooks in the readme

jgrund (approved):

PR #2419, opened 2020-12-06
Skip identical daemon plugin updates

jgrund:
Skip sending updates for all daemon plugins is the previous update is
equal to the current one within a session.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2419)

### DIFF
--- a/iml-agent/src/http_comms/session.rs
+++ b/iml-agent/src/http_comms/session.rs
@@ -4,9 +4,9 @@
 
 use crate::{
     agent_error::{NoSessionError, Result},
-    daemon_plugins::{DaemonBox, OutputValue},
+    daemon_plugins::{DaemonBox, Output, OutputValue},
 };
-use futures::{channel::oneshot, future, future::Either, Future};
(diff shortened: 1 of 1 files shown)


ip1981 (approved):

iml-agent/src/http_comms/session.rs
```diff
@@ -287,6 +294,14 @@ impl Session {
                     Ok(None)
                 }
                 Either::Right((Ok(Some(x)), _)) => {
+                    let mut last_update = last_update.lock().await;
+
+                    if last_update.deref().as_ref() == Some(&x) {
```
mkpankov, line 299:
```suggestion
                    if Some(last_update.deref().as_ref()) == &x {
```

mkpankov (approved):

iml-agent/src/http_comms/session.rs
jgrund: Left side is already an `Option`.

PR #2426, opened 2020-12-09
Use kill_on_drop for all agent commands

jgrund:
If an action is cancelled, any associated child processes will keep
running.

Use
[kill_on_drop](https://docs.rs/tokio/0.3.5/tokio/process/struct.Command.html#method.kill_on_drop
) to make sure the child process is killed.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2426)

### DIFF
--- a/iml-agent/src/action_plugins/firewall_cmd.rs
+++ b/iml-agent/src/action_plugins/firewall_cmd.rs
@@ -11,6 +11,7 @@ static DUPLICATE_MESSAGE: &[u8] = b"Warning: ALREADY_ENABLED\n";
 async fn port_change(action: &str, port: u16, proto: String) -> Result<(), ImlAgentError> {
     Command::new("firewall-cmd")
         .arg(format!("--{}-port={}/{}", action, port, proto))
+        .kill_on_drop(true)
         .checked_output()
         .await
         .map(drop)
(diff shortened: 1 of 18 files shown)


iml-agent/src/action_plugins/firewall_cmd.rs
```diff
@@ -11,6 +11,7 @@ static DUPLICATE_MESSAGE: &[u8] = b"Warning: ALREADY_ENABLED\n";
 async fn port_change(action: &str, port: u16, proto: String) -> Result<(), ImlAgentError> {
     Command::new("firewall-cmd")
         .arg(format!("--{}-port={}/{}", action, port, proto))
+        .kill_on_drop(true)
```
johnsonw, line 14: I noticed on check_stonith the `kill_on_drop` came before the args. It probably doesn't matter though, right?

johnsonw (approved):

iml-agent/src/action_plugins/firewall_cmd.rs
jgrund: doesn't matter it's a builder interface. If it mattered it would fail to compile :)

ip1981 (approved):

PR #2427, opened 2020-12-09
Add LNet action plugins

johnsonw:
Create LNet action plugins to be invoked in the rust agent.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2427)

### DIFF
--- a/iml-agent/src/action_plugins/action_plugin.rs
+++ b/iml-agent/src/action_plugins/action_plugin.rs
@@ -5,7 +5,7 @@
 use crate::{
     action_plugins::{
         check_kernel, check_stonith, firewall_cmd, high_availability, kernel_module, lamigo, ldev,
-        lpurge, lustre,
+        lnet, lpurge, lustre,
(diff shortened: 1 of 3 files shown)


jgrund (changes requested):
A few things need to be updated.

1. All commands should use `.kill_on_drop(true)`
1. Move over the existing comments from the Python impl to this one, they looked like they were useful.
1. `configure_lnet` is missing from this impl. Instead of the modprobe stuff it did before it should use dynamic lnet configuration https://wiki.lustre.org/Dynamic_LNet_Configuration_and_lnetctl

nlinker (commented):
Do you have the examples of the rest calls to test it locally against vagrant VMs?

johnsonw:
@nlinker these will be added in a separate patch.

jgrund (changes requested):
This PR needs to be moved.

PR #2443, opened 2020-12-17
Build rust rpm's

johnsonw:
The sccahe installation from crates is not functioning properly on
centos 7. Instead, pull the repo down and build sccache in release mode
with no default features. This fixes the issue. Additionally, add a
"--clean" flag that will rebuild all of the rpms. This is useful if only
a specific rpm is being built but others also need to be built to ensure
that db migrations are baked into each service. The operation will still
be fairly quick since sccache will be used.

### DIFF
--- a/vagrant/scripts/build_rust_rpms.sh
+++ b/vagrant/scripts/build_rust_rpms.sh
@@ -1,21 +1,36 @@
+set -x
 yum install -y openssl-devel
-yum install -y cargo rpm-build
+yum install -y cargo rpm-build git
 
-[ -f /root/.cargo/bin/sccache ] && echo "sccache already installed. Skipping." || cargo install sccache
+unset RUSTC_WRAPPER
 
+[ -f /sccache/target/release/sccache ] \
+    && echo "sccache already installed. Skipping." \
+    || (cd / \
+        && rm -rf /sccache \
+        && git clone https://github.com/mozilla/sccache.git \
+        && cd /sccache \
+        && cargo build --release --no-default-features)
+
+sed -i '/^PATH=\$PATH/d' /root/.bash_profile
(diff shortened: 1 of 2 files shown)


vagrant/scripts/build_rust_rpms.sh
```diff
@@ -1,21 +1,36 @@
...
 export SCCACHE_DIR=/root/.cache
-export RUSTC_WRAPPER=/root/.cargo/bin/sccache
+export RUSTC_WRAPPER=/sccache/target/release/sccache
 EOF
 
 source /root/.bash_profile
 
+
 cd /integrated-manager-for-lustre \
+    && [ "$1" = "--clean" ] && cargo clean || echo "Attempting to build rpm's without cleaning." \
```
nlinker, line 35: Does it clean `sscache` as well? This seems a bit unexpected, since cargo doesn't know about sscache (or does it?).

jgrund:
> The sccahe installation from crates is not functioning properly on centos 7.

What is the issue?

vagrant/scripts/build_rust_rpms.sh
johnsonw: In my case, I made a change to one of the migrations and changed a small piece of code in one of the services. When building the rpm's, cargo just saw that that one service was changed, so when it built the rpms, it used the cache data for the other services. This ultimately led to the other services having the previous SQL migrations baked into them while the service I made a change to baked in the new migration script. So when I installed the rpms, one service had a different migration baked into it, which created a race condition to run the SQL migrations and ultimately caused the services to crash. What this does is clean out cargo to ensure the other services are also built. Sccache will still be used to speed up compile time.

PR #2444, opened 2020-12-18
Update docker setup

jgrund:
Update the IML docker setup to not print output while starting.

In addition, use a container that becomes stable earlier to copy
settings to the host.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2444)

### DIFF
--- a/docker/copy-embedded-settings
+++ b/docker/copy-embedded-settings
@@ -1,10 +1,12 @@
 #!/usr/bin/env bash
-set -eux
+set -eu
+
 /bin/mkdir -p /etc/iml
-service=iml_influxdb
+service=iml_nginx
 trailer=$(/usr/bin/docker service ps -f "name=$service.1" $service -q --no-trunc | head -n1)
-until /usr/bin/docker exec $service.1.$trailer /bin/bash -c 'until [ -f /var/lib/chroma/iml-settings.conf ]; do echo "Waiting for settings."; sleep 1; done'; do
-  echo "Waiting for container"
-  sleep 1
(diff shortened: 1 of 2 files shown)


docker/copy-embedded-settings
```diff
@@ -1,10 +1,12 @@
...
+set -eu
+
 /bin/mkdir -p /etc/iml
-service=iml_influxdb
+service=iml_nginx
 trailer=$(/usr/bin/docker service ps -f "name=$service.1" $service -q --no-trunc | head -n1)
-until /usr/bin/docker exec $service.1.$trailer /bin/bash -c 'until [ -f /var/lib/chroma/iml-settings.conf ]; do echo "Waiting for settings."; sleep 1; done'; do
-  echo "Waiting for container"
-  sleep 1
+until /usr/bin/docker exec $service.1.$trailer sh -c 'until [ -f /var/lib/chroma/iml-settings.conf ]; do echo "copy-embedded-settings: Waiting for EMF container settings..."; sleep 1; done' 2>/dev/null; do
```
beevans, line 7: I think being explicit about what shell to use is a good thing.  sh is *usually* bash, but not always

jgrund: In this case, bash does not exist, *only* sh exists in this container.

beevans: That's weird enough that it merits a comment, I think

jgrund: Ok, I can add in a follow-on

beevans (approved):

PR #2448, opened 2020-12-21
Fixup copy-embedded-settings

jgrund:
`copy-embedded-settings` will find the trailer before looping. If the
container dies after entering the loop, then the loop will never exit as
the trailer is then non-existent.

Instead, we should find the trailer each turn of the loop to ensure we
are referencing a live container.

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2448)


utopiabound (commented):
Looks fine.

PR #2457, opened 2021-02-17
Require packages which work with Python2

ip1981:

---
This change is [Reviewable](https://reviewable.io/reviews/whamcloud/integrated-manager-for-lustre/2457)

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-bandit = "*"
+bandit = "==1.6.2"
 django-nose = "==1.4.6"
 flake8 = "*"
 mock = "==1.0.1"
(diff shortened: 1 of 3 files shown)


jgrund:
You need to update the lockfiles as well.

ip1981:
I didn't see them used though.

ip1981:
`Piplock` is a replacement for `requirements.txt` and we seem not using `Piplock`.

jgrund:
we are using pipenv. You need to update the Pipfile and Pipfile.lock

ip1981:
> we are using pipenv. You need to update the Pipfile and Pipfile.lock

Done. I was more worrying about building for production (which uses `requirements.txt`).

jgrund (changes requested):
Looks like python unit tests are now failing due to a change in bandit

ip1981:
Not needed.

